PR #363, opened 2014-06-09
Improves exceptions logging

BertrandBordage:
While using [ui/django-rq](https://github.com/ui/django-rq) and `django.utils.log.AdminEmailHandler`, I faced two issues:
- `UnpickleError` wasn't caught, and therefore I didn't get any mail when incorrect arguments were passed to RQ;
- Mails were poorly formatted.

At first I thought it was coming from django-rq, but after digging I realized this was coming from RQ.  This pull request fixes these two issues.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -475,9 +475,7 @@ class Worker(object):
         self.heartbeat((job.timeout or 180) + 60)
 
         try:
-            self.procline('Processing %s from %s since %s' % (
-                job.func_name,
-                job.origin, time.time()))
+            func_name = job.func_name
         except UnpickleError:
             # Use the public setter here, to immediately update Redis
             job.set_status(Status.FAILED)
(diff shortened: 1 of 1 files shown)


rq/worker.py
```diff
@@ -474,8 +474,17 @@ def perform_job(self, job):
         self.set_current_job_id(job.id)
         self.heartbeat((job.timeout or 180) + 60)
 
+        try:
+            func_name = job.func_name
+        except UnpickleError:
+            # Use the public setter here, to immediately update Redis
+            job.set_status(Status.FAILED)
+            self.log.exception('RQ failed to unpickle job %s' % job.id)
```
chaffeqa: In dealing with sentry, I've found the best solution is to do simple things like:

```python
except Exception as e:
    self.log.exception(e, extra={...})
```

With any got exception handler, passing the exception itself is the most supported approach, and sentry especially does a good job giving context and insight.

Basically you can reduce a ton of code and verbosity, as well as improve support.

chaffeqa: Which leads into my next proposal:  instead of passing `exc_info` to handlers, I'd rather get the `Exception` instance

untitaker:
FYI https://github.com/rq/rq/issues/1003 is probably relevant: sentry-sdk on PyPI has a new RQ integration that Sentry officially supports

selwin:

PR #391, opened 2014-07-30
Working queue

### LINKED ISSUE #380: WorkingQueue implementation
Having a `WorkingQueue` is useful for:
- Monitoring - such that we know which jobs are currently being worked on
- Tracking jobs that are lost (due to hard shutdown, seg faults, etc) 

The design is largely inspired by what was proposed by @annakan in [this thread](https://github.com/nvie/rq/issues/152#issuecomment-26728632). @glenfant has also been working on this [in his fork](https://github.com/glenfant/rq/tree/wip-queue) so I'll definitely be taking inspirations from there.
### Design
- `WorkingQueue` is a sorted set, scored by time at which a job's execution should end (job start + timeout).
- Jobs are added to `WorkingQueue` when they start and removed `WorkingQueue` when they fail (errors and timeout) or finished executing
- Jobs whose score are lower than the current timestamp is considered "lost".

selwin:
As per usual, comments are welcome :)

Fixes #380

### DIFF
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -40,11 +40,12 @@ class StartedJobRegistry:
 
     def get_job_ids(self, start=0, end=-1):
         """Returns list of all job ids."""
+        self.move_expired_jobs_to_failed_queue()
         return [as_text(job_id) for job_id in
                 self.connection.zrange(self.key, start, end)]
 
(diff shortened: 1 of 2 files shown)


aroberts:
Any plans to release this in a new version?

PR #417, opened 2014-09-05
Added test worker for unit tests - fixes #413

### LINKED ISSUE #413: Worker fork() does not play nicely with Django unit tests (e.g. sqlite3 in-memory)
As per documentation, I'm using the following inside my unit tests;
`get_worker().work(burst=True)`

If you use this in combination with an in-memory SQLite3 database, which is the default for Django 1.6 at least, each job execution causes a fork() to happen. Any changes that happen in the database during task execution are then effectively discarded, due to fork() creating a copy of the in-memory database.

To get around this, I had to specify `TEST_NAME` so that it didn't default to `in-memory`.

E.g.;

```
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.sqlite3',
        'NAME': '/tmp/db.sqlite3',
        'TEST_NAME' : '/tmp/dbtest.sqlite3'
    }
}
```

At the very least, this deserves a doc patch I think.

Thoughts?

foxx:
Seems to work fine for me, but let me know if this needs some more unit tests.

### DIFF
--- a/rq/__init__.py
+++ b/rq/__init__.py
@@ -7,7 +7,7 @@ from .connections import (Connection, get_current_connection, pop_connection,
 from .job import cancel_job, get_current_job, requeue_job
 from .queue import get_failed_queue, Queue
 from .version import VERSION
-from .worker import Worker, WorkerTest
+from .worker import Worker, SimpleWorker
 
 __all__ = [
     'use_connection', 'get_current_connection',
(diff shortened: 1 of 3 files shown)


DylanYoung:
I wonder if the functionality to run the sqlite db in shared mode should be documented: https://code.djangoproject.com/ticket/12118 

Just a thought.

PR #434, opened 2014-10-09
triggering shutdown by setting a redis flag

jtushman:
**Problem Statement**
We use rq on heroku and their way to the shutdown the worker that runs rq does not ensure a safe shutdown.
   
From Heroku docs:  https://devcenter.heroku.com/articles/dynos
```
When a worker is stoped by ps:scaling or pushing out a new release or any heroku operation.
The dyno manager sends the working SIGTERM

The process then has 10 seconds to shut down gracefully.

If the process is still alive -- then SIGKILL is then sent
```

So if a job takes more then 10 secs (which many of our jobs do) -- we are going to be out of luck.  The job will be killed in a potentially unsafe point

So we needed an approach where we could set a flag on redis, that the rq worker can ping.

I introduced the key:  'rq:worker:pause_work'  Which any process can set.  If the worker sees that it has been set -- it hops into a pause loop, until the key is deleted.

In Worker#work at the top of the `while True:` block

``` python
    try:
        before_state = None
        notified = False

        while Worker.paused() and not self.stopped:

            if burst:
                self.log.warn('Paused in burst mode -- exiting.')
                self.log.warn('Note:  There could still be unperformed jobs on the queue')
                raise StopRequested

            if not notified:
                self.log.warn('Stopping on pause request REALLY.')
                before_state = self.get_state()
                self.set_state('paused')
                notified = True
            time.sleep(1)
    except StopRequested:
        break
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,3 @@
 .tox
 .vagrant
 Vagrantfile
-
-# PyCharm
-.idea
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-redis
-click
+redis==2.7.0
+click>=3.0.0
--- a/rq/cli/cli.py
+++ b/rq/cli/cli.py
@@ -16,6 +16,7 @@ from rq import Connection, get_failed_queue, Queue
 from rq.contrib.legacy import cleanup_ghosts
 from rq.exceptions import InvalidJobOperationError
 from rq.utils import import_attribute
+from rq.suspension import suspend as connection_suspend, resume as connection_resume, is_suspended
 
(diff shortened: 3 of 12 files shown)


jtushman:
Happy New Year @nvie, @selwin!  I am back from break -- lemme know if this needs any more polish

selwin:
I'm going to merge this in. Thanks a lot for your hardwork @jtushman !

jtushman:
Woot!

dfrankow:
I see that this was merged in, including into the rq command-line client ("rq suspend" and "rq resume").

But how exactly should this be used for people running on heroku? Are there any docs?

PR #446, opened 2014-11-12
Adding optional list handling to the exc_handler option in Worker.

bradleyy:
Adding command line --exception_handler option (with multiple entries allowed) to `rq worker`
Added tests for command line options.


robodude666:
@selwin Is `exception_handlers` still an idea? I'm not seeing related changes in master.

selwin:
Yes it is. I don't know if anyone's working on it at the moment though.

bradleyy:
Mea culpa; I believe that I have this hacked up on my hard drive somewhere, but ADHD got the better of me. Look for a PR in the next couple of hours.

bradleyy:
Not 100% proud of the exception_handlers="default" kw param; couldn't think of a better method to:

a) default to default
b) if you specify a list, overwrite (with the option to specify a blank list-- no handlers)
c) if you specify a method, add that as a singleton (new behavior: this replaces default handler)

Open to changing the implementation if there is something better.

robodude666:
Given the fact that all of the `Worker.__init__`'s parameters use the value of `None` to enable the class default, I would suggest leaving it as `exception_handlers=None` to use the default fail queue.

We'd then be able to use:

`exception_handlers=[]` to disable completely, and
`exception_handlers=[...]` and  `exception_handlers=my_handler` would be user-defined handlers.

bradleyy:
@robodude666 thanks for the tip; implemented that.

rq/worker.py
```diff
@@ -122,7 +122,7 @@ def find_by_key(cls, worker_key, connection=None):
 
     def __init__(self, queues, name=None,
                  default_result_ttl=None, connection=None,
-                 exc_handler=None, default_worker_ttl=None, job_class=None):  # noqa
+                 exception_handlers=None, default_worker_ttl=None, job_class=None):  # noqa
```
selwin: We should provide a deprecation path for `exc_handler` and display a warning message if `exc_handler` is used. Something like this would be good:
```python
if exception_handlers is not None:
    for handler in exception_handlers:
        self.push_exc_handler(handler)

else:
    self.push_exc_handler(self.move_to_failed_queue)
    if exc_handler is not None:
         self.push_exc_handler(exc_handler)
        # Display warning and ask user to use `exception_handlers` instead

```

bradleyy:
@selwin @nvie I've updated this and merged; is there anything else that needs done before this is merged?

nvie:
I think this is looking good, don't you, @selwin? The only thing missing to wrap it up is to write up some docs (see the `gh-pages` branch) to mention the option, and to specify the format of the argument value you'll have to pass it (and how that is interpreted). (I'm talking about the `import_attribute(…)` logic here.)

rq/cli/cli.py
```diff
@@ -166,6 +167,9 @@ def worker(url, config, burst, name, worker_class, job_class, queue_class, path,
     cleanup_ghosts(conn)
     worker_class = import_attribute(worker_class)
     queue_class = import_attribute(queue_class)
+    exception_handlers = []
```
selwin: `exception_handlers` should default to `None` instead of empty list. An empty list would disable the default behavior of moving failed jobs to `FailedQueue`.

selwin:
Thanks for the effort in creating this PR @bradleyy :). @nvie I left one (hopefully last) comment with regards to the default `exception_handlers` configuration behavior on `rqworker` script.

selwin:
Superseded by #557

PR #463, opened 2014-12-09
Skip queue

conslo:
You could view this as an alternative (and less-code) implementation of #452:

I ran into a situation where it would be very advantageous to queue a specific job *now* (instead of the end of the queue). So I implemented a `skip_queue` option the the various `Queue.enqueue` methods, which does an lpush instead of an rpush.

To accomplish #452 with this you'd enqueue each job specifying `skip_queue=True`. A better way to do this might be changing the default for `skip_queue` from `False` to `None`, and having the lowest level method here (`Queue.push_job_id()`) pull a default from the `Queue` (which we could set at initialization, default `False`). If this is desired renaming the argument from `skip_queue` to something more like `fifo` would likely be a good idea.

If the maintainers think this is a good idea/approach I'd be happy to implement it as part of this PR.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -161,14 +161,18 @@ class Queue(object):
             if self.job_class.exists(job_id, self.connection):
                 self.connection.rpush(self.key, job_id)
 
-    def push_job_id(self, job_id, pipeline=None):
-        """Pushes a job ID on the corresponding Redis queue."""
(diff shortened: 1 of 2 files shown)


conslo:
Yay holidays! Back to this; you mean add to the gh-pages? Sure, I'll open a PR to add to that branch.

selwin:
Thanks for this PR!

PR #471, opened 2015-01-09
Adding meta data: user_id and request_environ

mtsgrd:
It is often useful to pass request environment information and user id information to the background worker from the front end. See below for an example code of how I use it with Flask for impersonation while processing background jobs. I also pass a serialized version of the wsgi environment (after converting buffers to strings) and recreate Flask `RequestContext`'s in the worker (mostly for logging purposes).

```
class CustomJob(Job):
    """Subclassing RQ Job to customize behavior"""

    def __init__(self, *args, **kwargs):
        super(YoJob, self).__init__(*args, **kwargs)

    # Job execution
    def perform(self):
        """Invokes the job function with the job arguments."""
        if self.request_environ:
            # If a request environment is attached to the job then we simulate
            # a real request environment.
            request_context = current_app.request_context(
                load_environ(self.request_environ))
        else:
            # Otherwise we use a test_request_context.
            request_context = current_app.test_request_context()

        with request_context:
            current_app.preprocess_request()
            if not g.identity.user and self.user_id:
                identity = Identity(self.user_id)
                # Set this identity on the thread.
                principals.set_identity(identity)
                # Tell listeners that the identity has changed.
                identity_changed.send(current_app, identity=identity)

            _job_stack.push(self.id)

            try:
                self._result = self.func(*self.args, **self.kwargs)
                response = make_json_response({'result': self._result})
                current_app.process_response(response)
            finally:
                assert self.id == _job_stack.pop()

            return self._result
```

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -92,7 +92,8 @@ class Job(object):
     # Job construction
     @classmethod
     def create(cls, func, args=None, kwargs=None, connection=None,
-               result_ttl=None, ttl=None, status=None, description=None, depends_on=None, timeout=None,
+               result_ttl=None, ttl=None, status=None, description=None,
+               user_id=None, act_as_user_id=None, depends_on=None, timeout=None,
                id=None):
         """Creates a new Job instance for the given function, arguments, and
         keyword arguments.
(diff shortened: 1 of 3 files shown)


selwin:
The `request_environ` use case is too specific. But we do have `job.meta` that can be used to store arbitrary data. I think we should allow users to easily update this so I'm open to adding an `meta` argument to `enqueue` like this:
```python
queue.enqueue(
    my_function,
    metadata={'foo': 'bar')
)
```
Please open another PR for this.

PR #475, opened 2015-01-24
Deferred job status

selwin:
This pull request introduces a new job status: `DEFERRED`. Jobs that have not been queued due to unmet dependencies are now assigned this status. Deferred Jobs are also put in a `DeferredJobsRegistry` so we can keep track of jobs that will be queued when their dependencies are met.

I plan on merging this into master in a week, @nvie please let me know if there's any feedback.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -544,8 +544,14 @@ class Job(object):
 
             rq:job:job_id:dependents = {'job_id_1', 'job_id_2'}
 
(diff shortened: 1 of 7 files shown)


nvie:
I skimmed this change, and don't have any comments. That said, I'm not entirely sure what happens under the hood and if I understand all consequences and honestly lack the time to fully dive in now, so if you want to maintain this feature, I'm happy to merge it in.

PR #477, opened 2015-01-29
TTL tests and fixes

glaslos:
I did some tests regarding the comment https://github.com/nvie/rq/pull/453#issuecomment-66739929
Seems like the job finishes even thou the job has already expired, no hiccup but we get a result. Is this expected behavior?


selwin:
That's because the running Python process still has a copy of the job instance which is then saved to Redis again after execution is finished. However, in the cases of hard shutdown or power failure, job data will still be lost. Also, jobs should never expire in the case of failed execution so it can be inspected later. 

So I think we should still set jobs to never expire `job.ttl` right before execution.

glaslos:
Sounds fair to me. I'll try to fix that this week.

glaslos:
@selwin I added the never expire right before performing the job. I fetch the TTL from redis right after that. This could be also set always to -1. redis.ttl() is O(1).

rq/job.py
```diff
@@ -485,6 +485,8 @@ def delete(self, pipeline=None):
     # Job execution
     def perform(self):  # noqa
         """Invokes the job function with the job arguments."""
+        self.connection.persist(self.key)
+        self.ttl = self.connection.ttl(self.key)
```
selwin: What's the intention of this line? Doesn't `connection.ttl` always return `-1` since we called `connection.persist` on the line above?

glaslos: Yeah, see https://github.com/nvie/rq/pull/477#issuecomment-76408703 we could hard-code it to `-1`

selwin: Ah, I missed your comment above. Yes, please hard code it to -1. This saves a round trip to Redis :)

selwin:
Thanks for the PR :)

PR #478, opened 2015-02-01
Tiny fix of the indefinite article before "RQ"

shvechikov:
"An" should be used here because "RQ" sounds like "ar-kiu".

https://www.englishclub.com/pronunciation/a-an.htm


selwin:
Thanks for the PR!

Issue #480, opened 2015-02-02
job.meta is not being saved


cbenhagen:
test.py:
```python
import socket

from rq import get_current_job


def add(x, y):
    job = get_current_job()
    job.meta['handled_by'] = socket.gethostname()
    job.save()
    return x + y
```

tester.py:
```python
import time

from rq import Queue
from redis import Redis
from test import add

redis_conn = Redis()
q = Queue(connection=redis_conn)

job = q.enqueue(add, 1, 2)
time.sleep(1)
print job.result, job.meta
```
returns:
`3 {}` on 0.4.6

andrevmatos:
Yeah, just came here to look for this issue, that I experimented myself.
Just adding that both using Queue.fetch_job and Job.fetch doesn't restore meta, but only after Job is finished: if it's running (I'm trying to use meta to "report" work progress), meta is correctly fetched with job, but when job finishes, meta is restore to the one that got submited.

selwin:
Mind opening a PR with test case to fix this?

selwin:
Strange, I encountered no such issue in my test. Please see https://github.com/selwin/rq/tree/job-meta-modification . Here's the diff: https://github.com/selwin/rq/compare/job-meta-modification

robodude666:
@cbenhagen, your example doesn't work because the `meta` data is cached on load/creation while the `result` is pulled directly from redis the first time and cached. If you update your `meta` dictionary from within the job, you'll need to perform a `job.refresh()` before being able to access it. You can use the `status` property to determine if the job is finished yet before doing a refresh. After the above change the example works as intended in `0.4.6`.

@selwin's TestCase works because a job fetch is performed after starting the worker. This ends up accomplishing the same thing as the `meta` data is read from the redis store w/ the updated data.

-robodude666

PR #481, opened 2015-02-03
add test for custom meta from inside job

cbenhagen:
see issue #480

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -58,6 +58,12 @@ def access_self():
     return job.id
 
 
+def store_meta(new_meta):
+    job = get_current_job()
+    job.meta.update(new_meta)
+    job.save()
+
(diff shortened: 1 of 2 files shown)


selwin:
I don't think the test you wrote is correct. The `store_meta` function was never executed. It should be:

```python
    def test_custom_meta_from_inside_job_is_persisted(self):
        """Additional meta data on jobs are stored persisted correctly from inside job."""
        job = Job.create(func=store_meta, args=({'foo':'bar'},))
        job.save()
        job.perform()

        job2 = Job.fetch(job.id)
        self.assertEqual(job2.meta['foo'], 'bar')
```

In the updated snippet, no error is raised. Please also see my comment here: https://github.com/nvie/rq/issues/480#issuecomment-74645320

PR #486, opened 2015-02-18
lrem only first occurrence of the job id (we expect only one)

glaslos:
See https://github.com/nvie/rq/issues/472#issuecomment-74859743 for explanation.


selwin:
Thanks for the PR!

PR #491, opened 2015-02-23
Restored the ability to specify connection params in config

selwin:
This PR allows user to specify connection params via config file


nvie:
Thanks!

PR #406, opened 2014-08-28
handle Unicode in Job args & kwargs -- patch adapted from 7wonders

ghost:
This patch allows Jobs to have unicode args and kwargs.

Code is derived from 7wonders. I changed it from %-type formatting to str.format, so that args and kwargs could have non-string arguments. (Previously it was raising AttributeError.)   Test added.

This is my first pull request! Let me know if I should do something else.  Cheers!

PR #492, opened 2015-02-24
Allow non-ASCII characters in arguments

dwaynebailey:
get_call_string() failed if any arguments contained non-ASCII strings.

Fixes #406

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -514,8 +514,15 @@ class Job(object):
         if self.func_name is None:
             return None
 
-        arg_list = [repr(arg) for arg in self.args]
-        arg_list += ['%s=%r' % (k, v) for k, v in self.kwargs.items()]
+        # Python 2/3 compatibility
+        try:
+            arg_list = [repr(arg).decode('utf-8') for arg in self.args]
+        except AttributeError:
+            arg_list = [repr(arg) for arg in self.args]
+
+        kwargs = ['{0}={1!r}'.format(k, v) for k, v in self.kwargs.items()]
+        # Sort here because python 3.3 & 3.4 makes different call_string
+        arg_list += sorted(kwargs)
         args = ', '.join(arg_list)
         return '%s(%s)' % (self.func_name, args)
 
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -23,6 +23,26 @@ except ImportError:
 
 
 class TestJob(RQTestCase):
+    def test_unicode(self):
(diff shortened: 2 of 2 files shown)


selwin:
Thanks @dwaynebailey and @alternativshik !

alternativshik:
@selwin you'r welcome!

ydaniv:
Awesome! :+1:

rq/job.py
```diff
@@ -514,8 +514,15 @@ def get_call_string(self):  # noqa
         if self.func_name is None:
             return None
 
-        arg_list = [repr(arg) for arg in self.args]
-        arg_list += ['%s=%r' % (k, v) for k, v in self.kwargs.items()]
+        # Python 2/3 compatibility
+        try:
+            arg_list = [repr(arg).decode('utf-8') for arg in self.args]
+        except AttributeError:
+            arg_list = [repr(arg) for arg in self.args]
```
nvie: We already have a `as_text()` function that deals with the Py 2/3 differences, so you can just use that instead of doing a try/except here.

nvie: See https://github.com/nvie/rq/blob/219f21b6378bbd66a13e997876c7c7416b27ef46/rq/compat/__init__.py#L54-L62

tests/test_job.py
```diff
@@ -23,6 +23,26 @@
...
+            args=[12, "☃"],
+            kwargs=dict(snowman="☃", null=None),
+        )
+
+        try:
+            # Python 2
+            test_string = u"myfunc(12, u'\\u2603', null=None, snowman=u'\\u2603')".decode('utf-8')
+        except AttributeError:
+            # Python 3
+            test_string = "myfunc(12, '☃', null=None, snowman='☃')"
```
nvie: If you want to distinguish Python 2 from 3, please use an `if` statement for this, not catching an AttributeError. But you can probably also use `as_text()` here.

rq/job.py
alternativshik: Thnx for comments. I'll fix it asap.

tests/test_job.py
alternativshik: I've already fixed this and sent pull-request.

commandtab:
Is there a possibility of this fix getting released soon? I just ran into this bug, and the current release (0.6.0 @6a57d32) appears to still use the non-Unicode-friendly implementation.

Thanks all! ❤️

PR #494, opened 2015-02-25
Add missed `multi` command after `watch`

### LINKED ISSUE #487: Worker fails during adding new dependency to a current job
I tried to test what can happen if current job is added as dependency to a new job and it finishes before `pipe.execute()` here (https://github.com/nvie/rq/blob/master/rq/queue.py#L205).
Deferred set and set of dependents for a current job are filled and deferred job is created in any case. 
If current job finishes before `pipe.execute()` then worker fails to enqueue dependents for this job:

```
rq.exceptions.NoSuchJobError: Unexpected job format: {'status': 'deferred'}
```

If I add `pipe.multi()` after `pipe.watch()` then it seems that everything works. Probably it has been missed there.

ta2-1:
`watch` command should be used in conjunction with `multi` command
which was missed in enqueuing of job with dependencies.
I'm not sure about why `DEFERRED` status is set via different connection (not the current pipe). 
Fix #487

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -199,6 +199,7 @@ class Queue(object):
                     try:
                         pipe.watch(depends_on.key)
                         if depends_on.get_status() != JobStatus.FINISHED:
+                            pipe.multi()
                             job.set_status(JobStatus.DEFERRED)
                             job.register_dependency(pipeline=pipe)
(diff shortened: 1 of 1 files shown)


nvie:
Thanks!

PR #500, opened 2015-03-06
Add REDIS_SSL option in worker config file

zzerk:
Allow the worker to connect to a Redis instance through SSL (ex: Azure
Redis Cache use SSL only by default)

The following option is added to the worker optional config file: 
REDIS_SSL = True

### DIFF
--- a/rq/cli/helpers.py
+++ b/rq/cli/helpers.py
@@ -36,6 +36,7 @@ def get_redis_from_config(settings):
         port=settings.get('REDIS_PORT', 6379),
         db=settings.get('REDIS_DB', 0),
         password=settings.get('REDIS_PASSWORD', None),
+        ssl=settings.get('REDIS_SSL',False)
     )
 
 


nvie:
Even though you should probably use `REDIS_URL` instead, this is good to have for completeness.

nvie:
Thanks!

nvie:
Whoops, too soon. It seems to not play well with older versions of StrictRedis not knowing about the `ssl` param. See https://travis-ci.org/nvie/rq/jobs/53330741. We may need to check for the used redis-py version upfront, before passing this param in.

zzerk:
Ok. I'll try to check that. 

Actually, you cannot always use REDIS_URL:  in some circumstances, you may have to deal with passwords containing '/' or '@' characters which are not handled well by redis-py (this is the case for me for my  Redis Cache instance on Microsoft Azure and I cannot change the password).

zzerk:
StrictRedis ssl parameter was introduced in redis-py 2.10.0

PR #504, opened 2015-03-09
Creating directly "deferred" jobs

glenfant:
Hi,

I have a use case that requires to let a queue create deferred jobs directly, in order to let the programmer create "future" dependencies.

The change is straightforward and adds a ```deferred=False``` keyword arg to ```enqueue(...)``` and ```enqueue_call(...)``` methods of  queue.

This PR includes this change, the specific test cases and does not break existing tests.

Expecting a positive feedback... Best regards

Pinging @Annakan BTW


selwin:
@glenfant can you give an example code on how future dependencies can be created with your proposed API?

I think we can pretty easily create future dependencies without supporting manual creation of deferred jobs. Since we already support creating jobs [using custom IDs](https://github.com/nvie/rq/pull/415), I *think* we can easily support something like this:
```python
deferred_job = queue.enqueue(foo, depends_on="future_job_id")

# Else where in your code
future_job = queue.enqueue(bar, id="future_job_id")
```

This way, `deferred_job` will be enqueued when `future_job` is successfully executed.

glenfant:
Yeah, Many thanks @nvie !

This tip made my day. I didn't notice this new feature since I haven't been using rq for some months. Then I thought it needed the id of an existing job (recorded in Redis).

Just one fix for others who would try to use this tip, the second line should be

```python
future_job = queue.enqueue(bar, job_id="future_job_id")
```
This ticket can be closed. Notifying @Annakan

caseyjlaw:
Where can I find more documentation on the custom job ID feature? I've seen it come up in a number of issues, but still don't see how it could work for my use case. 
  I'm missing something obvious in the example above by @selwin. How the queuing/execution different from this:

    future_job = queue.enqueue(bar)
    deferred_job = queue.enqueue(foo, depends_on=future_job)

glenfant:
@caseyjlaw This works perfetly for me :
```python
import uuid
future_job_id = 'custom-{}'.format(uuid.uuid4()'
```
You can enqueue jobs that depend on job ids that do no exist, then create later this job forcing its id.

PR #506, opened 2015-03-17
Prevent `Queue#dequeue` from blowing the stack

tamird:
In the case of many sequential jobs having been deleted, a recursive
implementation of `Queue#dequeue` is prone to blowing the stack in the
absence of tail-recursion support. Change the implementation from
recursive to iterative to work around this issue in CPython.


nvie:
Thanks!

PR #524, opened 2015-04-30
Enable proper setup of signale handlers for SimpleWorker as well.

### LINKED ISSUE #523: SimpleWorker warm shutdown
Looking into the code, for some unknown reason, the `worker.SimpleWorker` class overrides the definition for `_install_signal_handlers()`. The problem with this is that, when a SIGTERM is received, the worker gets killed instantly and does not finish the job or simply stays in the loop and keeps starting jobs, eventually getting killed mid-job when a second SIGTERM arrives.

To me, this behaviour is highly unintuitive and buggy, since it does not offer the possibility to exit the job loop in a clean way. Signal handlers are not useless because they also set the state of the worker in `stopped`, does cleanly breaking out of the cycle as soon as possible.

cosminstefanxp:
Fixes #523

This allows a SIGTERM to make the worker perform a warm shutdown and cleanly break out of the loop and finish the current job, if any.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -646,11 +646,6 @@ class Worker(object):
 
 
 class SimpleWorker(Worker):
-    def _install_signal_handlers(self, *args, **kwargs):
-        """Signal handlers are useless for test worker, as it
-        does not have fork() ability"""
-        pass
-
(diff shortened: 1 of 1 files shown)


selwin:
Thanks for the PR!

PR #530, opened 2015-05-14
updated worker API see issue 255

RyanMTB:
This is a fix to: https://github.com/nvie/rq/issues/255

Now all of these will work.

```python
queue_names = ['high', 'normal', 'default']
q = Queue(name = 'foo')
w = Worker(queue_names)
w = Worker('high')
w = Worker((map(Queue, queue_names))
w = Worker((list(map(Queue, queue_names)))
w = Worker(q, ['high', 'low'])
```

This is my first ever attempt at contributing to an open source project. So please give me all the feedback you have.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -70,6 +70,9 @@ class Queue(object):
     def __len__(self):
         return self.count
 
+    def __iter__(self):
+        yield  self
+
     @property
     def key(self):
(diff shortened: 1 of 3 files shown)


RyanMTB:
Looks Like I have some work to do to make it 2.x compatible.

I will give this a try soon.

PR #531, opened 2015-05-14
updated worker api see issue 255

RyanMTB:
This is a fix to: https://github.com/nvie/rq/issues/255

Now all of these will work.

```python
w = Worker('high')  

q = Queue(name = 'high')
w = Worker(q)

queue_names = ['high', 'normal', 'default']
w = Worker(queue_names)
w = Worker((map(Queue, queue_names))
w = Worker((list(map(Queue, queue_names)))

w = Worker([queue_names, q])
```

This is my first ever attempt at contributing to an open source project. All feedback is greatly appreciated.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -125,8 +125,7 @@ class Worker(object):
         if connection is None:
             connection = get_current_connection()
         self.connection = connection
-        if isinstance(queues, self.queue_class):
-            queues = [queues]
+        queues = self.process_queue_args(queues)
         self._name = name
         self.queues = queues
         self.validate_queues()
@@ -160,11 +159,18 @@ class Worker(object):
 
     def validate_queues(self):
         """Sanity check for the given queues."""
-        if not iterable(self.queues):
-            raise ValueError('Argument queues not iterable.')
         for queue in self.queues:
             if not isinstance(queue, self.queue_class):
-                raise NoQueueError('Give each worker at least one Queue.')
+                raise NoQueueError('{0} is not a queue'.format(queue))
+
+    def process_queue_args(self, queue_args):
+        """ allow for a string, a queue an iterable of strings
+        or an iterable of queues"""
+        if isinstance(queue_args, text_type):
+            return self.queue_class(name = queue_args)
+        else:
+            return  [self.queue_class(name=queue_arg) if isinstance(queue_arg, text_type) 
+                    else queue_arg for queue_arg in queue_args]
 
     def queue_names(self):
         """Returns the queue names of this worker's queues."""
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -70,6 +70,9 @@ class Queue(object):
     def __len__(self):
         return self.count
 
+    def __iter__(self):
+        yield  self
+
     @property
     def key(self):
         """Returns the Redis key for this Queue."""
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -12,6 +12,7 @@ from tests.helpers import strip_microseconds
 
 from rq import get_failed_queue, Queue, SimpleWorker, Worker
 from rq.compat import as_text
+from rq.exceptions import NoQueueError
 from rq.job import Job, JobStatus
 from rq.registry import StartedJobRegistry
 from rq.suspension import resume, suspend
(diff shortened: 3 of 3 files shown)


rq/worker.py
```diff
@@ -160,11 +161,22 @@ def __init__(self, queues, name=None,
...
-            raise ValueError('Argument queues not iterable.')
         for queue in self.queues:
             if not isinstance(queue, self.queue_class):
-                raise NoQueueError('Give each worker at least one Queue.')
+                raise NoQueueError('Give each worker at least one Queue.\
+                        {} was passed'.format(queue))
+
+    def process_queue_args(self, queue_args):
+        """ allow for a variety arguements to queues """
+        if isinstance(queue_args, six.text_type):
```
selwin: Can you please change this line to use `rq.compat.text_type` so we don't need need to depend on `six`?

selwin:
Thanks for the PR @RyanMTB ! I'll merge this in when we lose the `six` dependency.

rq/worker.py
```diff
@@ -160,11 +161,22 @@ def __init__(self, queues, name=None,
...
+            return self.queue_class(name = queue_args)
+        else:
+            queue_args = [queue_args] if isinstance(queue_args, six.text_type) else queue_args
+            queue_args = [[queue_arg] if isinstance(queue_arg, six.text_type) 
+                    else list(queue_arg) for queue_arg in queue_args]
+            queue_args = itertools.chain.from_iterable(queue_args)
+            return  [self.queue_class(name = queue_arg) if isinstance(queue_arg,six.text_type)
```
selwin: Nitpicks: could you please remove the spaces between the `=` in `self.queue_class(name = queue_arg)`? Also, put a space after comma in `isinstance(queue_arg,six.text_type)`

RyanMTB:
@selwin Fixed.  Thank you for the feedback and comments.

My pull request somehow now contains my upstream/master merge commit.  Let me know if that is a problem.  I am not sure.

rq/worker.py
```diff
@@ -160,11 +160,22 @@ def __init__(self, queues, name=None,
 
     def validate_queues(self):
         """Sanity check for the given queues."""
-        if not iterable(self.queues):
-            raise ValueError('Argument queues not iterable.')
         for queue in self.queues:
             if not isinstance(queue, self.queue_class):
-                raise NoQueueError('Give each worker at least one Queue.')
+                raise NoQueueError('Give each worker at least one Queue.\
```
selwin: I know this error message has been here from the very start. Doesn't `Give each worker at least one Queue` sound a bit strange? I think we should just change the error message to `{} is not a Queue`.

rq/worker.py
```diff
@@ -160,11 +160,21 @@ def __init__(self, queues, name=None,
...
             if not isinstance(queue, self.queue_class):
-                raise NoQueueError('Give each worker at least one Queue.')
+                raise NoQueueError('{} is not a queue'.format(queue))
+
+    def process_queue_args(self, queue_args):
+        """ allow for a variety arguements to queues """
+        if isinstance(queue_args, text_type):
+            return self.queue_class(name = queue_args)
+        else:
+            queue_args = [queue_args] if isinstance(queue_args, text_type) else queue_args
```
selwin: This line of code will never be used since we already checked for `text_text` instance in [Line 169](https://github.com/nvie/rq/pull/531/files#diff-1aac421e8c53e895ee40f3b7dc63144fR169)

RyanMTB: I think we still need it. queue_args could be a list and there is still a chance for the user to pass in something that is not a queue.

As the code is currently written it will flatten iterables that are nested one deep. (maybe this is unnecessary?)

```Python
Worker([['foo']]) #this will work
Worker([[['foo']]]) #this will not work
```
I would be happy to write more tests if that would clear things up.

RyanMTB:
I cleaned up my code so it will no longer flatten nested lists(unnecessary).  I also added a test to show that the following still works:

```Python
w = Worker(list(map(Queue, ['high', 'low']))
```

selwin:
It looks much simpler now, very nice! Thanks @RyanMTB!

PR #534, opened 2015-05-23
Worker automatically cleans job registries every hour

### LINKED ISSUE #519: The finished jobs list (rq:finished:) never gets cleaned up
First of all, great job with the project. It's awesome.

Now, to get to my problem. I'm using RQ in a high throughput system, with hundreds of thousands of jobs/day.  Even though we have a `result_ttl` set up and the keys for the jobs themselves get cleared out, the `rq:finished:low` ZSET which contains ids of finished jobs does not seem to be cleared. As a result, since we put the system in production (4 days ago), it has grown to ~3M entries and keeps growing. This is starting to use a lot of memory and has a visible effect on the performance.

@nvie

selwin:
Version 0.5.0 introduced `StartedJobRegistry` and `FinishedJobRegistry`. Unmaintained, job registries could grow to really large sizes. This is reported by @cosminstefanxp in #519 .

This PR introduces a feature where workers will clean up job registries every hour without having to run out of band maintenance scripts, simplifying worker deployments.

If there's no complaints, I'll merge this in a few days.

Fixes #519.

### DIFF
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -8,8 +8,9 @@ from .utils import current_timestamp
 
 class BaseRegistry(object):
     """
-    Base implementation of job registry, implemented in Redis sorted set. Each job
-    is stored as a key in the registry, scored by expiration time (unix timestamp).
+    Base implementation of a job registry, implemented in Redis sorted set.
+    Each job is stored as a key in the registry, scored by expiration time
+    (unix timestamp).
     """
 
     def __init__(self, name='default', connection=None):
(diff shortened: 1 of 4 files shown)


cosminstefanxp:
It looks really good. Thanks for this. I'd suggest though that you add a way of specifying the cleanup time. Something like a property of the Worker that can either be changed when extending the class or set at creation.

selwin:
I agree, I also thought about adding a user configurable "maintenance_interval" argument that can be specified during Worker.init() . That can be added in a later pull request.

nvie:
Thanks, @selwin, this is very well done! I've renamed the `maintenance_date` to `last_cleaned_at`, to reflect the nature of the state value a bit more accurately (see ec0e04727bb79ace3894ab084fa7293206fe98d7).

PR #539, opened 2015-05-28
Added test coverage for unicode keyword argument support

brownstein:
Added test coverage for unicode keyword argument support in method signatures (#536). Confirmed that this fails pre-536, and passes post-536.


selwin:
Thanks a lot for this test! Sorry for not contributing much, I'm not well versed in unicode stuff :\

PR #548, opened 2015-06-11
Refresh job data before sending job to exception handlers

olingerc:
Currently the job instance given to the exception handlers is the one created during enqueueing. This PR simply calls the refresh method to have the job data as it was at the moment the exception happened.
Someone using a custom exception handler can use for example job.meta to decide on how to handle exceptions.

Note: I did not immediately see a way to add this call to the pipeline since within refresh() redis data is needed.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -523,6 +523,7 @@ class Worker(object):
                 pipeline.execute()
 
             except Exception:
+                job.refresh()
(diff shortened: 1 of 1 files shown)


selwin:
Hello and thanks for the PR. I don't think it's necessary to add an extra Redis call to something which may or may not be needed by the exception handler. If needed, the custom exception handler can call `job.refresh()` to decide on how to handle the exception.

Issue #241, opened 2013-08-07
Feature request: pipelined enqueues and fetches


sylvinus:
We often need to queue thousands of tasks, so we wrote a wrapper to queue them all in the same pipeline to get reasonable performance.

It looks like this (send_task() is basically a customized Queue.enqueue_call()) :

```
def send_tasks(task, params_list, **kwargs):

  pipeline = get_redis_connection().pipeline(transaction=False)
  kwargs["connection"] = pipeline

  jobs = [send_task(task, params, **kwargs) for params in params_list]

  pipeline.execute()

  return jobs
```

It is pretty simple and could be useful to others if it was bundled in some way in RQ for instance in an "Queue.enqueue_calls" method.

On the other hand, we also need to fetch the statuses of many jobs at once. In this case RQ could use a small refactor in Job.refresh() to split the hgetall() call from the parsing and unpickling. Here is the code we use as an example:

```
def get_tasks(tasks_ids):
  pipeline = get_redis_connection().pipeline(transaction=False)

  jobs = [Job(jobid, connection=get_redis_connection()) for jobid in tasks_ids]

  for job in jobs:
    pipeline.hgetall(job.key)

  values = pipeline.execute()

  for i, obj in enumerate(values):
    if len(obj) == 0:
      jobs[i] = None
    else:
      # TODO other used attributes
      jobs[i]._status = obj.get('status') if obj.get('status') else None

  return jobs
```

That's it, I hope it is helpful to others, would love to see this integrated in RQ in some way.

jhorman:
Counter argument, against it not being a real problem in practice. I am also trying to queue thousands of jobs, but I am using a redis hosting provider with seemingly high latency between my data center and theirs. We are pipelining everything, which removes most of the trouble. `enqueue_many` would allow us to do this. Another option would be to pass `pipeline` into the delay call as a kwarg, though that is exposing probably too much detail.

I am thinking for now, of creating a new queue object on the fly pointing to a `pipeline` for `connection` but it scares me a bit in that there could be places that are expecting to get a synchronous response back.

selwin:
Mind making a PR to propose an implementation? Bulk enqueue is something I'd like to have.

PR #553, opened 2015-06-26
Adds support for batch writing jobs to redis via pipelines. 

jhorman:
Example:

``` python
with queue.batch() as batch:
  batch.enqueue(fn, 'test')
```

(this is in response to issue #241)


selwin:
Thanks for the PR and sorry for the delayed response.

I think we should support enqueueing multiple jobs without context managers. I'd like to enable a behavior similar to `autocommit` in database adapters like `psycopg2` where you can do something like:

```python
queue = Queue('default', autocommit=False)
job = queue.enqueue(foo) # Doesn't commit to Redis, perhaps this returns a job with id = None?
job_2 = queue.enqueue(foo)
queue.commit() # Commits all jobs to Redis
```
With this in place, we can then create a context manager that will commit deferred jobs on exit:
```python
with queue.batch():
    job = queue.enqueue(foo)
```
This should be achievable by storing `args` and `kwargs` passed to `enqueue` on thread local and enqueueing it in one batch when `queue.commit()` is called.

What do you think?

jhorman:
I don't love configuring `autocommit` mode on the queue itself. The queues I am using are initialized once, and are used in various contexts. Or, I guess, I would support it, in addition to being able to start a batch mode on an existing queue. I believe that would be the equivalent of starting a pyscopg2 transaction (possibly `transaction` is a better name too?). 

I guess another model, which is sort of like pipelines, or cursors, would be to have "batch" return a queue like object which is now in `autocommit` false mode.

```python
batch = queue.batch()
batch.enqueue()
batch.enqueue()
batch.commit()

with queue.batch() as batch:
  batch.enqueue()
```

What do you think of that?

selwin:
Yeah, I think you should be able to start a batch mode regardless of the initial queue configuration.

I don't mind `transaction` because it's a common terminology used by database adapters. Does `commit` sound better with `batch` though? Not a native English speaker here, I think in the English sense, committing a batch of jobs sounds better than committing a transaction.

jhorman:
Hmm. `commit` is the term used for transactions too. I do think saying transaction maybe implies more than is actually happening here. `batch` reflects the simplicity of the implementation. I will code up a version of `.batch` and `autocommit=False` and update this pull request to see what you think.

selwin:
@jhorman cool, thanks :)

selwin:
@jhorman reminder to update this Pr when you have the time :)

jhorman:
It has been a while. Sorry. What do you think of this simple approach. I didn't do the autocommit boolean, just a queue.batch() method, which can be used directly, or in as a context manager.

jhorman:
I maybe should have just killed the original branch, and created a new PR. I reset it, made changes, and force pushed it.

selwin:
@jhorman thanks for the update. I'll try to take a look at it this weekend.

rq/queue.py
```diff
@@ -419,6 +423,33 @@ def __str__(self):
         return '<Queue {0!r}>'.format(self.name)
 
 
+class Batch(Queue):
+    def __init__(self, name='default', default_timeout=None, connection=None,
```
selwin: I think it would be simpler to take a single `queue` argument instead of multiple arguments like `name` and `connection`. We'll be able to infer more information from the `queue` instance should we need more information about this queue in the future.

What do you think @jhorman?

selwin:
@jhorman what do you think of using `queue.pipeline()` instead of `queue.batch()`? I think `pipeline` is also a good description for this functionality.

Also, would it be possible to refactor the `Batch` class into a separate file? The current `queue.py` is big enough as it is.

Docs for this functionality would also be a welcome addition :)

theodesp:
@jhorman It looks like `Batch` is a one level of abstraction on top of the conn.pipeline.

In order to proceed with this PR you need to update the Docs, Resolve the code review requests

jhorman:
Sorry, I ended up moving on from this. Going to close this PR, as it is also really old.

Issue #558, opened 2015-07-07
perform_job leaves worker current_job set after exception handling


grzaks:
The workers `perform_job` function has the `try ... catch ...` block for exception handling but when the job function throws an exception the `self.set_current_job_id(None, pipeline=pipeline)` is never called. 

This leaves the worker in weird state where it's waiting for new job but still has `current_job` ID set in redis.

The `set_current_job_id(None, ...)` should be either in the `catch` or `finally` block IMHO.

https://github.com/nvie/rq/blob/master/rq/worker.py#L564

selwin:
Yeah, sounds like a bug to me. Pull request welcome :)

selwin:
Fixed in https://github.com/nvie/rq/pull/637

PR #560, opened 2015-07-07
worker left with current_job set after job exception

grzaks:
Issue #558 fix


selwin:
Thanks dor the PR! Can you add a small testcase for this?

selwin:
@grzaks reminder to update this PR with a testcase when you have the time :)

selwin:
Fixed in https://github.com/nvie/rq/pull/637

Issue #512, opened 2015-03-27
depends_on job queue always used


caseyjlaw:
When submitting a job with a dependency, the queue of the dependency seems to always be used. For example:

    with Connection():
        q = Queue('high')
        jobh = q.enqueue_call(func, args)
        q = Queue('low')
        jobl = q.enqueue_call(func, args, depends_on=jobh)

In my small tests, the 'jobl' is submitted to the 'high' queue. I am somewhat new to rq, but that behavior was unexpected.

caseyjlaw:
Sorry, I just saw that this issue is well known (https://github.com/nvie/rq/issues/387). Closing...

selwin:
I'm reopening this as we now have the building blocks in place to allow job dependency to be used across different queues (in the same Redis DB).

caseyjlaw:
Nice to hear! Let me know if I can help.
  Do I understand correctly that using multiple Redis DBs and connections would get around this issue? For example, the following should work?
    with Connection(Redis('localhost', 6379)):
        q = Queue('high')
        jobh = q.enqueue_call(func, args)
        with Connection(Redis('localhost', 6380)):
            q = Queue('low')
            jobl = q.enqueue_call(func, args, depends_on=jobh)

caseyjlaw:
Nice to hear! Let me know if I can help.
Do I understand correctly that using multiple Redis DBs and connections would get around this issue? For example, the following should work?

    with Connection(Redis('localhost', 6379)):
        q = Queue('high')
        jobh = q.enqueue_call(func, args)
        with Connection(Redis('localhost', 6380)):
            q = Queue('low')
            jobl = q.enqueue_call(func, args, depends_on=jobh)

selwin:
No, it will not work on different Redis DBs.

jlopex:
@selwin any chance to look at the PR I proposed?

jlopex:
@selwin Should we close this?

PR #564, opened 2015-07-21
support multiple queues on dependent jobs

jlopex:
Current implementation is using the first job queue to insert all dependent jobs, ignoring the queue where the jobs were enqueued.

With this change, we will use job origin attribute to fetch the original queue, then insert the dependent job to the fetched queue.

fixes issue #512

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -265,21 +265,21 @@ class Queue(object):
 
         If Queue is instantiated with async=False, job is executed immediately.
         """
-        if not pipeline:
-            pipeline = self.connection._pipeline()
+        pipe = pipeline if pipeline is not None else self.connection._pipeline()
 
         # Add Queue key set
-        self.connection.sadd(self.redis_queues_keys, self.key)
-        job.set_status(JobStatus.QUEUED, pipeline=pipeline)
+        pipe.sadd(self.redis_queues_keys, self.key)
+        job.set_status(JobStatus.QUEUED, pipeline=pipe)
 
         job.origin = self.name
         job.enqueued_at = utcnow()
 
         if job.timeout is None:
             job.timeout = self.DEFAULT_TIMEOUT
-        job.save(pipeline=pipeline)
+        job.save(pipeline=pipe)
 
-        pipeline.execute()
+        if pipeline is None:
+            pipe.execute()
 
         if self._async:
             self.push_job_id(job.id, at_front=at_front)
@@ -290,8 +290,6 @@ class Queue(object):
         """Enqueues all jobs in the given job's dependents set and clears it."""
         # TODO: can probably be pipelined
         from .registry import DeferredJobRegistry
-
-
 
         while True:
             job_id = as_text(self.connection.spop(job.dependents_key))
@@ -306,6 +304,7 @@ class Queue(object):
                 else:
                     queue = Queue(name=dependent.origin, connection=self.connection)
                     queue.enqueue_job(dependent, pipeline=pipeline)
+                pipeline.execute()
 
     def pop_job_id(self):
         """Pops a given job ID from this Redis queue."""
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -260,24 +260,26 @@ class Queue(object):
                                  description=description, depends_on=depends_on,
                                  job_id=job_id, at_front=at_front)
 
-    def enqueue_job(self, job, at_front=False):
+    def enqueue_job(self, job, pipeline=None, at_front=False):
         """Enqueues a job for delayed execution.
 
         If Queue is instantiated with async=False, job is executed immediately.
         """
-        with self.connection._pipeline() as pipeline:
-            # Add Queue key set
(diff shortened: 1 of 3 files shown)


rq/queue.py
```diff
@@ -289,15 +289,20 @@ def enqueue_dependents(self, job):
...
             job_id = as_text(self.connection.spop(job.dependents_key))
             if job_id is None:
                 break
             dependent = self.job_class.fetch(job_id, connection=self.connection)
+            registry = DeferredJobRegistry(dependent.origin, self.connection)
             registry.remove(dependent)
-            self.enqueue_job(dependent)
+
+            # Get the correct queue for the dependent job origin
+            q = self.from_queue_key(as_text(
```
selwin: I think this is better:
```python
if dependent.origin == self.name:
    self.enqueue_job(dependent)
else:
    queue = Queue(name=dependent.origin, connection=self.connection)
    queue.enqueue_job(dependent)
```

rq/queue.py
```diff
@@ -289,15 +289,20 @@ def enqueue_dependents(self, job):
...
-        registry = DeferredJobRegistry(self.name, self.connection)
+
 
         while True:
             job_id = as_text(self.connection.spop(job.dependents_key))
             if job_id is None:
                 break
             dependent = self.job_class.fetch(job_id, connection=self.connection)
+            registry = DeferredJobRegistry(dependent.origin, self.connection)
             registry.remove(dependent)
```
selwin: Can you also pipeline the `registry.remove` and `queue.enqueue_job` methods? In addition to being more performant, this also protects us from having the job already removed `DeferredJobRegistry`, but not properly enqueued in case something fails (usually network).

jlopex:
@selwin I made the requested changes to pipeline the Redis calls.

But I'm not sure if I understand correctly the pipeline changes, could you take a look to it?

rq/queue.py
```diff
@@ -289,15 +289,22 @@ def enqueue_dependents(self, job):
...
             if job_id is None:
                 break
             dependent = self.job_class.fetch(job_id, connection=self.connection)
-            registry.remove(dependent)
-            self.enqueue_job(dependent)
+            registry = DeferredJobRegistry(dependent.origin, self.connection)
+            with self.connection._pipeline() as pipeline:
+                registry.remove(dependent, pipeline=pipeline)
+                if dependent.origin == self.name:
+                    self.enqueue_job(dependent)
```
selwin: Thanks for the update! Do you mind adding a `pipeline` argument to `enqueue_job` so we can do all these commands in one network call?

Thanks again :)

selwin: Please update this when you have the time @jlopex :)

jlopex: I just came back from holidays. Expect a PR in a few minutes.

jlopex:
@selwin Done! Please let me know if you see other issues.

rq/queue.py
```diff
@@ -260,24 +260,26 @@ def enqueue(self, f, *args, **kwargs):
...
-            # Add Queue key set
-            self.connection.sadd(self.redis_queues_keys, self.key)
-            job.set_status(JobStatus.QUEUED, pipeline=pipeline)
+        if not pipeline:
+            pipeline = self.connection._pipeline()
 
-            job.origin = self.name
-            job.enqueued_at = utcnow()
+        # Add Queue key set
+        self.connection.sadd(self.redis_queues_keys, self.key)
```
selwin: This call should be pipelined.

rq/queue.py
```diff
@@ -260,24 +260,26 @@ def enqueue(self, f, *args, **kwargs):
...
-            job.save(pipeline=pipeline)
+        job.origin = self.name
+        job.enqueued_at = utcnow()
 
-            pipeline.execute()
+        if job.timeout is None:
+            job.timeout = self.DEFAULT_TIMEOUT
+        job.save(pipeline=pipeline)
+
+        pipeline.execute()
```
selwin: If `pipeline` is supplied as an argument, we shouldn't call `execute`, leaving the `execute` call to be called from the place where this method is called.

rq/queue.py
```diff
@@ -289,15 +291,21 @@ def enqueue_dependents(self, job):
...
-            registry.remove(dependent)
-            self.enqueue_job(dependent)
+            registry = DeferredJobRegistry(dependent.origin, self.connection)
+            with self.connection._pipeline() as pipeline:
+                registry.remove(dependent, pipeline=pipeline)
+                if dependent.origin == self.name:
+                    self.enqueue_job(dependent, pipeline=pipeline)
+                else:
+                    queue = Queue(name=dependent.origin, connection=self.connection)
+                    queue.enqueue_job(dependent, pipeline=pipeline)
```
selwin: We should call `pipeline.execute()` inside this context.

rq/queue.py
jlopex: OK!

jlopex:
@selwin here you are!

selwin:
Thanks for the PR!

PR #573, opened 2015-08-13
dequeue_any not recursive

jcsaaddupuy:
Currently, `rq.queue.dequeue_any` is prone to maximum recursion depth exceeded errors.

This is due to a recursive call done whenever a [`NoSuchJobError`]( https://github.com/nvie/rq/blob/master/rq/queue.py#L378-L381) occurs.

This PR try to fixes this by removing the recursive call and uses a [`while True`](https://github.com/jcsaaddupuy/rq/blob/feature/dequeue_any_not_recursive/rq/queue.py#L370-L390) loop instead, as it is done in the [`rq.queue.dequeue`](https://github.com/nvie/rq/blob/master/rq/queue.py#L341-L349) method.


selwin:
Thanks for the PR!

PR #578, opened 2015-08-31
log job result in DEBUG level and truncate to 2k symbols

tbaugis:
I think it makes sense moving output to DEBUG as you'd care about it in logs only while debugging. Over certain size the output in logs also gets quite unwieldy, so truncating it to 2k chars.

Keep up the good work and let me know if things should be done differently!

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -602,8 +602,6 @@ class Worker(object):
         self.log.info(green('Job OK'))
         if rv:
             log_result = "{0!r}".format(text_type(rv))
-            if len(log_result) > 2000:
-                log_result = "{0}..\n..output truncated".format(log_result[:2000])
             self.log.debug('Result: {0}'.format(yellow(log_result)))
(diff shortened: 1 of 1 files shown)


selwin:
@tbaugis thanks for the PR!

I like the logging level changes but this is not the right place to truncate logging output. Truncating output should on the logging formatter level. If we truncate the output at this level, you won't be able to log the whole output even if you wanted to.

tbaugis:
I see, thanks for getting back!

Would you mind simply changing the log level to DEBUG then, or should i make a one-liner pull request?

selwin:
Please keep the current code you have now, but don't truncate the result to 2k characters :). So `Job Ok` will still be logged on `info`, but the `result` will be only be logged on `debug`.

tbaugis:
updated as asked!

selwin:
Thanks!

Issue #479, opened 2015-02-02
get_current_job() throws NoRedisConnectionException in 0.5.0


cbenhagen:
While this works in 0.4.6:
```python
from rq import get_current_job

def add(x, y):
    job = get_current_job()
    print 'Current job: %s' % (job.id,)
    return x + y
```

on 0.5.0 this throws an error:
```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/rq/worker.py", line 543, in perform_job
    rv = job.perform()
  File "/usr/local/lib/python2.7/site-packages/rq/job.py", line 490, in perform
    self._result = self.func(*self.args, **self.kwargs)
  File "./test.py", line 4, in add
    job = get_current_job()
  File "/usr/local/lib/python2.7/site-packages/rq/job.py", line 81, in get_current_job
    return Job.fetch(job_id, connection=connection)
  File "/usr/local/lib/python2.7/site-packages/rq/job.py", line 293, in fetch
    job = cls(id, connection=connection)
  File "/usr/local/lib/python2.7/site-packages/rq/job.py", line 298, in __init__
    self.connection = resolve_connection(connection)
  File "/usr/local/lib/python2.7/site-packages/rq/connections.py", line 70, in resolve_connection
    raise NoRedisConnectionException('Could not resolve a Redis connection.')
NoRedisConnectionException: Could not resolve a Redis connection.
```

This works:
```python
from rq import get_current_job, Connection

def add(x, y):
    with Connection():
        job = get_current_job()
        print 'Current job: %s' % (job.id,)
        return x + y
```
But in https://github.com/nvie/rq/issues/164#issuecomment-12422939 @nvie wrote you shouldn't do that.

nvie:
Whoops, that _is_ a regression. You initial example _should_ work. We should fix this in 0.5.1.

If you're not too worried about consuming more Redis connections, you can probably still use the latter example for now, but because you can potentially exhaust your number of allowed connections to Redis this way, you should not.

mehaase:
Hey @nvie, did this get fixed in 0.5.1?

I'm getting an error, but I'm not sure if its a bug in RQ or if I'm doing something wrong. I want to use `get_current_job()` so that the function executing a task can read/edit the job's metadata. I'm trying something like this, but it's not working:

```python
from rq import get_current_job

def foo(bar):
    job = get_current_job()
    job.meta['baz'] = 'bat'
    job.save()

    return ''.join(reversed(bar))
```

```python
[...]

PR #581, opened 2015-09-22
export worker id and job id as env vars to work horse

mgk:
This was initially spurred by #479. Idea behind doing this:

 + including worker id and job id in logs and relevant HTTP request headers
 + no additional redis connection or request required
 + no API change: workers can look for env vars and use them if found

Environment variables can be abused much as global variables often are, but they feel like a good fit here for communicating contextual read-only information from the parent worker to the child work horse process.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -491,6 +491,8 @@ class Worker(object):
         within the given timeout bounds, or will end the work horse with
         SIGALRM.
         """
+        os.environ['RQ_WORKER_ID'] = self.name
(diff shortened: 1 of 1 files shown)


mgk:
Any thoughts on this one?

selwin:
@mgk sorry I missed this PR. I think this is a good idea. Could you please update the PR so that it merges cleanly into master?

mgk:
No worries. Updated.

Issue #584, opened 2015-10-01
warm shutdown on heroku and long jobs


samuelcolvin:
We have some jobs which run on a heroku worker and take quite a long time lets say 1-10 minutes.

As per [this](http://support.hirefire.io/discussions/problems/31480-valid-jobs-being-moved-to-failed-in-rq-on-warm-shutdown) or due to heroku restarting dynos these jobs occasionally recieve the `SIGTERM` signal and then shutdown.

Since the jobs are unlikely to end in the 10 second grace period Heroku give before killing the dyno the right thing to do here would be to raise an Exception and let the job fail. Then our standard code for recording and dealing with errors would work.

I've tried adding a handler to `SIGTERM` to do this however it's not being called because of [_install_signal_handlers](https://github.com/nvie/rq/blob/bb4336f612910f33ed72359a3f8b690b4acf8a4a/rq/worker.py#L321-L366). Because the jobs are eventually killed by the worker shutting down we're not getting an exception raised or sent to sentry, similarly the job doesn't seem to appear in the failed queue until it's exceed it's timeout (or perhaps "cleanup" deals with it, I don't know).

What's the best solution here?

All I can think of here is to create a custom worker and overwrite `_install_signal_handlers`. If that's the correct solution can I suggest defining `request_force_stop` and `request_stop` outside `_install_signal_handlers` so less code has to be duplicated?

samuelcolvin:
I've looked into this in more detail main problem (and I think it's quite a big problem) is [this](https://github.com/nvie/rq/blob/0.5.5/rq/worker.py#L528).

The work horse is terminated as soon as it receives `SIGTERM`. Heroku sends SIGTERM to all processes on the dyno 10 seconds before it shuts it down see [here](https://devcenter.heroku.com/articles/dynos#shutdown). 

> This signal is sent to all processes in the dyno, not just the process type.

This means on Heroku that none of rq's warm/cold shutdown, wait for jobs to finish logic is being called at all - as soon as a dyno starts to shut down all jobs are killed immediately with no exception or traceback.

My personal solution is to send `SIGRTMIN` to `horse_pid` in `request_stop`, that way most jobs won't be affected and might well finish before the 10 second grace period runs out, but processes which won't finished can handle `SIGRTMIN` and throw an exception, we can even use `frame` to get some information about the state of the job when the signal was received. 

[...]

PR #587, opened 2015-10-14
Signal changes

samuelcolvin:
ref #584, this doesn't fix that problem but it should make implementing a better Heroku worker easier, I think it should also improve coverage since `_install_signal_handlers`, `request_force_stop` and `request_stop` are currently not tested at all.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -323,47 +323,47 @@ class Worker(object):
         gracefully.
         """
 
-        def request_force_stop(signum, frame):
-            """Terminates the application (cold shutdown).
-            """
-            self.log.warning('Cold shut down')
+        signal.signal(signal.SIGINT, self.request_stop)
+        signal.signal(signal.SIGTERM, self.request_stop)
 
-            # Take down the horse with the worker
-            if self.horse_pid:
-                msg = 'Taking down horse {0} with me'.format(self.horse_pid)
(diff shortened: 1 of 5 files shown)


rq/worker.py
```diff
@@ -323,47 +323,47 @@ def _install_signal_handlers(self):
...
-                               'Press Ctrl+C again for a cold shutdown.')
-            else:
-                raise StopRequested()
-
-        signal.signal(signal.SIGINT, request_stop)
-        signal.signal(signal.SIGTERM, request_stop)
+        signal.signal(signal.SIGINT, self.request_stop)
+        signal.signal(signal.SIGTERM, self.request_stop)
+
+    def request_force_stop(self, signum, frame):
```
samuelcolvin: no changes here other than moving these two functions out of `_install_signal_handlers` so they can be overridden independently.

run_tests
```diff
@@ -17,9 +17,9 @@ else
     safe_rg=cat
 fi
 
-export ONLY_RUN_FAST_TESTS=1
+export RUN_SLOW_TESTS_TOO=1
```
samuelcolvin: this variable's name seemed to indicate it did the opposite of what it actually did, hence the switch.

selwin:
Thanks for the PR!

PR #588, opened 2015-10-14
more info in 'Job OK' log entry

samuelcolvin:
Changed to mirror the [job started](https://github.com/nvie/rq/blob/master/rq/worker.py#L461-L462) log entry.

The current simple "Job OK" message is pretty useless if you have multiple queues executing jobs, it can take ages work out which job was ok and it's impossible if you have multiple workers running and can't differentiate between their logs.


selwin:
Thanks for the PR!

PR #589, opened 2015-10-14
add job.started_at

samuelcolvin:
currently only `enqueued_at` and `ended_at` are available which means it's not possible to calculate either how long the job was in the queue for or how long it took to be performed. Adding the time the job was started at means both these numbers can be calculated.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -564,6 +564,7 @@ class Worker(object):
             started_job_registry = StartedJobRegistry(job.origin, self.connection)
 
             try:
+                job.started_at = utcnow()
                 with self.death_penalty_class(job.timeout or self.queue_class.DEFAULT_TIMEOUT):
                     rv = job.perform()
 
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -549,6 +549,7 @@ class Worker(object):
             registry = StartedJobRegistry(job.origin, self.connection)
             registry.add(job, timeout, pipeline=pipeline)
             job.set_status(JobStatus.STARTED, pipeline=pipeline)
+            job.set_started_at_now(pipeline=pipeline)
             pipeline.execute()
 
         msg = 'Processing {0} from {1} since {2}'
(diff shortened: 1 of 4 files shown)


XayOn:
+1 to this one.

tests/test_worker.py
```diff
@@ -83,6 +83,29 @@ def test_work_via_string_argument(self):
...
+        job = q.enqueue(say_hello)
+        self.assertIsNotNone(job.enqueued_at)
+        self.assertIsNone(job.started_at)
+        self.assertIsNone(job.ended_at)
+        self.assertEquals(w.work(burst=True), True,
+                          'Expected at least some work done.')
+        self.assertEquals(job.result, 'Hi there, Stranger!')
+        after = utcnow()
+        job.refresh()
+        self.assert_dt_between(job.enqueued_at, before, after)
```
selwin: I'm not a big fan of using a method to replace a one liner. Mind if we just simplify this to `self.assertTrue(before <= job.enqueued_at <= after)`?

samuelcolvin: absolutely, i thought this was going to do more than it ended up doing, also the method is ugly.

rq/worker.py
```diff
@@ -564,6 +564,7 @@ def perform_job(self, job):
             started_job_registry = StartedJobRegistry(job.origin, self.connection)
 
             try:
+                job.started_at = utcnow()
```
selwin: Could you please move this assignment to `prepare_job_execution`? This will persist the information into Redis along with `job.status`.

ducu:
Yes +1 from me.

I'm doing this already inside the job, and also storing `ended_at` but now I see this was already there.
The point is to calculate how long a job takes.

It would be great to display few stats about this in rq-dashboard as well, I'm looking into this right now.

selwin:
ping @samuelcolvin :)

samuelcolvin:
Sorry, sorry, sorry; life got in the way I forgot this.

I believe it should be passing now, the failure is due to the latest version of click released today, see https://github.com/mitsuhiko/click/issues/470. I'll rebuild when that's fixed.

samuelcolvin:
# done

sorry for delay.

selwin:
Thanks!

PR #594, opened 2015-10-28
mark as busy before forking horse

horida:
Set the worker to busy before forking the horse. Right now only the fork is set to busy and SIGINT will kill the worker without waiting for the horse to finish.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -491,6 +491,7 @@ class Worker(object):
         within the given timeout bounds, or will end the work horse with
         SIGALRM.
         """
+        self.set_state('busy')
         child_pid = os.fork()
         if child_pid == 0:
             self.main_work_horse(job)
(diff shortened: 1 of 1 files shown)


selwin:
@horida I think you could simply move the line in https://github.com/nvie/rq/blob/master/rq/worker.py#L502 to the top of `execute_job`. That way, we won't have redundant status setting statements. Mind updating the PR?

Thanks!

horida:
@selwin done. Sorry for the mess, I'm in a hurry and wanted to get it done as soon as possible :(

rq/worker.py
```diff
@@ -429,6 +429,7 @@ def work(self, burst=False):
 
                 job, queue = result
                 self.execute_job(job)
+                self.set_state('idle')
```
selwin: This line isn't needed either, because the idle state would have been set in https://github.com/nvie/rq/pull/594/files#diff-1aac421e8c53e895ee40f3b7dc63144fR505 :)

selwin:
Thanks for the PR!

PR #596, opened 2015-11-02
[Docs] Prevent running worker when the connection instance is imported

kevgathuku:
Heroku Example:
When importing the redis connection instance from another file,
the worker is run immediately, blocking the execution of the rest of
the code.
This fixes the worker to only be run only if the worker is executed from
the command line and not when imported.
[Sample Use Case](https://devcenter.heroku.com/articles/python-rq)


selwin:
Thanks for the PR!

kevgathuku:
:+1:

PR #599, opened 2015-11-06
rqinfo config parameter

kfrendrich:
I had a problem with the rqinfo command, because I keep the rq in the 1st database instead of 0.
And because of this the rqinfo command can't show me any  information.
This patch is allows me to use config to get the rq informations from the 1st db.


selwin:
Thanks for the PR!

Issue #597, opened 2015-11-03
Long cleanup for large registries


glaslos:
I just upgraded to rq-0.5.6 from pre-registry-cleanup times. One of my workers never started actually working. Investigating the issue uncovered that the StartedJobRegistry had more than a million items (I have large queues and many tasks). Cleanup takes some time and I should have just waited a bit longer.

Would it make sense to chop the cleanup task into junks of hundreds to prevent long blocking of the worker? Or at least add a log message that the working is currently cleaning up?

selwin:
I think adding logging message would be good. Could you please make a PR for this?

The problem you ran into is a one time issue caused by upgrading from a previous version. I *think* it shouldn't affect normal operations because high trafficked queues will likely have many workers so cleanups will be done in chunks by many workers.

If my assumption is incorrect, we can always revisit this issue :)

glaslos:
PR created: #601
Your assumption seems to be correct as I don't experience any issues since the first cleanup.

PR #601, opened 2015-11-06
log.info added when cleaning registries

glaslos:
See discussion: #597


glaslos:
The failed Travis test seems to be a unrelated timing issue: https://travis-ci.org/nvie/rq/jobs/89649275

Issue #528, opened 2015-05-06
Allow setting meta in enqueue_call


johtso:
It would be rather convenient to be able to define job meta information in a keyword argument to `enqueue_call`, unless I'm missing something and it's already possible!

selwin:
Yes, ai think this is a reasonable request. Feel free to open a PR for this.

johtso:
Good stuff, I'll see if I can put something together.

selwin:
Fixed in https://github.com/nvie/rq/pull/609

PR #609, opened 2015-11-16
Allow meta when enqueing

tornstrom:
Makes it very easy to add meta data to a Job in the Queue.enqueue call. Currently you would have to create the Job separately, set its metadata and then call enqueue_job. This way you only need to provide the 'meta' keyword arg to the enqueue call and you're set.

This was mentioned in issue #528 too.

Let me know what you think.

Thanks

-Tim

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -89,7 +89,7 @@ class Job(object):
     @classmethod
     def create(cls, func, args=None, kwargs=None, connection=None,
                result_ttl=None, ttl=None, status=None, description=None,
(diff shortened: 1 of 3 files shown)


rblakemesser:
That's super smart. They'll probably want some test coverage-- looks like their test suite is pretty clean/extensible.

selwin:
@tornstrom thanks for the PR! Could you please modify one of the existing testcase to verify that this works?

tornstrom:
@selwin Will do, thanks for taking a look!

selwin:
Thanks!

PR #614, opened 2015-11-24
Round Robin Queue selection by worker

srajbr:


nvie:
This goes straight against the RQ philosophy that queue priority is defined by the startup argument order passed to the worker command. Technically there is no way to consider two queues of equal importance in RQ. If your jobs demand this, consider writing them to the same queue.

Issue #514, opened 2015-04-03
RQ doesn't support FakeRedis


jmmills:
RQ doesn't seem to support using FakeRedis to mock redis connection, thus disallowing unit tests sans redis-server.

    E
    ======================================================================
    ERROR: rq_fakeredis.test_rq_with_fakeredis
    ----------------------------------------------------------------------
    Traceback (most recent call last):
      File "/Users/jmills/virtualenv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
        self.test(*self.arg)
      File "/Users/jmills/rq_fakeredis.py", line 8, in test_rq_with_fakeredis
        queue = Queue(connection=redis)
      File "/Users/jmills/virtualenv/lib/python2.7/site-packages/rq/queue.py", line 58, in __init__
        self.connection = resolve_connection(connection)
      File "/Users/jmills/virtualenv/lib/python2.7/site-packages/rq/connections.py", line 66, in resolve_connection
        return patch_connection(connection)
      File "/Users/jmills/virtualenv/lib/python2.7/site-packages/rq/compat/connections.py", line 23, in patch_connection
        raise ValueError('A StrictRedis or Redis connection is required.')
    ValueError: A StrictRedis or Redis connection is required.
    
    ----------------------------------------------------------------------
    Ran 1 test in 0.001s
    
    FAILED (errors=1)

[rq_fakeredis.py | uploaded via ZenHub](https://files.zenhub.io/551eedd312f7fefa28fc43a7)

jonafato:
Duplicate of #282

jmmills:
@jonafato yeah, but it was closed and doesn't seem to point to a resolution.

jonafato:
It was still open when I made that comment :). It's the more active issue, so if more discussion is to happen, it should probably take place there and possibly reopen the issue if there is a path to a fix.

robodude666:
Maybe the official [testing](http://python-rq.org/docs/testing/) simply need to include an example of how to use `mock` with `rq`? There are already a few examples within #282 and googling shows similar solutions; they just need to be documented so people don't think it's a "bug" of `rq`.

selwin:
I think we can modify `patch_connection` to this:
```python
def patch_connection(connection):
    # Don't patch already patches objects
    PATCHED_METHODS = ['_setex', '_lrem', '_zadd', '_pipeline', '_ttl']
    if all([hasattr(connection, attr) for attr in PATCHED_METHODS]):
        return connection

[...]

Issue #282, opened 2013-11-14
Can not mock rq's redis connection with mockredis


zgoda:
Because of check if connection is instance of `StrictRedis` in `rq.compat.connection.patch_connection()`. Is this check really necessary?

yrik:
@robodude666 could you show how you patch `patch_connection `?
I have the same issue..

homeworkprod:
I struggle with this as well. Bummer.

homeworkprod:
For my simple use case – avoiding the need for a running Redis server during tests – without any mock call verifications, this seems to work (Python 3.4.2, redis 2.10.3, rq 0.5.2):

```python
from unittest.mock import MagicMock

from redis import StrictRedis


class StrictRedisMock(MagicMock, StrictRedis):

    @classmethod
    def from_url(cls, arg):
        return cls()
```

The code under test creates `StrictRedis` instances via the `from_url` method. YMMV.

Apply like this:
```python

class SomeTest(SomeTestBase):

    @patch('redis.StrictRedis.from_url', StrictRedisMock.from_url)
    def setUp(self):
        super().setUp()

        # …
```

HTH

homeworkprod:
For the record, I switched from the mock class using diamond inheritance to this:

```python
def strict_redis_client_from_url(url):
    mock = MagicMock()
    mock.__class__ = StrictRedis
    return mock
```

Usage:
```python
class SomeTest(SomeTestBase):

    @patch('redis.StrictRedis.from_url', strict_redis_client_from_url)
    def setUp(self):
        super().setUp()

        # …
```

That said, my two suggested approaches might have serious limitations I haven't (yet) hit.

nvie:
I wish there was a better way for me to check this in `patch_connection`, but I don't think there is another way of distinguishing between `redis.StrictRedis` and `redis.Redis` other than using explicit `isinstance()` checks. If anyone has a better suggestion, I'd like to hear it.

Issue #88, opened 2012-07-16
Requeueing job resets timeout to 180


gperetin:
Is this intentionally or a bug? I have this fixed (preserving timeout), along with tests, so if it's a bug, I can send pull request.

PR #619, opened 2015-11-25
Fixes #613

### LINKED ISSUE #613: Not compatible with fakeredis
RQ does not work with [FakeRedis][1], due to [this][2]

```
  File "hello.py", line 4, in <module>
    q = Queue(connection=r)
  File "/home/vagrant/.virtualenvs/vagrant/lib/python3.4/site-packages/rq/queue.py", line 58, in __init__
    self.connection = resolve_connection(connection)
  File "/home/vagrant/.virtualenvs/vagrant/lib/python3.4/site-packages/rq/connections.py", line 66, in resolve_connection
    return patch_connection(connection)
  File "/home/vagrant/.virtualenvs/vagrant/lib/python3.4/site-packages/rq/compat/connections.py", line 23, in patch_connection
    raise ValueError('A StrictRedis or Redis connection is required.')
ValueError: A StrictRedis or Redis connection is required.
```

Perhaps there should be some sort of flag that prevents type checking? Or perhaps in the spirit of duck typing, it should raise a warning rather than an exception?

[1]: https://github.com/jamesls/fakeredis
[2]: https://github.com/nvie/rq/blob/master/rq/compat/connections.py#L26

foxx:
This has been discussed in #514, #282 and #88.

Using an explicit type check via `isinstance`, rather than duck typing, is typically considered unpythonic and breaks compatibility with mock objects such as FakeRedis. This patches removes the type check, and instead looks for a common method that should be present on the object as a hint on whether it's compatible or not.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 redis>=2.7
-click>=3.0.0
+click<6.0
--- a/rq/compat/connections.py
+++ b/rq/compat/connections.py
@@ -22,9 +22,6 @@ PATCHED_METHODS = ['_setex', '_lrem', '_zadd', '_pipeline', '_ttl']
 
 
 def patch_connection(connection):
-    if not isinstance(connection, StrictRedis):
-        raise ValueError('A StrictRedis or Redis connection is required.')
-
     # Don't patch already patches objects
     if all([hasattr(connection, attr) for attr in PATCHED_METHODS]):
         return connection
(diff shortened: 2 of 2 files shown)


foxx:
Failed tests seem to be unassociated, can you confirm @selwin?

selwin:
This is due to a bug in `click==6.0` which was just released: https://github.com/mitsuhiko/click/issues/464 . Could you please modify `click` in `requirements.txt` to `click<6.0`? This should hopefully pass the test.

foxx:
Beautiful, that seemed to fix it @selwin :)

foxx:
Just been thinking about this some more, perhaps raising a warning is actually a better approach, because otherwise RQ is trying to act as a user library. It's also a bit misleading because `setex` doesn't necessarily mean it will work, it's just an error message hint. Do you think I should remove that and instead make it raise a warning? If the user doesn't want the warning, they can add a filter.

selwin:
Thanks!

PR #628, opened 2015-12-06
make sure state and exc_info is text_type

SkyLothar:
`hget` will return bytes value in python3.
make sure state is text type and json serialisation-able.
https://github.com/nvie/rq-dashboard/pull/90 same problem


nvie:
Thanks!

PR #633, opened 2015-12-11
Save date on which a busy worker receives a warm shutdown request

olingerc:
Would you be interested in something like this? I use it on a custom UI to tell the user that he already clicked on the "delete after job finishes" button.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -257,11 +257,9 @@ class Worker(object):
             p.expire(self.key, 60)
             p.execute()
 
-    def set_warm_shutdown_requested_date(self):
-        """Sets the date on which the worker received a warm shutdown request"""
-        with self.connection._pipeline() as p:
-            p.hset(self.key, 'warm_shutdown_requested_at', utcformat(utcnow()))
-            p.execute()
+    def set_shutdown_requested_date(self):
+        """Sets the date on which the worker received a (warm) shutdown request"""
+        self.connection.hset(self.key, 'shutdown_requested_date', utcformat(utcnow()))
 
     @property
     def birth_date(self):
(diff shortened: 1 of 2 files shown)


olingerc:
If you want to add this feature, I can also implement a new worker state. I'll also add a test case.

selwin:
More state information can't be bad. Please write a testcase for this!

rq/worker.py
```diff
@@ -257,6 +257,12 @@ def register_death(self):
             p.expire(self.key, 60)
             p.execute()
 
+    def set_warm_shutdown_requested_date(self):
```
selwin: Can you please rename this to "set_shutdown_requested_date"?

rq/worker.py
```diff
@@ -265,6 +271,13 @@ def birth_date(self):
             return utcparse(as_text(birth_timestamp))
 
     @property
+    def warm_shutdown_requested_date(self):
```
selwin: This can be changed to "shutdown_requested_date"

rq/worker.py
```diff
@@ -257,6 +257,12 @@ def register_death(self):
             p.expire(self.key, 60)
             p.execute()
 
+    def set_warm_shutdown_requested_date(self):
+        """Sets the date on which the worker received a warm shutdown request"""
+        with self.connection._pipeline() as p:
```
selwin: I don't think pipeline is needed as we're only updating one key. If this method is not used in more than one places, perhaps we could just inline the code in "request_stop", since it's a one liner anyway.

olingerc:
Updates as wanted. Two notes:
I could add a "real" state to the WorkerStatus enum equivalent to BUSY, IDLE, ... but we do not have a DEAD state either and I do not see a reason to. If you want it, I can of course do it.
In order to know if a worker died as a consequence of a warm or cold shutdown, it is enough to see if the state of the dead worker is idle (warm shutdown) or busy (cold shutdown). Maybe some users want to keep track of that. Worth mentioning in the documentation I think.

selwin:
Thanks for the PR!

Issue #632, opened 2015-12-11
job.delete() should also remove it from StartedJobRegistry


olingerc:
Currently jobs stay in the StartedJobRegistry even when job.delete is called. I know that the registries are cleaned up every hour, but maybe this could be added to the delete function in the job class (or cancel?).
I give the user the option to delete jobs even while they are running.

On the other hand, the StartedJobRegistry is currently not used anywhere else, so letting the automatic cleanup do its job might be sufficient? What do you think?

``` python
registry = StartedJobRegistry(job.origin, self.connection)
registry.remove(job, pipeline=pipeline)
```

selwin:
Yeah, deleting the job should also remove it from the registry. PR welcome :)

selwin:
Fixed in https://github.com/nvie/rq/pull/810

PR #635, opened 2015-12-13
Job delete should delete from StartedJobRegistry

olingerc:
First draft. I am no longer 100% sure about this, since deleting the job still leaves it on the worker until it has finished executing, so technically it still has been started. On the other hand, we delete the redis hash from rq:jobs so this should be fine also?


olingerc:
Refers to #632

olingerc:
Hmmm... Hold off on this for now, my tests this morning fail. Don't code at night without coffee ;-)

olingerc:
I updated the PR to resolve conflicts and use connection instead of pipeline. I also added a test case.
@selwin. Does updating the PR automatically generate a message to you oris it better to specifically add a comment?

selwin:
Sorry for the super late reply.

I agree with your assessment that allowing the user to delete a running job is probably not a good idea. Perhaps we should throw an exception with a descriptive error message if the job is still running. If the user still wants to delete this, he'd need to first stop/kill the job before deleting it.

selwin:
Ping @olingerc :)

selwin:
Fixed in https://github.com/nvie/rq/pull/810

PR #639, opened 2015-12-22
Correction to results-ttl value on worker command

nicksnell:
The `--results-ttl` value isn't currently being cast as an integer when passed from the command line, this causes issues further down when a value is sent as it remains a string i.e:

```
..../site-packages/rq/registry.py", line 32, in add
     score = ttl if ttl < 0 else current_timestamp() + ttl
TypeError: unorderable types: str() < int()
```

Fixed to cast value as integer.


selwin:
Thanks for the PR!

PR #640, opened 2015-12-28
When not async set job finished

ecarreras:
When you're working in not `async` mode (usually when debugging) and you're checking if the job is finished, never is finished because the `worker` sets the `FINISHED` status.

With this patch the job is marked as finished when working in async mode.


selwin:
Thanks for the PR!

PR #641, opened 2016-01-09
Fix connection stack in stock worker

### LINKED ISSUE #479: get_current_job() throws NoRedisConnectionException in 0.5.0

kampfschlaefer:
When running the stock worker itself, the `_connection_stack` is empty. Sadly this only shows when the `rqworker` runs in its own python process. The existing tests in test_worker.py and test_cli.py don't do that.

This branch adds tests that run the stock worker in its own subprocess, thus showing the problem. And it extends the cli to push (and pop) the redis connection when running the worker.

Also I changed the existing `access_self` function to also assert on `get_current_connection`. And this extends the existing worker test to check the failed queue to see that no job failed.

This fixes #479.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -16,7 +16,7 @@ from datetime import timedelta
 
 from rq.compat import as_text, string_types, text_type
 
-from .connections import get_current_connection
+from .connections import get_current_connection, push_connection, pop_connection
 from .defaults import DEFAULT_RESULT_TTL, DEFAULT_WORKER_TTL
 from .exceptions import DequeueTimeout
 from .job import Job, JobStatus
@@ -581,6 +581,9 @@ class Worker(object):
         self.prepare_job_execution(job)
 
         with self.connection._pipeline() as pipeline:
+
+            push_connection(self.connection)
+
             started_job_registry = StartedJobRegistry(job.origin, self.connection)
 
             try:
@@ -622,6 +625,9 @@ class Worker(object):
                     pass
                 self.handle_exception(job, *sys.exc_info())
                 return False
+
+            finally:
+                pop_connection()
 
         self.log.info('{0}: {1} ({2})'.format(green(job.origin), blue('Job OK'), job.id))
         if rv is not None:
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -9,7 +9,7 @@ from __future__ import (absolute_import, division, print_function,
 import os
 import time
 
-from rq import Connection, get_current_job
+from rq import Connection, get_current_job, get_current_connection
 from rq.decorators import job
 from rq.compat import PY2
 
@@ -55,6 +55,7 @@ def create_file_after_timeout(path, timeout):
 
 
 def access_self():
+    assert get_current_connection() is not None
     assert get_current_job() is not None
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ python:
   - "pypy"
 install:
   - if [[ $TRAVIS_PYTHON_VERSION == 2.6 ]]; then pip install -r py26-requirements.txt; fi
-  - pip install -r requirements.txt
+  - pip install -e .
   - pip install pytest-cov
   - pip install coveralls
   #- pip install pytest  # installed by Travis by default already
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,6 @@ install:
   - pip install coveralls
   #- pip install pytest  # installed by Travis by default already
 script:
-  - RUN_SLOW_TESTS_TOO=1 py.test --cov rq
+  - RUN_SLOW_TESTS_TOO=1 py.test --cov rq --durations=5
 after_success:
   - coveralls
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -302,6 +302,7 @@ class TestJob(RQTestCase):
         q.enqueue(fixtures.access_self)  # access_self calls get_current_job() and asserts
         w = Worker([q])
         w.work(burst=True)
+        assert get_failed_queue(self.testconn).count == 0
 
     def test_job_access_within_synchronous_job_function(self):
         queue = Queue(async=False)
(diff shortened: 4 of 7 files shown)


kampfschlaefer:
okay, I have to figure out how to start the rqworker without it being installed. and without knowing how the python binary is called. in tox it worked...

tests/fixtures.py
```diff
@@ -55,9 +55,15 @@ def create_file_after_timeout(path, timeout):
 
 
 def access_self():
+    assert get_current_connection() is not None
     assert get_current_job() is not None
 
 
+def schedule_access_self():
```
selwin: Since this function is only used in one place, could you please move this function definition to within `test_run_scheduled_access_self`?

selwin:
@kampfschlaefer I just read the comments of #479 again. The main issue is that the worker runs inside its own process which is unable to call `get_current_connection`, correct?

The current patch will only fix the issue for people running the stock `rqworker` command - and I know some people have their own worker scripts. I think it would be better to call `push_connection()` in `worker.prepare_job_execution()`. This should fix this bug for everyone. I haven't tried this in code though, do you think this would work?

kampfschlaefer:
> The current patch will only fix the issue for people running the stock rqworker command - and I know some people have their own worker scripts. I think it would be better to call push_connection() in worker.prepare_job_execution(). This should fix this bug for everyone. I haven't tried this in code though, do you think this would work?

I will try this the next days and update this PR.

Would adding this `push_connection` break anything for people that already do that in their own workers? Or would it just be one more item on the stack?

kampfschlaefer:
Hi,

I did the wanted changes. Also I pulled in master and did some fixes for flake8 to stop complaining (at least the flake8 as run by tox, my editor still has lots of red).

This PR should be ready for merging. Or you just merge the stuff before my reverse-merge. Or if wanted I can rebase it all against current master…

kampfschlaefer:
funny how the slow execution of the pypy tests marks the whole PR bad :-(

selwin:
@kampfschlaefer I re-ran the test on Travis and everything now passes. Could you please move `TestWorkerSubprocess` from `test_cli` to `test_worker` though? Thanks!

kampfschlaefer:
Ping. I rebased and squashed this PR into three commits. (By accident, got asked to rebase a completely different PR in another project but didn't read the mail correctly. :wink: )

Maybe the the cleaner history now is more suited for merging? (Apart from the pypy tests failing again…)

selwin:
Sorry, I totally forgot about this PR @kampfschlaefer . Could you please fix the failing pypy test and resolve the conflicts. Again, sorry for forgetting about this, I've been very busy with work lately.

kampfschlaefer:
I would love to fix the failing test. Unfortunately I can not reproduce the failure locally. I will rebase this PR against current master, maybe travis will not have a problem then.

kampfschlaefer:
Finally I got travis to be green on this PR. I have the feeling that the timeouts are highly volatile…

ping @selwin

selwin:
@kampfschlaefer I'm really sorry that I missed this PR. Due to recently merged changes, this PR no longer merges cleanly. Could you please resolve the merge conflicts so I can pull this in?

rq/worker.py
```diff
@@ -635,6 +637,8 @@ def perform_job(self, job, queue):
         else:
             self.log.warning('Result will never expire, clean up result key manually')
 
+        pop_connection()
```
nvie: This should be called in a `finally:` block, to make sure it's always called, even in case of an exception.

kampfschlaefer:
Aaand I am back to slow pypy marking the PR red…

@selwin @nvie please restart the travis run or just merge this PR despite being red.

rq/worker.py
```diff
@@ -560,6 +560,8 @@ def prepare_job_execution(self, job):
         """
         timeout = (job.timeout or 180) + 60
 
+        push_connection(self.connection)
```
selwin: One more minor thing. Should we move this statement to `perform_job()`, right before `prepare_job_execution()` is called? Having `push_connection()` and `pop_connection()` in the same scope makes more sense.

selwin:
I restarted the build and they all pass :)

rq/worker.py
nvie: +1 to that. The push/pop calls should always be balanced.

kampfschlaefer:
Moved the `push_connection` call into `perform_job()`. Now its on the same logical level as the `finally` that contains the `pop_connection`.

selwin:
Merged, thanks!

dan-blanchard:
Could we please get a release with this bugfix? We're still pinned to 0.4.2 because of this issue.

nvie:
@dan-blanchard I just released rq==0.7.0 to PyPI, which includes this bugfix.  Let me know if that works for you!

dan-blanchard:
thanks @nvie!

PR #642, opened 2016-01-10
Worker only creates log handlers if there aren't any.

butla:
I am running worker on a separate process starting from my app using `multiprocessing`. My application configures the logs, but the worker class also wants to do it (actually, it just adds one handler). The effect is that logs from worker are doubled and don't go to the stream I want.

I hope that I understood the original intention behind this code.

### DIFF
--- a/rq/logutils.py
+++ b/rq/logutils.py
@@ -16,10 +16,25 @@ from rq.utils import ColorizingStreamHandler
 
 def setup_loghandlers(level='INFO'):
     logger = logging.getLogger('rq.worker')
-    if not logger.handlers:
+    if not _has_effective_handler(logger):
         logger.setLevel(level)
         formatter = logging.Formatter(fmt='%(asctime)s %(message)s',
                                       datefmt='%H:%M:%S')
         handler = ColorizingStreamHandler()
         handler.setFormatter(formatter)
         logger.addHandler(handler)
+
+
+def _has_effective_handler(logger):
+    """
+    Checks if a logger has a handler that will catch its messages in its logger hierarchy.
+    :param `logging.Logger` logger: The logger to be checked.
+    :return: True if a handler is found for the logger, False otherwise.
+    :rtype: bool
+    """
+    while True:
+        if logger.handlers:
+            return True
+        if not logger.parent:
+            return False
+        logger = logger.parent


selwin:
I'm not a logging expert, but what we want to do is to setup logging if no current handler is set. Does the proposed patch still honor the original intention of the implementation and also fixes your problem?

butla:
Yes, it does. My implementation is just slightly more thoughtful than the original one - it not only checks for handler in the one specific logger, but also in its parent loggers (e.g. root logger)

redbaron4:
Couldn't you just configure the "rq.worker" qualname logger in your application to send logs to wherever you want?

```
[loggers]
keys = root, your_app, rq

[handlers]
keys = console

[formatters]
keys = generic

[logger_root]
level = INFO
handlers = 

[logger_your_app]
level = DEBUG
handlers = console
qualname = your_app

[logger_rq]
level = DEBUG
handlers = console
qualname = rq.worker
propagate = 0

[handler_console]
class = StreamHandler
args = (sys.stderr,)
level = NOTSET
formatter = generic

[formatter_generic]
format = %(asctime)s %(levelname)-5.5s [%(name)s][%(threadName)s] %(message)s
```

Or some similar config which also sets a handler for logger "rq.worker". That way when rqworker starts, it should find a handler already attached to the logger and not add the second handler.

butla:
Well, that's what I did in [my application] (https://github.com/butla/PyDAS/blob/master/data_acquisition/app.py#L48). I needed to specifically create a null handler for RQ, because if I didn't, the log messages were getting duplicated (they were caught by rq handler and root handler with my configuration).

I simply think that this is inelegant. While using the library I shouldn't need to worry about it's logging internals. That's the idea behind Python's logging framework.

redbaron4:
I understand your motivation for this pull request and agree that logging needs to be streamlined. However, adding a new check that checks logging hierarchy all the way to root seems to be complicated.
IMO an alternative simpler way maybe to remove the logging setup from the core worker code so that any user application which uses ``Worker`` may configure ``rq.worker`` logger as it pleases.

[This line](https://github.com/nvie/rq/blob/master/rq/worker.py#L418) adds the handler in the library code by calling `setup_loghandlers()`.
As you have pointed out that when using library you shouldn't be worried about logging internals, removing the above maybe an alternate option.

The CLI worker code already sets up its own logging (by calling the `setup_loghandlers()` separately [here](https://github.com/nvie/rq/blob/master/rq/cli/cli.py#L171)), so ``rqworker`` will not be affected.
As your code does not configure rq.worker beyond adding the NullHandler and only configures root logger, it should be unaffected (as long as you do not set ``propagate`` to False in ``rq.worker`` logger).
Any third-party programs using the rq.worker library can get rq worker logs by appropriately configuring the `rq.worker`` logger in addition to its own loggers

proofit404:
remove the logging setup from the core worker code

Definitely agree. Also it's possible to make logger global variable for easy configuration.

butla:
I've also thought about removing the line with setup_loghandlers() but I thought that someone may actually rely on it. But if you say that that's OK (you're the maintaineers, right? :) ) you can just remove this line and close my PR.

proofit404:
you're the maintaineers, right?

No.

redbaron4:
> Also it's possible to make logger global variable for easy configuration.
Yes that would be a great idea. Ideally the RQ library should use the ``rq`` top-level logger and individual modules (workers, jobs) etc. use next-level logger names like ``rq.worker``, ``rq.job`` etc.

> you're the maintaineers, right?
No.

butla:
Then I guess I'l create an alternative PR with the handler setup line removed. About the global loggers - is it really necessary to expose them as attributes? Wouldn't it be enough to just document how they are called so the code user can get the logger with standard `logging.getLogger(NAME)`?

redbaron4:
We don't need to expose the loggers themselves....just their names. And document them of course.

If a library user knows that this library uses `XYZ` as top level logger name, it is easier for user to configure how all logs of that library are handled

butla:
FYI, I'll work on a new PR later this week.

butla:
I've created issue #669 and linked my second PR to it. I've actually just cut the one line out. I don't think that any more documentation of logger names is necessary since they use standard module names and can be visible in people's logs.

selwin:
Thanks for the PR @butla.

PR #647, opened 2016-02-03
docs: Use rq command instead of rqworker/rqinfo

orangain:
It is confusing to newbie like me whether to use `rqworker` or `rq worker`. According to [the source code](https://github.com/nvie/rq/blob/0.5.6/setup.py#L48-L49), use of `rq worker` seems to be recommended rather than `rqworker`. So this pull request replaces all the `rqworker` and `rqinfo` with `rq worker` and `rq info` in the documentation.


selwin:
Thanks for the PR!

PR #651, opened 2016-02-04
Update outdated sample codes in README.md

orangain:
* Use explicit connection management instead of `use_connection()`.
* Use `rq worker` instead of `rqworker`.


selwin:
Thanks!

PR #656, opened 2016-02-16
Test fixes

antoineleclair:
Somes minor fixes and adding a few tests.

### DIFF
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -347,12 +347,6 @@ class TestWorker(RQTestCase):
         worker = Worker([q], job_class=CustomJob)
         self.assertEqual(worker.job_class, CustomJob)
 
-    def test_custom_job_class_by_string(self):
-        """Ensure Worker accepts custom job class using dotted notation."""
-        q = Queue()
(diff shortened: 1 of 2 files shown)


tests/test_worker.py
```diff
@@ -347,6 +347,16 @@ def test_custom_job_class(self):
         worker = Worker([q], job_class=CustomJob)
         self.assertEqual(worker.job_class, CustomJob)
 
+    def test_custom_job_class_is_not_global(self):
```
selwin: Thanks for this PR! I'm not sure whether this is really needed though.

selwin: Perhaps this should be made clear on the documentation, but I'm not sure we need to test this.

antoineleclair: This is some kind of regression test that would prevent someone from accidentally assigning it to the class instead of the instance.

I can remove this test if you want. Also, I included a similar test in my other pull request to add a custom queue class in the worker. Let me know if you think I should remove those tests.

selwin: Never mind, this makes sense if you take https://github.com/nvie/rq/pull/657/files into account. We'll keep this here :).

PR #659, opened 2016-02-16
Add ttl argument to decorator

javimb:
Add a ttl argument to decorator which will set a TTL for the job.
Default value is None.


selwin:
Thanks!

Issue #658, opened 2016-02-16
Use Job.set_status() instead of setting protected method


VicarEscaped:
In [worker.py](https://github.com/nvie/rq/blob/master/rq/worker.py#L595), instead of set job status with set_status(), status assign via protected attribute. Should we use `job.set_status(JobStatus.FINISHED, pipeline)` ?

selwin:
Yes, definitely. Mind opening a PR for this?

PR #661, opened 2016-02-16
Update job status with set_status

VicarEscaped:
Update job status with set_status instead of setting attribute
PR for #658


selwin:
Thanks for the Pr!

Issue #665, opened 2016-02-24
Logging is not configured properly on Python-2.6


redbaron4:
On Python2.6 no logging messages are obtained on startup
```

$ ~/envs/tscript/bin/rqworker 
^C $
```

Patch for [this issue](http://bugs.python.org/issue18046) probably changed the way ``setLevel`` works Python-2.7 onwards.

redbaron4:
The culprit seems to be [this line in the code](https://github.com/nvie/rq/blob/master/rq/logutils.py#L20) which calls ``logger.setLevel()``

In Python-2.6
```
Python 2.6.6 (r266:84292, Jan 22 2014, 09:42:36) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import logging
>>> log=logging.getLogger("test")
>>> log.setLevel("INFO")
>>> log.getEffectiveLevel()
'INFO'
```


In Python-2.7
```
Python 2.7.6 (default, Jul  1 2015, 17:40:54) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import logging
>>> log=logging.getLogger("test")
>>> log.setLevel("INFO")
>>> log.getEffectiveLevel()
20
```

Since log-levels are actually ints, they are not correctly set for python-2.6 by rqworker

One way is to use [getLevelName](https://docs.python.org/3.4/library/logging.html#logging.getLevelName), however the accompanying note states that this behaiviour is an anomaly (which is nevertheless retained at this moment).
Other way is to have a dict mapping right in the code which maps string names to log-levels

```
{"ERROR": logging.ERROR,
 "DEBUG": logging.DEBUG,
 "INFO": logging.INFO}
```
and then use this mapping to get the actual level value.

PR #666, opened 2016-02-25
Fix logging regression for Python-2.6

redbaron4:
logger.setLevel() doesn't work when a string is passed in Python-2.6.
This patch checks if the level has been set and if not (in Python-2.6),
tries to set it again using appropriate level constants
Also unused dictConfig import is removed


redbaron4:
Closes #665

selwin:
Thanks!

PR #668, opened 2016-02-29
Log handlers created only when running with CLI.

butla:
When using rq.worker as I library, clients won't need to worry about internal logging configuration. Oh, and pyflakes doesn't file because of my change :)


selwin:
I still think having some kind of logging is useful as a default. If people don't want this, they can be override the logging handler.

redbaron4:
@selwin IMO the logging decision should be left to the user. The library should not instantiate handlers for the log as it leads to unintended problems (as posted by OP).
About having some kind of default, I think that is best handled at script level. The "rq" or "rqworker" command line script should instantiate the handlers (as is being done already), so that messages can be flashed to user.
The library handler instantiation can be removed without any changes in existing logging behaviour.

PR #674, opened 2016-03-13
Ignore lambdas in flake8

kampfschlaefer:
Sadly the flake8 isn't run in travis (and I don't change that here). But it should still be made green. And the config should be placed in setup.cfg so editors can use the same config for flake8 as tox does.


selwin:
Thanks!

PR #676, opened 2016-03-14
cannot depend on jobs with result_ttl == 0

horida:
see https://github.com/nvie/rq/issues/675


selwin:
As I mentioned in the issue thread, I don't think this is an issue. If you can produce a failing test casw, I'll reopen this again.

PR #685, opened 2016-04-01
Ability to set a custom namespace for the redis keys.

rberrelleza:
This is based on the work we tried before on https://github.com/nvie/rq/pull/459 (It was easier to create a new PR than rebase and all that). I took all the comments you had, plus Jay's original idea, and re-implemented it.

Some testing is still needed, but wanted to get your thoughts on this approach before going further down this path.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -12,12 +12,12 @@ from .defaults import DEFAULT_RESULT_TTL
 from .exceptions import (DequeueTimeout, InvalidJobOperationError,
                          NoSuchJobError, UnpickleError)
 from .job import Job, JobStatus
-from .utils import import_attribute, utcnow
+from .utils import import_attribute, utcnow, build_key
 
 
-def get_failed_queue(connection=None):
+def get_failed_queue(connection=None, namespace=None):
     """Returns a handle to the special failed queue."""
-    return FailedQueue(connection=connection)
+    return FailedQueue(connection=connection, namespace=namespace)
 
 
 def compact(lst):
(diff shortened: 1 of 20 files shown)


tests/test_utils.py
```diff
@@ -0,0 +1,11 @@
+from rq.utils import build_key
+from tests import RQTestCase
+
+
+class TestHelpers(RQTestCase):
+
+    def test_build_key(self):
+        """Ensure the redis key is correctly build when there's a namespace present"""
```
tomviner: I think you mean *is correctly built*

rq/queue.py
```diff
@@ -32,33 +32,35 @@ class Queue(object):
...
         the internal Redis keys.  Can be used to reverse-lookup Queues by their
         Redis keys.
         """
-        prefix = cls.redis_queue_namespace_prefix
+        prefix = build_key(cls.redis_queue_namespace_prefix, namespace)
+
         if not queue_key.startswith(prefix):
             raise ValueError('Not a valid RQ queue key: {0}'.format(queue_key))
         name = queue_key[len(prefix):]
         return cls(name, connection=connection)
```
olruzhytskyi: There should be a namespace=namespace positional argument to create an instance with namespace.

torfsen:
This feature would come in very handy for us -- is there anything I can do to help it along?

proofit404:
Why anyone want to customize such internal stuff of rq?

TkTech:
👍 this would be of great help to the ckan/ckan project.

proofit404:
I'm sorry I still don't get it. Can you elaborate more on the term "multiple simultaneous job queues". What problems does this pattern solve? In which way it do that?

rberrelleza:
@proofit404 in our case, we have different services that use RQ in the same redis instance.  Currently, rq assumes that the rq:{id} or rq:{queue_name} are exclusive, which doesn't work well in the described scenario, mainly due to name collisions.

proofit404:
@rberrelleza Why you use one database for all you services? It's simple to use database 0 for service A and database 1 for service B. Is this appropriate solution to you?

torfsen:
@proofit404 Many small/free plans for hosted Redis only include a single database.

proofit404:
@torfsen Ok. Different databases does not work in this case.

@rberrelleza As rq documentation says you must provide unique job ids (if you use don't use uuid module) and unique queue names for all your running services. The only possible name collision not under your control is failing queues. You can override Worker.move_to_failed_queue and FailedQueue.__init__ if you want to separate failures from different services.

At current state rq intended to be "one cluster per one redis instance". If you want to change this paradigm into "multiple clusters per one redis instance" then add explicit API part for this and leave calculated prefixes as internal part.

PR #687, opened 2016-04-04
Allow Worker work method to specify the log verbosity

jlopex:
This is just a patch to expose the logging level on the Worker.work() method.

For instance, it will allow to mute all RQ output if you're using the RQ Worker class.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -388,7 +388,7 @@ class Worker(object):
         if before_state:
             self.set_state(before_state)
 
-    def work(self, burst=False):
+    def work(self, burst=False, logging_level=logging.INFO):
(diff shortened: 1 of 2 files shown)


rq/worker.py
```diff
@@ -388,7 +388,7 @@ def check_for_suspension(self, burst):
         if before_state:
             self.set_state(before_state)
 
-    def work(self, burst=False):
+    def work(self, burst=False, level=logging.INFO):
```
selwin: Could you please rename the argument to logging_level? Thanks!

jlopex:
@selwin you got it :-)

selwin:
Thanks!

PR #600, opened 2015-11-06
Cancel and Delete differences

glaslos:
This addresses the differences between job.cancel() and job.delete() see discussion in #362 and especially https://github.com/nvie/rq/issues/362#issuecomment-49991135

PR #691, opened 2016-04-15
Don't call job.cancel if job has finished

amyangfei:
This issue #600 migrate `job.cancel()` to `job.delete()` .

When enqueue a job with `result_ttl=0`, after this job is executed, `job_status` is not updated and the `job.cleanup(result_ttl, pipeline=pipeline)` will be called immediately. In `job.cleanup`, `job.cancel()` will be called to remove itself from its `origin queue`, however the job object has already been poped from the `origin queue`.

A redis `lrem ` is used in  `job.cancel()`, this command is O(N) in time complexity and will be very slow if have millions of jobs in the queue. What's more calling `job.cancel()` with a finished job is meaningless.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -480,9 +480,10 @@ class Job(object):
             queue.remove(self, pipeline=pipeline)
         pipeline.execute()
 
-    def delete(self, pipeline=None):
+    def delete(self, pipeline=None, remove_from_queue=True):
         """Cancels the job and deletes the job hash from Redis."""
-        self.cancel()
+        if remove_from_queue:
+            self.cancel()
         connection = pipeline if pipeline is not None else self.connection
         connection.delete(self.key)
         connection.delete(self.dependents_key)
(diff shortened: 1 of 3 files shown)


rq/worker.py
```diff
@@ -604,6 +604,8 @@ def perform_job(self, job):
 
                     finished_job_registry = FinishedJobRegistry(job.origin, self.connection)
                     finished_job_registry.add(job, result_ttl, pipeline)
+                else:
+                    job.patch_status(JobStatus.FINISHED)
```
selwin: This seems a bit hackish to me. Let me try to think of a better way.

rq/job.py
```diff
@@ -482,7 +488,9 @@ def cancel(self):
 
     def delete(self, pipeline=None):
```
selwin: I think it would be better to introduce an argument "remove_from_queue" which defaults to True. In "job.cleanup", we would then call  `job.delete(remove_from_queue=False)`

amyangfei: missed origin commit https://github.com/amyangfei/rq/commit/2921b8c9d0270fe349b620ab3ca02e2bda125f8b

Add an argument "remove_from_queue" looks better, I have made another commit.

selwin:
Thanks for the PR!

Issue #568, opened 2015-07-30
Propagate custom Job class down to queue in worker


aroberts:
When invoking the worker with a custom Job class, I noticed that the queue was returning an `rq.job.Job` instance (instead of my custom job class) during `Queue.dequeue_any()`. I think I have a workaround, but it's confusing that the worker returns a custom job type for `get_current_job()`, but the jobs the worker receives to execute from the queue aren't of the custom type.

selwin:
Yeah, that's confusing. I think `dequeue_any()` should return a custom job class. I think we should store the job class information in the job so `Queue.dequeue_any()` can use the right job instance.

Do you mind making a PR for this?

selwin:
Fixed in https://github.com/zamzterz/rq/commit/78bd55695c85ce284de8f8c02a911ad6b3d6f015

PR #694, opened 2016-04-21
Store the job class information in the queue, closes issue #568.

rebeckag:
Pickle the class object of the Job instance and include it when
serializing the Job instance for storage in the Redis queue.


selwin:
Hello, I don't think pickling the job class and storing it in Redis is a good idea. For one, this would definitely break if people use other serialization option like JSON (this has been discussed forever, just that no one has had the time to implement it) which can't serialize complex Python objects.

You should be able to achieve what you want to do by using custom `Queue` classes though.

Issue #700, opened 2016-05-03
cancel_job method doesn't work properly


jlopex:
I realised that somehow the cancel_job helper function doesn't work properly anymore.

This is the code I used to detect the problem

```
from test2 import to_run
from rq import Queue, cancel_job
from rq.job import Job
from redis import Redis

def main():
    redis_conn = Redis()
    fix = Queue('fix', connection=redis_conn)
    fix.enqueue(to_run)
    job_ids = fix.get_job_ids()

    print job_ids
    for id in job_ids:
        cancel_job(id, connection=redis_conn)

    jobs = fix.get_jobs()
    print jobs
    for job in jobs:
        job.cancel()

    jobs = fix.get_jobs()
    print jobs

if __name__ == '__main__':
    main()
```

The output of this code is:

```
[u'ee114e13-e860-481e-809e-5e14edfd9551']
[Job(u'ee114e13-e860-481e-809e-5e14edfd9551', enqueued_at=datetime.datetime(2016, 5, 3, 13, 0, 48))]
[]
```

With this I was able to trace down the issue, and it's related to how do we obtain the Job object on the cancel_job function, I'll prepare a test and a fix for this issue.

selwin:
Fixed in #701

PR #701, opened 2016-05-03
Fixes issue #700

jlopex:
This patch includes a fix and a simple test for the cancel_job helper function that was not working as expected.

fixes issue #700


jlopex:
@selwin I noticed something weird when travis was passing the tests on pypy build.

I forced a new build by updating the commit_id and they passed correctly in all platforms

selwin:
Thanks for the PR!

PR #703, opened 2016-05-06
Refresh Job to keep meta after execute

realPy:
Meta when job is finished is override by the initial value
The job must be refreshed to keep meta given in the job process.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -32,7 +32,7 @@ from .version import VERSION
 try:
     from setproctitle import setproctitle
 except ImportError:
-    def setprocname(*args, **kwargs):  # noqa
+    def setproctitle(*args, **kwargs):  # noqa
         pass
 
 green = make_colorizer('darkgreen')
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -30,7 +30,7 @@ from .utils import (ensure_list, enum, import_attribute, make_colorizer,
 from .version import VERSION
 
 try:
-    from procname import setprocname
+    from setproctitle import setproctitle
 except ImportError:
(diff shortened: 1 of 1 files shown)


proofit404:
Do you mean job meta will be overwritten by result save if some one already change meta during job execution?

realPy:
Not exactly,
As you can see in the modify file job is performed and reused just after but its possible that is job has been altered..
If in this performed method the meta is changed or saved,  when the method is finished the job instance used to perform is old one before the operation. So the dict meta associated is the old meta and not  the last one.
It's easy pretty to reproduce this bug. Enqueue a func with a meta. Get the job in the func and alter meta.
Get the meta when it's finished and it's the initial meta :)

proofit404:
Job function may alter meta too. How to process this conflict?

realPy:
There's no conflict, the queue is sequential. So if job has altered meta just after the job executed reload data from redis with job.refresh() and it's ok.

realPy:
Just fix a bug with setprocname.
For strange reason i have with queue name "myqueue" setprocname set "rq: Listening on m   myqueue"
setproctitle is the same but without this bug.

selwin:
@realPy thanks for the PR.

RQ does not expect `job.meta` to be modified. Rather than relying on `job.meta()`, your function should keep track of its state somewhere else (and not rely on the job queue). `job.meta()` is designed to used by other libraries in the RQ ecosystem such as rq-scheduler.

If you really want to keep doing this though, I suggest using a custom `Worker` class that calls `job.refresh()` so it doesn't introduce an additional call to Redis because most people don't need it.

PR #707, opened 2016-05-16
Fix Job.cancel() method on failed queue

jlopex:
The cancel method was not working correctly for the special failed queue.


selwin:
Thanks for the PR!

PR #710, opened 2016-05-25
Fix issue 702

spiliopoulos:
In order to solve issue 702 we have to check whether a work-horse terminated unexpectedly (by inspecting the exit code of the work-horse process). If it exited unexpectedly we check if the job has either been marked as finished, failed or other valid states. If it's not in any valid state we mark it as failed and move it to the failed queue. Since the process was terminated unexpectedly (think OOM) we do not have any exception context and we can't run any custom exception handlers.
    
There is still a chance that the job will finish successfully but the work-horse process will be killed before the job is marked as finished and we will erroneously mark it as failed. The users should take care to write idempotent jobs.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -530,10 +530,10 @@ class Worker(object):
                         break
                     if job_status not in [JobStatus.FINISHED, JobStatus.FAILED]:
                         with self.connection._pipeline() as pipeline:
-                            job.set_status(JobStatus.FAILED, pipeline=pipeline)
-                            started_job_registry = StartedJobRegistry(job.origin, self.connection)
-                            started_job_registry.remove(job, pipeline=pipeline)
-                            self.set_current_job_id(None, pipeline=pipeline)
+                            self.handle_current_job_failure(
+                                job=job,
+                                pipeline=pipeline
+                            )
                             try:
                                 pipeline.execute()
                             except Exception:
@@ -611,6 +611,27 @@ class Worker(object):
         msg = 'Processing {0} from {1} since {2}'
         self.procline(msg.format(job.func_name, job.origin, time.time()))
 
+    def handle_current_job_failure(
+        self,
+        job,
+        started_job_registry=None,
+        pipeline=None
+    ):
+        """Handles the failure or an executing job by:
+            1. Setting the job status to failed
+            2. Removing the job from the started_job_registry
+            3. Setting the workers current job to None
+        """
+
+        if started_job_registry is None:
+            started_job_registry = StartedJobRegistry(
+                job.origin,
+                self.connection
+            )
+        job.set_status(JobStatus.FAILED, pipeline=pipeline)
+        started_job_registry.remove(job, pipeline=pipeline)
+        self.set_current_job_id(None, pipeline=pipeline)
+
     def perform_job(self, job, queue):
         """Performs the actual work of a job.  Will/should only be called
         inside the work horse's process.
@@ -651,9 +672,11 @@ class Worker(object):
                 pipeline.execute()
 
             except Exception:
-                job.set_status(JobStatus.FAILED, pipeline=pipeline)
-                started_job_registry.remove(job, pipeline=pipeline)
-                self.set_current_job_id(None, pipeline=pipeline)
+                self.handle_current_job_failure(
+                    job=job,
+                    started_job_registry=started_job_registry,
+                    pipeline=pipeline
+                )
                 try:
                     pipeline.execute()
                 except Exception:
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -523,6 +523,25 @@ class Worker(object):
         while True:
             try:
                 _, ret_val = os.waitpid(self._horse_pid, 0)
+                if not (ret_val == os.EX_OK):
+                    job_status = job.get_status()
+                    if job_status is None:
+                        # Job completed and its ttl has expired
+                        break
+                    if job_status not in [JobStatus.FINISHED, JobStatus.FAILED]:
+                        with self.connection._pipeline() as pipeline:
+                            job.set_status(JobStatus.FAILED, pipeline=pipeline)
+                            started_job_registry = StartedJobRegistry(job.origin, self.connection)
+                            started_job_registry.remove(job, pipeline=pipeline)
+                            self.set_current_job_id(None, pipeline=pipeline)
+                            try:
+                                pipeline.execute()
+                            except Exception:
+                                pass
+                            self.move_to_failed_queue_unhandled(
+                                job,
+                                "Work-horse proccess was terminated unexpectedly"
+                            )
                 break
             except OSError as e:
                 # In case we encountered an OSError due to EINTR (which is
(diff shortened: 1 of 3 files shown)


rq/worker.py
```diff
@@ -516,20 +512,55 @@ def execute_job(self, job, queue):
...
+        while True:
+            try:
+                _, ret_val = os.waitpid(self._horse_pid, 0)
+                if not (ret_val == os.EX_OK):
+                    job_status = job.get_status()
+                    if job_status is None:
+                        # Job completed and its ttl has expired
+                        break
+                    if job_status not in [JobStatus.FINISHED, JobStatus.FAILED]:
+                        with self.connection._pipeline() as pipeline:
```
selwin: Since we're using this exact same routine (setting failed status, removing from `StartedJobRegistry`) in other parts of the code to handle failed job, we should move this to a method that can be called in different places.

rq/worker.py
```diff
@@ -516,20 +512,55 @@ def execute_job(self, job, queue):
...
-                    # caused by a SIGINT or SIGTERM signal during
-                    # os.waitpid()), we simply ignore it and enter the next
-                    # iteration of the loop, waiting for the child to end.  In
-                    # any other case, this is some other unexpected OS error,
-                    # which we don't want to catch, so we re-raise those ones.
-                    if e.errno != errno.EINTR:
-                        raise
+
+    def monitor_work_horse(self, job):
+        """The worker will wait for the work horse and make sure it executes
```
selwin: I think the docstring for this method is no longer accurate since this method doesn't wait for the job to finish.

rq/worker.py
```diff
@@ -516,20 +512,55 @@ def execute_job(self, job, queue):
...
+        within the given timeout bounds, or will end the work horse with
+        SIGALRM.
+        """
+        while True:
+            try:
+                _, ret_val = os.waitpid(self._horse_pid, 0)
```
selwin: I'm not too familiar with `waitpid`'s behavior, will this function detect processes that terminate unexpectedly (segfault or OOM)?

tests/test_worker.py
```diff
@@ -621,3 +625,33 @@ def test_working_worker_cold_shutdown(self):
...
+        sentinel_file = '/tmp/.rq_sentinel_work_horse_death'
+        if os.path.exists(sentinel_file):
+            os.remove(sentinel_file)
+        fooq.enqueue(create_file_after_timeout, sentinel_file, 100)
+        job, queue = w.dequeue_job_and_maintain_ttl(5)
+        w.fork_work_horse(job, queue)
+        p = Process(target=kill_work_horse, args=(w._horse_pid,))
+        p.start()
+        p.join(1)
+        w.monitor_work_horse(job)
```
selwin: I know this is very difficult to test, but this doesn't test the case where the process terminates unexpectedly after `w.monitor_work_horse(job)` is called.

proofit404: Run os.kill 9 in a thread?

rq/worker.py
spiliopoulos: waitpid waits for a change in the state of the process (the child terminated; the child was stopped by a signal; or the child was resumed by a signal;). OOM sends a SIGKILL to the process which terminates it and a segfault is just a SIGSEGV to a process (the default handler terminates the process). Both cases should be handled by waitpid (with the exception where a custom signal handler is used for SIGSEGV, in which case I don't think we should intervene since the process handled whatever happened itself).

rq/worker.py
spiliopoulos: Will do. Since I am not very familiar with the codebase, let me know if you think I missed any occurences of this routine

tests/test_worker.py
spiliopoulos: Yes that should work. It's also a better fit to what we are trying to test for.

rq/worker.py
spiliopoulos: I'll try to write something that better fits this use case.

spiliopoulos:
Should I rebase to the updated master?

selwin:
Thanks for the update. Yes, please update this PR so that it merges cleanly with the master branch.

spiliopoulos:
Done. Let me know if you need anything else :)

tests/test_worker.py
```diff
@@ -649,6 +654,34 @@ def test_working_worker_cold_shutdown(self):
...
+            os.remove(sentinel_file)
+        fooq.enqueue(create_file_after_timeout, sentinel_file, 100)
+        job, queue = w.dequeue_job_and_maintain_ttl(5)
+        w.fork_work_horse(job, queue)
+        p = Process(target=wait_and_kill_work_horse, args=(w._horse_pid, 0.5))
+        p.start()
+        w.monitor_work_horse(job)
+        job_status = job.get_status()
+        p.join(1)
+        if os.path.exists(sentinel_file):
```
selwin: Sorry for the long feedback loop. Why do we have an `if` clause in this test? If I'm reading this right, the sentinel file should never exist because the job is killed before the file is created, correct?

rq/worker.py
```diff
@@ -518,20 +514,55 @@ def execute_job(self, job, queue):
...
+
+    def monitor_work_horse(self, job):
+        """The worker will monitor the work horse and make sure that it
+        either executes successfully or the status of the job is set to
+        failed
+        """
+        while True:
+            try:
+                _, ret_val = os.waitpid(self._horse_pid, 0)
+                if not (ret_val == os.EX_OK):
```
selwin: Nitpick, could we change this to `if ret_val != os.EX_OK:`?

tests/test_worker.py
spiliopoulos: My reasoning was that in the ***extreme*** case where the **create_file_after_timeout** job might run to completion before **wait_and_kill_work_horse** gets to kill it (think extremely high system load or something killing the **wait_and_kill_work_horse** process), in which case we don't get to actually test what we are interested in, the test case should succeed (we did not find a failure) than randomly fail. I was probably being overly defensive, since in general I don't expect it to ever hit the if part of that if-else. If you want I can remove it.

rq/worker.py
spiliopoulos: Fixed

rq/worker.py
```diff
@@ -678,6 +732,11 @@ def move_to_failed_queue(self, job, *exc_info):
         self.log.warning('Moving job to {0!r} queue'.format(self.failed_queue.name))
         self.failed_queue.quarantine(job, exc_info=exc_string)
 
+    def move_to_failed_queue_unhandled(self, job, message):
```
selwin: Since this method is only used in one place, let's just place the code inline in `monitor_work_horse()`

rq/worker.py
```diff
@@ -580,6 +611,27 @@ def prepare_job_execution(self, job):
         msg = 'Processing {0} from {1} since {2}'
         self.procline(msg.format(job.func_name, job.origin, time.time()))
 
+    def handle_current_job_failure(
```
selwin: Could you please rename this to `handle_job_failure`?

tests/test_worker.py
selwin: Yeah, please remove the `else` block so we can more easily tell what the expected behavior is by reading the test case.

spiliopoulos:
Done. There are some timing issues with tests that are sleeping for some time and then try to do something to another process. As a result some of them randomly fail on Travis. I could not see any relation to the failing tests with my changes and from what I have seen it has been reported again on previous PRs.

selwin:
Thanks for the PR!

PR #715, opened 2016-06-04
adding heroku worker

samuelcolvin:
see #584.

Not yet ready.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -714,3 +714,81 @@ class SimpleWorker(Worker):
     def execute_job(self, *args, **kwargs):
         """Execute job in same thread/process, do not fork()"""
         return self.perform_job(*args, **kwargs)
+
+
+class ShutDownImminentException(Exception):
+    def __init__(self, msg, extra_info):
+        self.extra_info = extra_info
+        super(ShutDownImminentException, self).__init__(msg)
+
+
+class HerokuWorker(Worker):
+    """
+    Modified version of rq worker which:
+    * stops work horses getting killed with SIGTERM
+    * sends SIGRTMIN to work horses on SIGTERM to the main process so they can crash as they wish
+    Note: coverage doesn't work inside the forked thread so code expected to be processed there has pragma: no cover
+    """
+    imminent_shutdown_delay = 8
+
+    def main_work_horse(self, job, queue):
+        """Modified entry point which ignores SIGINT and SIGTERM and only handles SIGRTMIN"""
+        random.seed()
+
+        signal.signal(signal.SIGRTMIN, self.handle_shutdown_imminent)
+        signal.signal(signal.SIGINT, signal.SIG_IGN)
+        signal.signal(signal.SIGTERM, signal.SIG_IGN)
+
+        self._is_horse = True
+        self.log = logger
+
+        success = self.perform_job(job, queue)
+        os._exit(int(not success))
+
+    def request_stop(self, signum, frame):
+        """Stops the current worker loop but waits for child processes to
+        end gracefully (warm shutdown).
+        """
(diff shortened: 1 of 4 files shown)


samuelcolvin:
before I go any further. @nvie @selwin how do we feel about dropping support for 2.6?

Obviously I can fix the currently failing issue but more broadly 2.6 is pretty ancient and haunted these days.

proofit404:
I know that python 2.6 is old and hard to maintain in projects supports both python versions. But it used in production (in my case as well).

selwin:
Python 2.6 has reached end of life in 2013 so I think we can justify not supporting it in the next release.

Thanks for working on this!

samuelcolvin:
For me this is ready.

rq/worker.py
```diff
@@ -708,3 +714,51 @@ def main_work_horse(self, *args, **kwargs):
     def execute_job(self, *args, **kwargs):
         """Execute job in same thread/process, do not fork()"""
         return self.perform_job(*args, **kwargs)
+
+
+class ShutDownImminentException(Exception):
```
selwin: Can we move this to `exceptions.py` to keep things tidy?

samuelcolvin: Absolutely, I'm away from my computer for the weekend, will do next week.

samuelcolvin: done.

selwin:
@samuelcolvin thanks for the fix, but could you please resolve the conflicts so I can merge this in? It must have conflicted with another [PR](https://github.com/nvie/rq/pull/641) I merged in earlier.

It would be awesome if you could also add a little documentation about this in the `workers` section in the `gh-pages` branch (in a separate PR).

samuelcolvin:
ok, rebased as requested. Should be ready to merge. I'll do a docs pull request when I have a chance.

selwin:
@samuelcolvin thanks a lot!

nvie:
Wow, this is a nice one! Thanks for adding this—this is going to be very useful for people on Heroku!

Sorry for my late (post-merge, even!) response here, but could we perhaps move this into a "contrib" style package? I.e. `from rq.contrib.heroku import HerokuWorker, ShutdownImminentException` ? That keeps it nicely outside of the main package, so non-Heroku users aren't bothered with it.

spiliopoulos:
@selwin 

Hey @samuelcolvin, nice work! I have a quick question though. I have been working on another PR and while running the test suite I noticed some of your tests failing with
    
     AttributeError: 'frame' object has no attribute 'f_exc_traceback'

I took a quick look at your commits and I noticed that you make use of the following attributes of a frame on your signal handler:

     frame_properties = ['f_code', 'f_exc_traceback', 'f_exc_type', 'f_exc_value',                                                                                                                                                                                         
                         'f_lasti', 'f_lineno', 'f_locals', 'f_restricted', 'f_trace']

However, if I am not mistaken 3 of them have been removed in python3:
      f_exc_traceback, f_exc_type, f_exc_value

So I was wondering whether they should be removed and I was also very curious how did this pass from travis.

spiliopoulos:
To answer my second question (why did this pass from Travis), it seems that as long as something raised an exception (even if it was not ShutDownImminentException) before the finished sentinel file was created the test was successful. In my case the exception was:

     AttributeError: 'frame' object has no attribute 'f_exc_traceback'

In any case I feel that the signal handler should be working correctly for all supported python versions. I could create a PR if you agree.

samuelcolvin:
Sorry about the problem, I guess it's probably my responsibility to fix it but if you submit a pr that would be great too, otherwise I'll fix as soon as I can.

spiliopoulos:
I'll see if I can work on it today.

spiliopoulos:
PR #727

Issue #717, opened 2016-06-10
Notify job finished via publish


canni:
Everybody knows that pooling for job result sucks :)

Currently when my jobs finish work I `publish` a message on `channel` to wake up any subscribers interested in job's result, this has one major drawback, all my jobs have to be aware of redis connection, and I must inject it from context. This is a really good example of code smell.

I thought that adding support for `publish`/notifications withing job context might be a good idea, my proposition:

- Add a `notify` parameter a job with default value of `None`
- If `notify` is set to `True`, use `job_id` as a `channel` for notification
- If `notify` is set to `str` use this as a notify channel
- Published message probably can contain just a status (`success`, `exception` etc)

Then a job creator can just use it's own redis connection to get a `PubSub` object and wait for result to come without nasty pooling.

Now before I jump start to hacking, is my way of thinking correct, or am I missing some serious flaw with this idea?

selwin:
I think this is a very interesting idea and something I'd like to have.

Rather than configuring this on a per job basis though, I think I'd like to see this configured on a per queue basis.

Workers would then broadcast `start`, `success`, `fail`,`job_id` along with other relevant information to this channel.

proofit404:
We need to discuss new part of API rather implementation first. I don't think using raw pub/sub redis object is great idea for user experience. We can add high level methods which will use this object under the hood. Maybe something like this

```python
job = queue.enqueue(add, 1, 2) # first class world problem
result = job.wait_result(timeout=360) # listen on channel, use job.fetch when message received
```
It also can raise timeout error if we wait for too long.

selwin:
@proofit404 you're absolutely right. 

The example you gave works on a per job basis and that could already be achieved using `async=False` option. While running job synchronously is useful for tests, it most likely won't be very useful in production.

The API I currently have in mind is:
```python
worker = Worker(queues=[queue], broadcast_events=True) # This will cause the worker to publish job started, finished and other events
```

To observe the events (this is done on a per queue basis):
```python
queue = Queue()
queue.wait_for_events(): # This is simply a thin wrapper around Redis' subscribe command
    ...
```

What do you think?

nvie:
Seems like a good idea to let workers always send these events when jobs complete. Or at the very least, make it the default. (If they push an event out and there are no listeners on the pubsub channel it's effectively a no-op, right?)

[...]

PR #718, opened 2016-06-10
PoC Notifications of status updates

canni:
Proof of Concept for #717 

Notifications are sent for every status update


canni:
This PoC is no longer needed, as we don't consider this kind of per-job behaviour.

PR #722, opened 2016-07-15
Added better signal name lookup

probablytom:
There was a comment asking if there's a better way to do signal name lookups than what was there. There is in Python3.5, so I added it!


selwin:
Thanks for the PR!

PR #726, opened 2016-07-24
Fix failing test import error

spiliopoulos:
The test_run_scheduled_access_self test was failing when the
test suite was run from the package root because the function
schedule_access_self was declared in the module test_workers.py
which was not visible. As a result when it tried to run the job
and it tried to import the module test_worker, it raised an
ImportError. To keep things clean and consistent (the function
access_self, used by schedule_access_self was already declared in
tests/fixtures) I moved schedule_access_self to test/fixtures

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -60,7 +60,6 @@ def access_self():
 
 def schedule_access_self():
     q = Queue('default', connection=get_current_connection())
-    assert get_current_job() is not None
     q.enqueue(access_self)
 
 def echo(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


selwin:
This PR no longer merges cleanly because #710 was pulled in. Could you please resolve the conflicts so I can merge this in?

spiliopoulos:
This test is timing out randomly on pypy on Travis while waiting for the subprocess rqworker to finish running in burst. My guess is that the queue might contain work that might have been there from previous tests leading in a too long burst time (especially when the Travis infrastructure is overloaded). However, I can't reproduce this behaviour locally. If I find time I'll try to figure out what the original author was trying to test and rewrite these tests.

selwin:
This PR increases the worker timeout to 15 and it seems to do the trick: https://github.com/nvie/rq/pull/786/files

PR #727, opened 2016-07-24
Fix request_force_stop_sigrtmin failing for python3

spiliopoulos:
The signal handler request_force_stop_sigrtmin (introduced with the heroku specific worker) would fail in python3 because it was trying to read frame properties that have been removed in python3.
We now handle python2 as a special case with more frame properties.

Also fixed some warnings about using the war function of the logger

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -736,7 +736,7 @@ class HerokuWorker(Worker):
     imminent_shutdown_delay = 6
 
     frame_properties = ['f_code', 'f_lasti', 'f_lineno', 'f_locals', 'f_trace']
-    if sys.version_info[:2] <= (3, 0):
+    if sys.version_info[:2] < (3, 0):
         frame_properties.extend(
             ['f_exc_traceback', 'f_exc_type', 'f_exc_value', 'f_restricted']
         )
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

rq/worker.py
```diff
@@ -734,8 +734,12 @@ class HerokuWorker(Worker):
     causes the horse to crash `imminent_shutdown_delay` seconds later
     """
     imminent_shutdown_delay = 6
-    frame_properties = ['f_code', 'f_exc_traceback', 'f_exc_type', 'f_exc_value',
-                        'f_lasti', 'f_lineno', 'f_locals', 'f_restricted', 'f_trace']
+
+    frame_properties = ['f_code', 'f_lasti', 'f_lineno', 'f_locals', 'f_trace']
+    if sys.version_info[:2] < (3, 0):
```
samuelcolvin: Would have been good to get a chance to review this, surely no need to merge it immediately?

Would it not have been better to use `if PY2` from `compat`? That's surely what `compat` is there for.

samuelcolvin:
Also (and arguably the first step in fixing this) we should have a test that the error raised is indeed `ShutdownImminentException` otherwise this might happen again.

rq/worker.py
spiliopoulos: You are right, I completely forgot about compat. I'll fix that and work also on a test about it too as you suggested.

samuelcolvin: done, see #729

selwin: Sorry about that, was in a hurry. Thanks for the improvements.

PR #729, opened 2016-07-25
Improved fix for python 3 frame handling and tests

samuelcolvin:
Uses compat for PY2 check and add tests which check the error raised by signals is correct.

Fixes #715, improves #727


selwin:
Thanks!

PR #733, opened 2016-07-28
fetch_job: queue correct queue

### LINKED ISSUE #728: fetch_job has wrong behaviour
I'm kind of amazed this hasn't come up before. It just caused us a very serious error in production.

`queue.fetch_job(job_id)` has no check that the job is actually in that queue, so for example `failed_queue.fetch_job(maybe_this_job_failed_id)` will return a job even if the job hasn't failed!

Surely `fetch_job` should check that `job_id` is in that queue before getting and returning the job?

samuelcolvin:
fix #728.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -107,9 +107,12 @@ class Queue(object):
 
     def fetch_job(self, job_id):
         try:
-            return self.job_class.fetch(job_id, connection=self.connection)
+            job = self.job_class.fetch(job_id, connection=self.connection)
         except NoSuchJobError:
             self.remove(job_id)
+        else:
+            if job.origin == self.name:
+                return job
 
     def get_job_ids(self, offset=0, length=-1):
         """Returns a slice of job IDs in the queue."""
--- a/tests/test_queue.py
+++ b/tests/test_queue.py
@@ -462,6 +462,32 @@ class TestQueue(RQTestCase):
 
         self.assertEqual(q.job_ids, [])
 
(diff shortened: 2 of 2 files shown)


samuelcolvin:
Sadly there's no neat way to check for containment on a redis list so this seems like the simplest approach.

rq/queue.py
```diff
@@ -106,6 +106,9 @@ def is_empty(self):
         return self.count == 0
 
     def fetch_job(self, job_id):
+        if job_id not in self.job_ids:
```
selwin: In addition to adding an extra Redis call, this line will be really expensive to run on a long queue because it will fetch all `job_ids` in this queue and checks whether `job_id` is in that list.

A more performant approach would be to modify Line 113 to:
```python
try:
    job = self.job_class.fetch(job_id, connection=self.connection)
    if job.queue.name != self.name: # Something like this
        return
    return job
except NoSuchJobError:
```

Could you please update the PR?

This change is not backwards compatible so we'll have to note this in the changelog along with https://github.com/nvie/rq/pull/743#issuecomment-241607859

selwin: @samuelcolvin any update on this PR?

samuelcolvin: fixed. I think I tried this to start with by couldn't find the reference to the queue form the job because it's rather confusingly called `.origin`.

selwin: I agree it's not intuitively named. Thanks!

PR #737, opened 2016-08-08
Update worker.py

suxianbaozi:
fix this error:
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe7 in position 248: ordinal not in range(128)


selwin:
@suxianbaozi could you please fix the failing tests?

selwin:
And could you please add a test case for this @suxianbaozi ?

selwin:
This should be fixed in https://github.com/nvie/rq/pull/817

PR #741, opened 2016-08-12
custom connection class

julyclyde:
dependency injection

### DIFF
--- a/rq/cli/cli.py
+++ b/rq/cli/cli.py
@@ -37,7 +37,7 @@ config_option = click.option('--config', '-c',
                              help='Module containing RQ settings.')
 
 
-def connect(url, config=None, connection_class='redis.StrictRedis'):
+def connect(url, config=None, connection_class=StrictRedis):
     if url:
         return connection_class.from_url(url)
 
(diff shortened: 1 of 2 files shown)


rq/cli/cli.py
```diff
@@ -37,12 +37,12 @@
                              help='Module containing RQ settings.')
 
 
-def connect(url, config=None):
+def connect(url, config=None, connection_class='redis.StrictRedis'):
     if url:
-        return StrictRedis.from_url(url)
+        return connection_class.from_url(url)
```
proofit404: attribute error: str does not have attribute from_url.

julyclyde: I should have tested it before PR.

selwin:
May I know the rationale behind this change?

julyclyde:
I want to use RedisCluster class instead.

selwin:
@julyclyde alright, merged. Could you please open a PR to `gh-pages` branch documenting this change with an example on how to use this feature with RedisCluster? I think that will be really useful. I also recently got an inquiry about this https://github.com/ui/django-rq/issues/177

Issue #739, opened 2016-08-11
Race condition with job dependencies


th3hamm0r:
I stumbled across this bug:
If a new job B with a dependency to job A is inserted during job A is executed, it may happen, that job B is never executed.
The problem is that `enqueue_dependents` in worker.py ([L682](https://github.com/nvie/rq/blob/master/rq/worker.py#L682)) is called outside of the pipeline which changes the job's status to finished. `enqueue_dependents` just does not use the pipeline (there is a "todo" for that: [L293](https://github.com/nvie/rq/blob/master/rq/queue.py#L293)).
The redis-watch in `enqueue_call()` ([L204](https://github.com/nvie/rq/blob/master/rq/queue.py#L204)) does not help in this case, because it executes normally during the job is in a non-FINISHED state.

I found https://github.com/nvie/rq/issues/387 about race conditions but it seems that result of the discussion was, that it should be safe (https://github.com/nvie/rq/issues/387#issuecomment-50255180). But actually it isn't...

th3hamm0r:
As a quick fix I think I will move the line with `queue.enqueue_dependents(job, pipeline=pipeline)` to follow the `pipeline.execute()` call.
This should avoid the race condition, but of course it makes it a bit less fail-safe: if the worker crashes after finishing, but before enqueuing the dependents, we again have a "hanging" dependency chain.

To fix this issue correctly, I think the worker-side also needs a redis-watch (on the `dependents_key`). But the actual code change may be a bit more complex (eg. handling of `WatchError`) and for that I've too few knowledge about rq's code base.

th3hamm0r:
Fyi: I've started an approach to fix the issue with an additional redis-watch in the worker, which should be failsafe too: https://github.com/th3hamm0r/rq/commit/7b8d4e075e30ca64fa8caa09867624cf2480c983
I'm currently testing it in a project.

selwin:
Thanks for working on this. Could you please open a PR when you find a solution?

th3hamm0r:
@selwin I think I've found a solution (#742)

selwin:
Fixed in https://github.com/rq/rq/pull/742

PR #742, opened 2016-08-13
#739 fix race condition

th3hamm0r:
I completely pipelined `enqueue_dependents` which still works for standalone calls, but also fixes the race condition (see #739). I also had to fix some missing pipeline usages.
Further I found a way to test that issue.


rq/queue.py
```diff
@@ -280,33 +280,61 @@ def enqueue_job(self, job, pipeline=None, at_front=False):
...
+        pipe = pipeline if pipeline is not None else self.connection._pipeline()
+        dependents_key = job.dependents_key
+
         while True:
-            job_id = as_text(self.connection.spop(job.dependents_key))
-            if job_id is None:
+            try:
+                # if a pipeline is passed, the caller is responsible for calling WATCH
+                # to ensure all jobs are enqueued
+                if pipeline is None:
```
selwin: I think we can always call `pipe.watch()` here to contain the logic for enqueueing dependent jobs in this method.

`Worker.perform_job` already does way too many things and I don't want to make it even more complex. Would it be possible to revert the changes to `Worker.perform_job`?

th3hamm0r: I don't understand it completely:
The optional `pipe.watch()` is part of the code that allows to call `enqueue_dependents` standalone. Ok that's some kind of nice to have which could be kicked, since it is only used by the worker, which needs the pipeline to be race condition free.
But even if we kick the possibility of calling it standalone, I think the `watch` should be called by the worker to understand, where the `WatchError` can come from. But ok, that could be documented?
Reverting the code in `Worker.perform_job` adds the race condition again (enqueuing of the dependants AND the status change has to happen in one transaction). So where should I put it?

selwin: I see your point.

Let me think about this a bit more and see if we can come up with a better way.

rq/worker.py
```diff
@@ -667,24 +669,36 @@ def perform_job(self, job, queue):
...
-                    job.save(pipeline=pipeline)
 
-                    finished_job_registry = FinishedJobRegistry(job.origin,
-                                                                self.connection)
-                    finished_job_registry.add(job, result_ttl, pipeline)
+                while True:
+                    try:
+                        # if dependencies are inserted after enqueue_dependents
+                        # a WatchError is thrown by execute()
+                        pipeline.watch(job.dependents_key)
```
selwin: I think we can limit the scope of the `while` loop and `try/except WatchError` to `queue.enqueue_dependents`. That way, it will be much easier to understand since the only place that can raise `WatchError` is `queue.enqueue_dependents()`.

Could you also move the entire block of code that handles successful job execution to a separate method to make this method easier to digest? We already have `handle_job_failure` so I think we can call it `handle_successful_job_execution` (I'm terrible at naming so better suggestions are welcome :)).

th3hamm0r: That does not work: The `try/except` in `enqueue_dependents` was only some kind of "feature". For the usual case, where `enqueue_dependents` is called by the worker, the outer try/except is important, because the `WatchError` is actually raised at L698 (worker.py, below). If only `enqueue_dependents` is watched, there is again a possible race-condition if a job was enqueued afterwards, just before the job is finished ;)
But maybe we could find a compromise: extract the finalizing logic (into `handle_successful_job_execution`) as you suggested, but also move the code of `enqueue_dependents` into it. Why the latter one? The `watch/multi` commands are tightly coupled, but atm `watch` is called in the worker, and `multi` is called in `enqueue_dependents`. Maybe it is just not possible or easy to understand, if this happens in two different places. Merging the code into one new method in the worker could make the code much clearer.

selwin: I still think we can make the scope of the `try/except` here smaller. Perhaps like this:
```python
# Other book keeping stuff in done outside the while loop
while True:
    try:
        queue.enqueue_dependents(job, pipeline=pipeline)
        pipeline.execute()
    except WatchError:
        continue
```
The code above will ensure that all dependents are enqueued and changes to job status/deletion are committed in a single transaction. I think that should work.

Could you please update this branch and move the logic to `handle_successful_job_execution` as we discussed? I hope to get this into the next release :)

th3hamm0r: I'm sorry to interrupt you again, but that doesn't work either, respectively it doesn't fix the race condition. What if the following happens sequentially:

1) Your suggested block is executed, so all current dependencies are enqueued. The job is still not finished, as this should happen in a separate transaction as you suggested.
2) afterwards a new dependency is enqueued (while the worker is running)
3) Now the worker executes the job status/deletion changes in another transaction and sets the job to FINISHED

The result is a deferred job that gets never executed.
We have to enqueue the jobs AND change the status in ONE atomic transaction, to avoid a race condition. Also important related to redis is, that if a `WatchError` occurs, you have to replay all statements related to the transaction (normally starting with a `watch`) again.

So I think we could only extract the code and maybe find some other changes (eg. move all the code from `enqueue_dependents` into `handle_successful_job_execution`) to improve the readability.

AshwinRamesh:
Sorry for interrupting this thread, but I was considering using the dependency feature in a production project at work with v 0.6.0. I came across this thread. Does this mean that I should avoid using the feature until a fix is pushed?

rq/worker.py
selwin: @th3hamm0r no worries. Your interruption ensure that we don't make silly mistakes :).

Sorry for not being clear, perhaps I should have made it clear in my previous suggestion that we wouldn't call `pipeline.execute()` on the book keeping stuff so we can enqueue dependents and commit changes to job in one single transaction. I was thinking something like:
```python
self.set_current_job_id(None, pipeline=pipeline)

if result_ttl != 0:
    job.set_status(JobStatus.FINISHED, pipeline=pipeline)
    job.save(pipeline=pipeline)

finished_job_registry = FinishedJobRegistry(job.origin, self.connection)
job.cleanup(result_ttl, pipeline=pipeline, remove_from_queue=False)
started_job_registry.remove(job, pipeline=pipeline)

while True:
    try:
        queue.enqueue_dependents(job, pipeline=pipeline)
        pipeline.execute()
    except WatchError:
        continue
```
But I suppose it wouldn't work either because as you pointed out, we have to replay all statements when `WatchError` occurs.

Would this work then?
```python
while True:

    self.set_current_job_id(None, pipeline=pipeline)

    if result_ttl != 0:
        job.set_status(JobStatus.FINISHED, pipeline=pipeline)
        job.save(pipeline=pipeline)

    finished_job_registry = FinishedJobRegistry(job.origin, self.connection)
    job.cleanup(result_ttl, pipeline=pipeline, remove_from_queue=False)
    started_job_registry.remove(job, pipeline=pipeline)

    try:
        queue.enqueue_dependents(job, pipeline=pipeline)
        pipeline.execute()
    except WatchError:
        continue

    break
```
The scope of `try/except` is much smaller since `enqueue_dependents` is the only place where the exception could be raised. This also allows dependents to be added until right before the dependents are actually enqueued.

selwin:
@AshwinRamesh it depends on the types of usage I suppose. This feature has been part of RQ for quite a while and used in production by a number of users.

rq/worker.py
th3hamm0r: It looks tempting to reduce the lines of code inside the try/except, but this does not work either. Redis' optimistic locking protects against concurrent changes between `WATCH` and `EXEC` (http://redis.io/topics/transactions). This is why `pipeline.watch(job.dependents_key)` has to come before any pipeline operations ;)

The `MULTI` command, which I've added to the `queue.enqueue_dependents`, puts the transaction again into the pipelined mode. So to be most efficient, we should leave the `watch()` + `enqueue_dependents()` at the beginning (so that's the reason I've moved it).
But I'm still convinced that we should move that code into our new `handle_successful_job_execution` ;)

selwin: Yeah, the code should definitely be moved into `handle_successful_job_execution`, regardless of whether we could simplify the code.

I re-read the code snippet I wrote and I still can't quite get why it wouldn't work. Dependent jobs should still be modifiable until `enqueue_dependents` is called and `pipeline.execute()` would ensure that `job` will be cleaned up in the same transaction.

I'll try to re-read this again this weekend. In the meantime, if you have the time, moving the code block into `handle_successful_job_execution` would be great :)

Thanks!

th3hamm0r: The problem with the code above is that the protected code starts with redis' `WATCH` call. I don't know the used redis-library, so I don't know what happens to the statements before the `watch()` call, but it is probably wrong or undefined. It is up to the library what happens with the statements before the `watch()` call. So if a concurrent modification occurs, the library may even commit the changes before `watch()`, so the job in redis is actually FINISHED, but the loop would restart the transaction.
So the only way is to:

1. call the `watch()` command, which sets the connection into the immediate-execution mode
2. get all the necessary information: for us it is the dependency list together with the necessary metadata (happens in `enqueue_dependents`)
3. call `multi()` to get back into the pipelined mode for efficiency
4. enqueue all dependencies and finish the job
5. commit the transaction with `execute`

If the transaction fails, nothing is written to redis and we have to restart again. But this "all or nothing" principle protects us from any inconsistency.

Ok, I will move the code.

selwin: @th3hamm0r any update on moving this block of code to `handle_successful_job_execution`? :)

th3hamm0r: @selwin hey sorry, I was quite busy the last weeks, but I will try to make the change this week.

selwin: No worries. Thanks for the contribution :)

th3hamm0r:
@selwin I've pushed my changes. I used the name "handle_job_success" which fits to the existing `handle_job_failure()`.
Actually I did some more refactorings and reduced the size of the pipeline context.
Unfortunately one test failed with pypy, I will look into this. Beside this, what do you say?

th3hamm0r:
@selwin Ok, I've tried the tests with pypy on my computer and it worked, maybe you an trigger the travis build again. I don't have access to it.

selwin:
@th3hamm0r I restarted the build and the pypy test passed. Thanks for the patch, I'll review it again this weekend since this is quite a meaty one.

selwin:
@th3hamm0r sorry for the long hiatus. I have reviewed the code and apart from some cosmetic changes, I think it's ready for merging. Could you please fix the merge conflicts (it should be relatively minor) so I can merge this in and credit you with the pull request?

Otherwise, I can also pull in your changes manually.

th3hamm0r:
@selwin done :)

selwin:
Thanks @th3hamm0r!

PR #743, opened 2016-08-13
allow executing the job immediately if the dependency does not exist

th3hamm0r:
It's a bit related to #742. For users of rq it is currently not possible, to add dependencies in a safe way. Atm it may happen that the parent job expires just before the dependency is added. In this case, the new dependency stays in the deferred state forever.
This change does not touch the existing logic, which can be helpful too (eg. when creating a dependency chain first), but allows to add dependencies where the caller knows that the parent job existed in the past.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -175,7 +175,7 @@ class Queue(object):
 
     def enqueue_call(self, func, args=None, kwargs=None, timeout=None,
                      result_ttl=None, ttl=None, description=None,
-                     depends_on=None, job_id=None, at_front=False, meta=None):
+                     depends_on=None, require_dependency=True, job_id=None, at_front=False, meta=None):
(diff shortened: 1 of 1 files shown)


selwin:
Actually, I'd rather that RQ raises an error in the case of invalid job dependency so it can be resolved by the user like this:
```python
try:
    q.enqueue(foo, depends_on=invalid_job)
except InvalidJobDependency: 
    q.enqueue(foo)
```
It feels more pythonic this way. I realize that this is a backwards incompatible change, but I think this change is better for the long term.

th3hamm0r:
Sounds good, I will try to make that change (reverting the new param, adding exception, right?) during the week.

selwin:
Thanks!

th3hamm0r:
@selwin I've created a new PR #747, closing this one.

PR #744, opened 2016-08-22
Windows console fix

debjan:
On Windows console, Worker outputs something like this:

```
*** Listening on [32mdefault[39;49;00m...
```

This patch disables ANSI on Windows if `colorama` package is not installed.
If `colorama` is installed, Windows support [is initialized](https://github.com/tartley/colorama#usage).

### DIFF
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -15,11 +15,18 @@ import logging
 import sys
 from collections import Iterable
 
+try:
+    import colorama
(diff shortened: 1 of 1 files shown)


debjan:
Ah, this package does not support Windows at all: https://github.com/nvie/rq/blob/master/rq/worker.py#L514

Closing...

PR #747, opened 2016-08-25
Raise an exception if a given dependency does not exist

th3hamm0r:
Adapted some tests to the change: the dependency has to be saved first.

This is a replacement for the PR #743 (see https://github.com/nvie/rq/pull/743#issuecomment-241607859).


selwin:
Thanks for the PR!

PR #748, opened 2016-09-02
registry.get_job_ids should add timestamp

Iliker:
registry.get_job_ids can not custom cleanup time if no timestamp.

### DIFF
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -50,9 +50,9 @@ class BaseRegistry(object):
         return [as_text(job_id) for job_id in
                 self.connection.zrangebyscore(self.key, 0, score)]
 
-    def get_job_ids(self, start=0, end=-1):
+    def get_job_ids(self, start=0, end=-1, timestamp=None):
         """Returns list of all job ids."""
-        self.cleanup()
+        self.cleanup(timestamp=timestamp)
         return [as_text(job_id) for job_id in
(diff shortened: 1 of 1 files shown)


rq/registry.py
```diff
@@ -130,16 +130,16 @@ def __init__(self, name='default', connection=None):
         super(DeferredJobRegistry, self).__init__(name, connection)
         self.key = 'rq:deferred:{0}'.format(name)
 
-    def cleanup(self):
+    def cleanup(self, timestamp=None):
```
selwin: This method is not used. Could you please revert this change?

rq/registry.py
```diff
@@ -50,9 +50,9 @@ def get_expired_job_ids(self, timestamp=None):
         return [as_text(job_id) for job_id in
                 self.connection.zrangebyscore(self.key, 0, score)]
 
-    def get_job_ids(self, start=0, end=-1):
+    def get_job_ids(self, start=0, end=-1, timestamp=None):
         """Returns list of all job ids."""
-        self.cleanup()
+        self.cleanup(timestamp=timestamp)
```
selwin: This change should be reverted.

theodesp:
@Iliker Can you explain a little bit more why do we need to add this change?

selwin:

PR #764, opened 2016-12-01
fix issue with FailedQueue object could not fetch failed jobs.

pnuckowski:
Unfortunately `FailedQueue` object is unable to fetch failed jobs.

[django-rq](https://github.com/ui/django-rq) rely on failed queue


selwin:
Fixed in https://github.com/nvie/rq/pull/765

PR #768, opened 2016-12-06
Use HTTPTransport transport so sentry reporting works by default

zyegfryed:
The Threaded transport (default) is not reliable, breaking the error reporting. This PR switch it to the HTTPTransport bringing a working out-of-the-box experience.
I can also add a `--sentry-transport-class` option to the CLI if you want/prefer.


selwin:
Thanks!

PR #770, opened 2016-12-08
Don't run lrem directly if we have a pipeline

mtdewulf:
Not sure if this is correct, but I was looking through the code and noticed that if a pipeline is passed to `Queue.cancel` we add an `lrem` call to the pipeline and then also call `lrem` immediately on the connection.  I think we just want to add it to the pipeline in that case?


selwin:
Your observation seems to be correct. Thanks!

PR #774, opened 2016-12-20
allow for already dead horse on heroku worker termination

samuelcolvin:
I sometimes wonder if I'm the only person who actually bothers to writing unit tests for code that uses rq in `async` mode.

This fix prevents the heroku worker hanging indefinitely on termination of the process when the horse is already down. I guess my fault when I first submitted the heroku worker.

Further tests should not be required as the logic in `kill_horse` is taken from `request_force_stop` where is was previously tested.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -352,6 +352,19 @@ class Worker(object):
         signal.signal(signal.SIGINT, self.request_stop)
         signal.signal(signal.SIGTERM, self.request_stop)
 
+    def kill_horse(self):
+        """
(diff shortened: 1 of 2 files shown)


samuelcolvin:
Fixed and also skipped the test which broke the last release commit.

As I've said [before](https://github.com/nvie/rq/pull/715#issuecomment-223752616) continuing to support is becoming a real burden, see my bodge in 763a304 for an example.

selwin:
@samuelcolvin sorry for the late reply. I don't mind dropping Python 2.6 in the next release. I think it's also about time RQ graduates to 1.0 :)

PR #778, opened 2016-12-27
fix: set HTTP transport as default for sentry

Atala:
Related to nvie/rq#777

AFAIK, nobody would like to use the default async transport as it won't work. Maybe we can add a parameter `--sentry-transport-class`.

(and thanks for rq!)


selwin:
Thanks!

askaliuk:
When do you plan to release this one?
Could you suggest how to patch 0.7.1 with this fix?

PR #781, opened 2017-01-06
Trigger a refresh after job execution

WeatherGod:
* Prevents clobbering of user-supplied metadata
  modified during the job execution
* Allows the "Storing arbitrary data on jobs" example at http://python-rq.org/docs/jobs/ to actually persist.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -677,7 +677,8 @@ class Worker(object):
                     result_ttl = job.get_result_ttl(self.default_result_ttl)
                     if result_ttl != 0:
                         job.set_status(JobStatus.FINISHED, pipeline=pipeline)
-                        job.save(pipeline=pipeline)
+                        # Don't clobber the user's meta dictionary!
+                        job.save(pipeline=pipeline, include_meta=False)
 
                         finished_job_registry = FinishedJobRegistry(job.origin,
                                                                     self.connection)
@@ -705,7 +706,6 @@ class Worker(object):
         try:
             with self.death_penalty_class(job.timeout or self.queue_class.DEFAULT_TIMEOUT):
                 rv = job.perform()
-            job.refresh()
 
             job.ended_at = utcnow()
 
@@ -719,7 +719,6 @@ class Worker(object):
                 started_job_registry=started_job_registry
             )
         except Exception:
-            job.refresh()
             self.handle_job_failure(
                 job=job,
                 started_job_registry=started_job_registry
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -705,6 +705,7 @@ class Worker(object):
         try:
             with self.death_penalty_class(job.timeout or self.queue_class.DEFAULT_TIMEOUT):
                 rv = job.perform()
+            job.refresh()
 
             job.ended_at = utcnow()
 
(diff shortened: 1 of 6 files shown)


WeatherGod:
Is there an existing unit test for this feature that isn't working, or is a new test needed? This bug caused me to go down a rabbit hole for about 6 hours today.

WeatherGod:
The failed test seems to be transient.

I found the test "test_custom_meta_is_persisted" in test_job.py, but really, that only tests that the `save()` method works as intended. There is also the test_enqueue_meta_arg() in test_queue.py, but the metadata is set prior to the job execution, so that isn't the same thing. The issue at hand is that the worker is the one clobbering the data in the redis server.

I will try to make an integration test that properly tests this feature.

WeatherGod:
sigh... right... no set literals in 2.6. I'll fix that.

WeatherGod:
The failed test is again the pypy one, and it is the same test that seems to have been failing for some time now in other PRs. As for the coverage going down, I haven't a clue how that is happening, as my two tests specifically cover each of the two lines of code I added in worker.py.

WeatherGod:
ping

WeatherGod:
ping

WeatherGod:
This PR closes the separately reported issue #784

rq/worker.py
```diff
@@ -698,6 +698,7 @@ def perform_job(self, job, queue):
         try:
             with self.death_penalty_class(job.timeout or self.queue_class.DEFAULT_TIMEOUT):
                 rv = job.perform()
+            job.refresh()
```
selwin: Could you please move `job.refresh()` to just before `job.save()` inside `handle_job_success()`? This should only be called if job needs to be persisted in Redis.

What would be even better is if we create `job.refresh_meta()` that uses `redis.hget()` to fetch only the `meta` field from Redis.

rq/worker.py
```diff
@@ -711,6 +712,7 @@ def perform_job(self, job, queue):
                 started_job_registry=started_job_registry
             )
         except Exception:
+            job.refresh()
```
selwin: Could you please move `job.refresh()` to just before `job.save()` inside `handle_job_failure()`?

rq/worker.py
WeatherGod: The problem with moving the refresh call further in is that it would clobber any information that rq sets. From a high-level design point of view, it makes sense to do the save right before the job execution, and the fetch right afterwards. The state stored in redis should be considered canonical, and so when passing around the control of the state between the worker and the job, synchronization should occur at each transfer to prevent ever having an out-of-date state.

Another consideration which I am not qualified to assess. For those who subclass Worker, where should the refresh() call go in order to ensure that this behavior is maintained in the subclasses?

selwin: After thinking about this a bit further, I have a better solution. Instead of reading `meta` from Redis and saving it (race condition could also happen here), we should modify the `save()` call in `handle_job_success()` to exclude `meta` so the job key in Redis will keep its `meta` value.

Something like this would be great: `job.save(include_meta=False)`.

WeatherGod: That could work. It should probably be explicitly documented somewhere that the "meta" dict should never be modified by a Worker (or its subclasses), and that any other Job attributes (i.e., is_finished, return_value, etc.) should never be modified by any python code executed by the Job.

It would probably be useful in figure out a way to enforce those separation of concerns in the future, but I would settle for just this solution and some documentation instead. I can write up the function, where should the documentation go?

selwin:
Thanks!

PR #785, opened 2017-01-23
Closes issue #784: Avoid losing meta job information after successful finish.

### LINKED ISSUE #784: Worker.handle_job_success overwrites job meta information
When a job completes successfully, its `meta` dictionary is overwritten with old values. The use case:

1. When a job is queued, it's done with `meta={'progress: 'Queued'}`.
2. When a job is started, the work function updates with `job.meta['progress'] = 'Started'; job.save()`
3. As the job progresses, its `meta['progress']` info is updated with progress information in the same way.
4. The job completes, and at that point the job on Redis has some value of `meta['progress']` which indicates how far it had progressed.
5. `Worker.handle_job_success` is called with an in-memory copy of the job which has the original progress value of `Queued`. It saves this job instance with an updated status of `finished`, but *omits  to preserve the meta data* before doing this. As a result, the job's `meta` dict is overwritten with the original `{progress: Queued}` value, overwriting any other information in the meta dict.

IMO `Worker.handle_job_success` should get the job's `meta` data and restore it while saving the updated status. See pull request #785 for a fix.

vsajip:
As described in the issue.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -18,11 +18,11 @@ from redis import WatchError
 
 from rq.compat import as_text, string_types, text_type
 
-from .compat import PY2
+from .compat import PY2, decode_redis_hash
 from .connections import get_current_connection, push_connection, pop_connection
 from .defaults import DEFAULT_RESULT_TTL, DEFAULT_WORKER_TTL
 from .exceptions import DequeueTimeout, ShutDownImminentException
-from .job import Job, JobStatus
+from .job import Job, JobStatus, unpickle
 from .logutils import setup_loghandlers
 from .queue import Queue, get_failed_queue
 from .registry import FinishedJobRegistry, StartedJobRegistry, clean_registries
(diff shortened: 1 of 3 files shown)


vsajip:
Tests failing, needs more work.

vsajip:
Test is failing on PyPy, but appears unrelated to this change.

WeatherGod:
I think my PR #781 solves the same problem in a simpler way?

WeatherGod:
In addition, this PR doesn't protect against the same clobbering if the job fails

vsajip:
> I think my PR #781 solves the same problem in a simpler way?

Sorry @WeatherGod, I didn't spot your PR before creating mine.

Initially, I tried just adding a `job.refresh()` in `handle_job_success()`, but that caused the `result` to be overwritten and lost. Looks like I just put the `refresh()` in the wrong place.

> In addition, this PR doesn't protect against the same clobbering if the job fails

That's true, too. And you've covered the failure case in the tests, too. So I think I'll close my PR in favour of yours! Nice one.

PR #788, opened 2017-01-30
Better support for async job payloads

friedcell:
Make it easier to support async methods as job payloads by removing job interactions to separate functions that you can override in local implementation of AsyncJob and AsyncQueue.


friedcell:
Updated to match master...

selwin:
Thanks!

PR #786, opened 2017-01-23
Allow passing backend classes from CLI and other APIs

jezdez:
This is related to https://github.com/jezdez/Flask-RQ2/issues/2 and https://github.com/jezdez/Flask-RQ2/issues/12. While debugging those issues I've found many inconsistencies how the backend classes are used throughout the RQ code base, basically making it currently impossible to fully override job, queue, worker or connection classes.

This should be backward-compatible, but any testing in real-world projects would be appreciated.

PR #790, opened 2017-02-02
CLI: RQ_CONFIG environment variable

akx:
Add the `RQ_CONFIG` environment variable as an alias for `-c`/`--config`.

```shell
$ echo 'print("hello")' > my_config_module.py
$ env RQ_CONFIG=my_config_module rq worker
hello
13:37:00 RQ worker '...' started, ...
```


selwin:
Could you please document this feature in the `Using a config file` section of http://python-rq.org/docs/workers/ ?

akx:
@selwin Sure thing. ... Where are the sources for the docs?

akx:
Anyway, looks like this was implemented as side effect in #786, so closing this.

PR #796, opened 2017-02-28
Fix pip install from HEAD of origin/master

davidthewatson:
The pip install -e directions don't work for users without write access to the repo. This documentation fixes that problem.


selwin:
Thanks!

PR #798, opened 2017-03-07
Make `Queue.enqueue_job()` execute job immediately if `async=False`

flacerdk:
Currently, the job is being performed inside `enqueue_call()`, which
means that `async=False` has no effect if `enqueue_job()` is called
directly. This commit fixes that.


selwin:
@flacerdk could you please fix the conflicts?

flacerdk:
@selwin done

selwin:
Thanks!

PR #808, opened 2017-03-31
add sentinel support

strawposter:


selwin:
Before I merge this in, could you please document this feature?

selwin:
Merged, thanks! What I mean by documentation is actually by modifying the `gh-pages` branch of this repo so it shows up on http://python-rq.org . Would be great if you can add that :)

strawposter:
I'll do it.

Issue #804, opened 2017-03-22
Workers don't send heartbeat during suspension.


red-crown:
If you suspend workers for longer than the worker ttl, they will no longer appear in the monitoring dashboards. We should continue sending the heartbeat inside the check_for_suspension loop.

selwin:
You're right. Mind opening a PR for this?

red-crown:
Do you see a problem with just sending the heartbeat every second regardless of worker ttl?

red-crown:
I'm also wondering if we should look for a better place for the suspension check. As someone else has pointed out, idle workers will always process one additional job after they have been "suspended" due to how the queue blocks for a job. I've briefly tried having the main worker sending something like a USR1 signal to the child to notify it but had some issues when the system call is interrupted.

In my scripts, I've basically resorted to suspending the workers and then pushing a number of null or "fake" jobs onto the front of the queues equal to the number of workers. My use-case for this has been on either code updates/deployments or with dynamically auto-scaled worker servers based on work load.

selwin:
Yes, I'm aware of this limitation and it's something that we should fix.

Shutdowns (both warm and cold) are handled by the parent worker that kills the child worker if the child worker is busy. Can we use a similar method, using a different signal?

selwin:
Fixed in https://github.com/rq/rq/pull/896

PR #812, opened 2017-04-05
send heartbeat during supsension check

red-crown:
Have workers continue to send heartbeat during suspension check loop. #804

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -410,6 +410,7 @@ class Worker(object):
                 self.set_state(WorkerStatus.SUSPENDED)
                 notified = True
             time.sleep(1)
+            self.heartbeat()
 
         if before_state:
             self.set_state(before_state)


selwin:
Do you mind modifying check_for_suspension() and heartbeat() so they can both be done with Redis pipeline?

red-crown:
sure

red-crown:
Why do we want to make it work with redis pipeline? Inside check_for_suspension() the new call to heartbeat() is the only place it makes sense.  The internal calls to is_suspended() which is called every second and set_state() need a non-pipeline connection. Also, the only call to check_for_suspension() is within work() which does not accept a pipeline.

Is this for external user calls to check_for_suspension() or is the idea to basically make the heartbeat() call within the suspension check loop "optional"?

selwin:
You're right, it doesn't make sense for `check_for_suspension()` to support `pipeline`.

What I wanted to achieve is for RQ to be able to check whether RQ is suspended and sending `heartbeat ` in one single Redis call (instead of two) to make it more efficient, since this will be done every second.

Looking at the code, I think we can accomplish this by:
1. Modifying `is_suspended(connection, worker)` to accept `worker` argument
2. `worker.heartbeat()` already accepts `pipeline` argument

selwin:
@red-crown I plan on releasing a new version at the end of this month. Would be great if we can get this PR in before then.

red-crown:
Sure thing. Will work on it tonight.

selwin:
@red-crown gentle reminder to update this PR 😀

selwin:
@red-crown ping :)

theodesp:
Is it ok if I push a PR for this? 
What I've done is I've followed this list:
```
  - Modifying is_suspended(connection, worker) to accept worker argument
  - worker.heartbeat() already accepts pipeline argument
```
And added the request in a pipeline.

What I need is to add tests for coverage

selwin:
@theodesp yes, please submit a new PR for this if you can :)

PR #819, opened 2017-04-20
Add waitpid() return value to workhorse error/log messages

akx:
For easier debugging of unexpected exits of workhorses.


selwin:
Thanks!

Issue #811, opened 2017-04-04
job expires during execution


katichev:
Hi all,

I'm using RQ to perform long running jobs. Sometimes I need to update meta during job execution:
```
def long_job():
    # part A: do stuff
    job = get_current_job()
    job.meta['my_progress'] = 0.5
    job.save()
    # part B: do more stuff
    return result
```
If 'part B' duration is longer than ttl, the job key expires.

As I understand, when I call `get_current_job()`, a new job instance is being created and loaded from redis. Which means that `job.ttl` is also returned from storage, and updated key is not persistent after `job.save()`.

Shall we check current job state in `save()` and do not call `cleanup()` if job is in progress?

red-crown:
+1
Or perhaps we could also add a cleanup or "update" option to the `save()` method.

selwin:
What we can do is introduce a `job.update_meta()` method and make it part of the official API to make it easier for users to update `meta` without accidentally triggering other methods like `cleanup()`.

A PR (along with documentation) for this is welcome :)

PR #825, opened 2017-04-29
Add documentation for job.save_meta()

katichev:
This closes nvie/rq#811

### DIFF
--- a/docs/jobs.md
+++ b/docs/jobs.md
@@ -27,7 +27,7 @@ def add(x, y):
 
 ## Storing arbitrary data on jobs
 
-_New in version 0.3.3._
+_Improved in 0.8.0._
 
 To add/update custom status information on this job, you have access to the
 `meta` property, which allows you to store arbitrary pickleable data on the job
@@ -39,21 +39,14 @@ import socket
 def add(x, y):
     job = get_current_job()
     job.meta['handled_by'] = socket.gethostname()
-    job.update_meta()
+    job.save_meta()
     
     # do more work
     time.sleep(1)
     return x + y
 {% endhighlight %}
 
-<div class="warning">
-    <img style="float: right; margin-right: -60px; margin-top: -38px" src="{{site.baseurl}}img/warning.png" />
-    <strong>Warning!</strong>
-    <p>
-The `.update_meta()` was introduced in RQ 0.7.2. You can use the `.save()` method instead 
-for RQ <= 0.7.1. For that case you also have to set the job `ttl` greater than the job `timeout`, otherwise the job may dissappear from the queue during execution.
-    </p>
-</div>
+
 ## Time to live for job in queue
 
 _New in version 0.4.7._
--- a/docs/jobs.md
+++ b/docs/jobs.md
@@ -39,11 +39,21 @@ import socket
 def add(x, y):
     job = get_current_job()
     job.meta['handled_by'] = socket.gethostname()
-    job.save()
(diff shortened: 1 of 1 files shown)


docs/jobs.md
```diff
@@ -39,11 +39,21 @@ import socket
...
+    time.sleep(1)
     return x + y
 {% endhighlight %}
 
-
+<div class="warning">
+    <img style="float: right; margin-right: -60px; margin-top: -38px" src="{{site.baseurl}}img/warning.png" />
+    <strong>Warning!</strong>
+    <p>
+The `.update_meta()` was introduced in RQ 0.7.2. You can use the `.save()` method instead
```
selwin: Could you please change this to `RQ 0.8.0`? Thanks!

docs/jobs.md
```diff
@@ -39,11 +39,21 @@ import socket
...
     return x + y
 {% endhighlight %}
 
-
+<div class="warning">
+    <img style="float: right; margin-right: -60px; margin-top: -38px" src="{{site.baseurl}}img/warning.png" />
+    <strong>Warning!</strong>
+    <p>
+The `.update_meta()` was introduced in RQ 0.7.2. You can use the `.save()` method instead 
+for RQ <= 0.7.1. For that case you also have to set the job `ttl` greater than the job `timeout`, otherwise the job may dissappear from the queue during execution.
```
selwin: Perhaps I'm missing something, but I'm not sure why we need to mention about needing to have `ttl` greater than `timeout` here.

katichev: Hmm... My idea was to refer somehow to the #811. If I `get_current_job` and then `save` it during execution, `cleanup` sets back expiration time to the Redis key. If the job lasts more than `ttl` since that moment, the Redis key expires. So if I don't want the key to expire, I have to set `ttl > timeout` (i.e. an expected max job duration).
@selwin, what do you think, should we mentioned it here?

selwin: I think that's a different issue entirely.

The real fix is to modify the job stack so that "get_current_job" returns the current job (now it returns the current job_id) without trying to recreate it from Redis (which may have expired). Please open a separate PR for this :)

docs/jobs.md
selwin: Instead of mentioning that update_meta was introduced in 0.8, I think we should just change the "new in version 0.3.3" to "improved in 0.8.0".

docs/jobs.md
katichev: In current implementation, `self.ttl == -1` during a job execution.  If I did the fix, then any `save()` call inside a job would change `ttl` value in the Redis key to -1.

I would not modify `job.ttl` inside a job at all and remove the implicit `cleanup` call from `save`. Of course, I would add explicit `cleanup` everywhere I need (inside worker.py, queue.py).

However, modification of the job stack may be useful also. `get_current_job` would be faster then.

selwin:
Thanks!

docs/jobs.md
katichev: @selvin, this closed PR is probably is not the best place for discussion. Hovewer, the question is still opened for me, shouldn't I propose an another PR? I mean, is it worth modifying the code in the way I described above, i.e make all "cleanup" calls explicit? Thanks.

selwin: Yes, I think separating the "cleanup" calls from "save" is a good idea. Thanks 😀

PR #826, opened 2017-04-29
Worker.find_by_key should use hmget instead of repeated hget calls.

selwin:


lucastamoios:
Thanks for this improvement! ;)

lucastamoios:
After I deployed this version into production I noticed the django-rq response got even worse than before. Then I noticed HGET is O(1) complexity, and HMGET is O(N).

Is then possible to revert this change?

PR #829, opened 2017-05-05
updating the sys.path for import of a custom class

strawposter:
Without updating the sys.path, import of a custom class will not work

example:
rqworker -c config -P '/home/user/projects/workers' --worker-class custom.CustomWorker --job-class custom.CustomJob --name Custom

### DIFF
--- a/rq/cli/helpers.py
+++ b/rq/cli/helpers.py
@@ -226,7 +226,8 @@ class CliConfig(object):
         self.config = config
 
         if path:
-            sys.path = path.split(':') + sys.path
+            for pth in path:
+                sys.path.append(pth)
 
         try:
             self.worker_class = import_attribute(worker_class)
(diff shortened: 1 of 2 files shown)


rq/cli/helpers.py
```diff
@@ -219,11 +220,14 @@ class CliConfig(object):
     """A helper class to be used with click commands, to handle shared options"""
     def __init__(self, url=None, config=None, worker_class=DEFAULT_WORKER_CLASS,
                  job_class=DEFAULT_JOB_CLASS, queue_class=DEFAULT_QUEUE_CLASS,
-                 connection_class=DEFAULT_CONNECTION_CLASS, *args, **kwargs):
+                 connection_class=DEFAULT_CONNECTION_CLASS, path=None, *args, **kwargs):
```
selwin: Instead of splitting the `path` string by `:`, I think it would be better if we accept multiple `path` arguments like http://click.pocoo.org/5/options/#multiple-values-from-environment-values

So we can invoke RQ like this: `rqworker --path 'foo' --path 'bar'`

PR #831, opened 2017-05-07
Replace job id with job instance in the job stack. Remove implicit "cleanup" call from "job.save"

katichev:
As discussed [here](https://github.com/nvie/rq/pull/825#discussion_r114075946)

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -480,18 +480,18 @@ class Job(object):
 
     def save(self, pipeline=None, include_meta=True):
         """
-        Persists the current job instance to its corresponding Redis key.
+        Dumps the current job instance to its corresponding Redis key.
 
-        Exclude persisting the `meta` dictionary by setting
+        Exclude saving the `meta` dictionary by setting
         `include_meta=False`. This is useful to prevent clobbering
         user metadata without an expensive `refresh()` call first.
 
+        Redis key persistence may be altered by `cleanup()` method.
         """
         key = self.key
         connection = pipeline if pipeline is not None else self.connection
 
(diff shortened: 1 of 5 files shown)


rq/job.py
```diff
@@ -609,9 +606,12 @@ def cleanup(self, ttl=None, pipeline=None, remove_from_queue=True):
...
         elif not ttl:
             return
-        elif ttl > 0:
+        else:
             connection = pipeline if pipeline is not None else self.connection
-            connection.expire(self.key, ttl)
+            if ttl > 0:
+                connection.expire(self.key, ttl)
+            else:
+                connection.persist(self.key)
```
selwin: Do we need to call `persist` here? What issue does this address?

katichev: That's about the line in the docstring for `cleanup`: [If ttl is negative, don't set an expiry to it (persist
          forever)](https://github.com/katichev/rq/blob/explicit_cleanup/rq/job.py#L602)
Before, `cleanup(ttl=-1)` did not really clear the expiration time to unlimited.

The idea was to manage the key expiration in one single place. Now I see that I forget to replace `persist` with `cleanup` [here](https://github.com/katichev/rq/blob/explicit_cleanup/rq/job.py#L554). :)

Anyway I need another one commit. I have two options here: 1. Revert back the change you've mentioned; 2. replace `persist` with `cleanup(ttl=-1)`. What do you suggest?

selwin: I think reverting it back to what it was is the better option (don't do anything, because this already means that the key won't be expired).

katichev:
@selwin, fixed

selwin:
Thanks!

PR #832, opened 2017-05-08
Expand timeout string feature to both result_ttl and ttl, and add days possibility.

yaniv-g:
* Added support to use strings to specify both result_ttl and ttl.
* Added Days as possible timeout string (as 'd')


selwin:
Thanks!

Issue #711, opened 2016-05-29
Expose `Worker.move_to_failed_queue` as a standalone exception handler


gsakkis:
The cli has a nice `'--exception-handler'` option for passing in exception handlers by name. Unfortunately this allows only to replace the default behaviour (sending the job to the failed queue), there's no way to specify the extra handlers in addition to the default.

gsakkis:
It's probably doable but wouldn't say better; I'd either have to reimplement the cli or do without it.

proofit404:
Extending Worker class is a common technique for RQ as far as I can tell. But don't wanna say it must be the only option in this case.

selwin:
I think it'd be good if the cli allows multiple exception handlers to be definer like `rq worker --exception-handler foo --exception-handler bar`. PR welcome :)

gsakkis:
It already does allow it. The problem is that the default exception handler is a Worker method instead of a standalone function so it can't be specified in the cli. 

My proposal is to extend the exception-handler to allow a Worker method as ".{{method_name}}", in this case as ".move_to_failed_queue", but I'm open to other unambiguous schemes. Thus it will be also possible to defined other exception handler methods in a Worker subclass instead of special-casing only `Worker.move_to_failed_queue`.

selwin:
We can simply create a `move_to_failed_queue` function and simply change `worker.move_to_failed_queue` to use that function to keep it backwards compatible.

gsakkis:
That was my first idea but it doesn't seem easy/possible; there's a reason `move_to_failed_queue` is a `Worker` method and not a function since it accesses `self.log` and `self.failed_queue`.

selwin:
We cab to recreate "self.failed_queue" from "job.queue". If "self.log" can't be easily recreated in the function, we can just keep the logging call in the original instance method.

Once we have this reference exception handler function, people should be able to easily duplicate and add their own logging statements to it if needed.

proofit404:
@selwin Can you explain why `rq` worker hold logger instance? I frequently saw approach with loggers stored as global variables. Root logger in the `__init__.py` file, child logger in the `worker.py` file. This approach can solve `self.log` problem. Also maybe it isn't necessary to use `setup_loghandlers` inside worker (we can use it in the `cli` module instead).

selwin:
@proofit404 as far as I know, the logger instance in `Worker` is only there for convenience. Decoupling this from `Worker` like you suggested seems like a good idea.

[...]

PR #833, opened 2017-05-08
Expose move_to_failed_queue from Worker

yaniv-g:
Fixed issue #711 
* Separate move_to_failed_queue from the worker to allow it's use in cli exception handler.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -85,6 +85,13 @@ WorkerStatus = enum(
 )
 
 
+def move_to_failed_queue(job, *exc_info):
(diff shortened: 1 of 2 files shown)


rq/worker.py
```diff
@@ -85,6 +85,13 @@ def signal_name(signum):
 )
 
 
+def move_to_failed_queue(job, *exc_info):
```
selwin: Thanks for the PR. To keep things tidy, could you please move this function to `handlers.py`?

selwin:
Thanks!

Issue #802, opened 2017-03-11
Automatically reenqueue re-entrant jobs


Chronial:
I extended rq for our project with the following functionality:
Functions can be marked as reentrant (via a decorator). If a reentrant task is killed by a Heroku shutdown, it is not moved to the failed queue, but immediately re-enqueued at the begging of the queue that it came from.

I would like to get this functionality merged here, but wanted to ask whether you would want to merge such a feature before investing time into creating a PR.

selwin:
Yes, I think job retries is a nice feature to have. will gladly merge in a PR for this if we can come up with an elegant API and implementation.

Chronial:
I propose this API:
* A decorator `reentrant` in `rq.decorators`. If the job function is decorated with this decorator, it is considered reentrant.
   * An alternative to this would be to add a `reentrant` parameter to `enqueue()`. That would be more in line with the rest of the API, but I don't like this solution as much. From experience reentrancy is a property of the function, not of it's enqueued instance. Maybe a good solution would be to just implement both?
* Change signal handling: On `sigterm` give the worker 10 seconds (configurable) to shut down. After 10 seconds, raise a `WorkerShutdown` exception (that does not inherit from `Exception`, but from `BaseException`).
   * If the job is reentrant, reenqueue it at the front of the queue
   * Otherwise mark the job as failed

guillemcabrera:
Chronial, we'll be also interested on a similar feature. We had something similar implemented in an internal app, but we're facing some minor issues from 0.7.x, I guess due to changes in the job and queue management.

From your comments, I'd rather prefer the `reentrant` parameter for `enqueue()` since it feels more like all other features in this library.

About the reenqueueing policy, I'd also suggest to make it a choice. In some cases, it would be interesting to reenqueue the job at the end of the origin queue not to interfere in other jobs workflow.

Is there any pull request we can look at to see your work? We can try to help there as well 😄

Chronial:
I'm still waiting on @selwin to give some basic input on the design before I proceed.

Chronial:
> About the reenqueueing policy, I'd also suggest to make it a choice. In some cases, it would be interesting to reenqueue the job at the end of the origin queue not to interfere in other jobs workflow.

I don't understand this scenario – how could there be interference in other workflows? The job *was* at the beginning of the queue. Putting it back there can not cause it to interfere with anything it wasn't already interfering with. Putting it at the end of the queue is what causes hard to predict behavior.

[...]

PR #837, opened 2017-05-13
Extend worker shutdown

Chronial:
Implements the approach discussed in #802.

This is only a first draft and does not implement all the extensions discussed in #802. The tests are failing and the new behavior is completely missing tests (and is untested).

Since the `HerokuWorker` already implemented parts of this design, I tried to follow its decisions and be backwards compatible with it, even in cases where I disagreed with its decisions.


selwin:
Thanks for the PR @Chronial. I think immediately moving the job to `FailedQueue` during hard shutdowns is the right move, would be good if we can have a separate PR for this.

At the moment, this PR introduces a retry mechanism for jobs that are killed during hard shutdowns. I think we should take this one step further so that this mechanism works for all failed job executions (so we don't have different APIs for hard shutdowns and failed job executions).

My proposal is to change the API to `queue.enqueue(job, max_retries=5)`. We can then modify `handle_job_failure()` to check whether job should be retried and requeue accordingly.

Chronial:
But I don't want my jobs that failed because somethings broken to be retried. I have no reason to assume that a job that raised an exception would work if I just try it again, but every reason to assume that a job that got interrupted by a shutdown would work if I would just let it run.

What api would you propose for that?

selwin:
For now, let `Worker` calls `handle_job_failure()` on hard shutdowns, which I think is the right behavior. You can attach custom metadata to your jobs and override `worker.handle_job_failure()` on your custom worker class (you already use a custom `HerokuWorker` class anyway) to check whether your metadata exists and put it back in the queue if necessary.

I think it's worth taking the time to design a coherent retry mechanism that addresses common use cases. The proposed reentrant API feels like it's only trying to address the narrow issue of Heroku killing its process after a grace period.

The latest API I'm thinking of if something like this:
```python
```

Chronial:
Would it not be cleaner to handle retries in something like an exception handler as detailed here https://github.com/nvie/rq/issues/802#issuecomment-301645638?

selwin:
@Chronial yes, you could do it that way too. Do give it a try and let me know if it works :)

selwin:

PR #851, opened 2017-06-06
Properly add exception stacktrace

chaffeqa:
According to [logger docs](https://docs.python.org/3/library/logging.html#logging.Logger.debug):

> If exc_info does not evaluate as false, it causes exception information to be added to the logging message. If an exception tuple (in the format returned by sys.exc_info()) or an exception instance is provided, it is used; otherwise, sys.exc_info() is called to get the exception information.

**Note**

> New in version 3.2: The stack_info parameter was added.
> Changed in version 3.5: The exc_info parameter can now accept exception instances.

If you ever want to target newer python versions, those are better candidates

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -716,7 +716,7 @@ class Worker(object):
             self.handle_job_success(job=job,
                                     queue=queue,
                                     started_job_registry=started_job_registry)
-        except Exception:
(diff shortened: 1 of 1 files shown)


chaffeqa:
closing in preference of #363

PR #855, opened 2017-06-19
Windows: Use SIGTERM instead of SIGKILL

floqqi:
The `signal.SIGKILL`-constant is not available on the windows platform. Like in werkzeug (https://github.com/pallets/werkzeug/commit/68e1ba06500914089723a3744a5ab93aa212b3a3) using `signal.SIGTERM` should fix the problem.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -14,6 +14,11 @@ import traceback
 import warnings
 from datetime import timedelta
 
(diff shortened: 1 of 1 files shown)


BennyAlex:
@nvie  @selwin 
can you merge the request, please?

PR #856, opened 2017-06-30
Multi dependencies (my take)

olingerc:
This is the excellent branch created by @jlopex now merged into the current master with a few changes by myself.
- I kept the examples. althoug the tests cover quite a bit
- I extensively tested this on my site in an installation where rq runs since many years. I have used complex dependency graphs over multiple queues and workers. Nothing was lost.
- Things to discuss: If a job is cancelled all jobs depending on it will be cancelled as well. We introduce a CANCELED state. Is this okay?
- All tests pass (except I could not get the py26 and flake8 tests to run on my system)
- This is a starting point only. I am aware that it needs some work and comments.


rq/job.py
```diff
@@ -135,9 +136,15 @@ def create(cls, func, args=None, kwargs=None, connection=None,
...
         job.meta = meta or {}
 
-        # dependency could be job instance or id
-        if depends_on is not None:
-            job._dependency_id = depends_on.id if isinstance(depends_on, Job) else depends_on
+        # dependencies could be a single job or a list of jobs
+        if depends_on is None:
+            depends_on = []
+        if isinstance(depends_on, list):
+            job._dependency_ids = [x.id for x in depends_on]
```
olingerc: Needs to test for job instance as well.

rq/job.py
```diff
@@ -181,17 +188,48 @@ def is_started(self):
...
-        self._dependency = job
-        return job
+        if hasattr(self, '_dependencies'):
+            return self._dependencies
+        self._dependencies = [Job.fetch(
+            dependency_id, connection=self.connection)
+            for dependency_id in self._dependency_ids]
+
+        for job in self._dependencies:
+            job.refresh()
```
olingerc: Why do a refresh here?

rq/job.py
```diff
@@ -435,6 +483,16 @@ def to_date(date_str):
         self.ttl = int(obj.get('ttl')) if obj.get('ttl') else None
         self.meta = unpickle(obj.get('meta')) if obj.get('meta') else {}
 
+        if obj.get('dependency_ids'):
+            self._dependency_ids = as_text(obj.get('dependency_ids', '')).split(' ')
+        else:
+            self._dependency_ids = []
+
```
olingerc: Duplicate code. Did I want dependents_keys?

olingerc: Probably not.

olingerc:
No idea what happened to Travis there. I do not see the same errors on my system. ???

rq/queue.py
```diff
@@ -337,17 +350,30 @@ def enqueue_dependents(self, job, pipeline=None):
...
+                    if dependent.has_unmet_dependencies() is False:
+                        registry = DeferredJobRegistry(dependent.origin,
+                                                       self.connection,
+                                                       job_class=self.job_class)
+                        registry.remove(dependent, pipeline=pipe)
+                        if dependent.origin == self.name:
+                            self.enqueue_job(dependent, pipeline=pipe)
+                        else:
+                            queue = Queue(name=dependent.origin, connection=self.connection)
+                            queue.enqueue_job(dependent, pipeline=pipe)
```
olingerc: Big problem. It can happen that two workers get to this point at the same time. The consequence is that the same job is enqeued twice. In my system it happened about 1 in 100 times but it still is a problem. I am trying to find a solution but until now this code is not production safe !

olingerc:
I've added a watch on the job to be enqueued key during enqueue_dependents. This solves double enqueuing if multiple workers finish two upstream jobs simultaneously.
However, this now makes another test fail: when a job is enqueued during the enqueue_dependents call. I will check that now.

selwin:
Thanks for the PR @olingerc !

I think RQ's current job dependency feature is robust (free of race conditions etc) so we can start taking a look at supporting multiple dependencies soon.

As far as I know, `job.delete_dependents()` ([discussed here](https://github.com/nvie/rq/pull/427#issuecomment-317022979)) is one of the last remaining pieces RQ needs. After that, we can discuss this PR in more detail :)

olingerc:
@selwin that's great. I'll be ready to help finalize this PR

olingerc:
@nvie I realize this is not a priority, but could you please have a look at this if you find some time? Many thanks!

nvie:
Hi @olingerc, I'm not doing any active RQ maintenance work myself anymore, please ping either @selwin or @samuelcolvin (he's a collab since today!)

selwin:
@olingerc we'll need to have `job.delete_dependents()` before we can merge this in.

meowlz:
any ETA when this feature might be incorporated? I desperately need it :7

olingerc:
Sorry. My priorities have changed at this time. I’ll sure come back to this but not for another month at least. Don’t hesitate to take over from my branch or a new one. I was stuck with a race condition when two workers would start the same job and thus create duplicates.

jlopex:
Hey guys, I might have some time to work on this, @olingerc do the race condition have any test case?

olingerc:
That’s great!!! I still use my implfmentation since the race is very rare.
There is a test case at the end of:
https://github.com/olingerc/rq/blob/multi-deps-olingerc/tests/test_queue.py

PR #858, opened 2017-07-14
path option needs to be a shared command to ensure that settings file…

sjtbham:
… can be loaded from subcommands such as requeue, resume etc otherwise sentinel config cannot be passed


selwin:
Thanks!

PR #860, opened 2017-07-23
Added support for windows by doing the following :

leonardo-albertovich:
1) Replaced fork with executing the same process with the same command line plus some environment variables that when detected by WindowWorkers work method will cause it to refresh the job object, execute it and exit
2) Replaced waitforpid with a WaitForSingleOjbect call that uses the tasks timeout to simulate the timeout that cannot be enforced with SIGALRT
3) Added a dummy DeathPenalty class


theodesp:
I think if we want to progress with this PR we need to extract the worker OS fork execution dependent code in a separate folder like `platform` and let the worker class run the code for each OS

trapito:
The problem with that would be that since there is nothing like fork on windows the mechanism cannot be properly separated in that way (unless i'm wrong). However, if you have an idea / concept for it i'd like you to share it and i'd be glad to implement it.

On a side not, i don't quite understand why the heroku unit tests fail in my branch, does anyone have any insight on it?

theodesp:
Yes they fail for me also. I might check it out

trapito:
I've never used it but it seems like it's not foss / available for free so i don't know what the project manager will think about it.

Still, i don't quite understand the lack of input, it's been a month since i submitted the pull request.

theodesp:
I've used it. Its free for open source if thats the case. I also found this one
http://docs.tea-ci.org/usage/overview/
that is open source maybe its worth it

As for the tests failing its because the `SIGRTMIN` signal event is not defined in MacOS. Its a good idea to add platform tests for Travis
```
os:
  - linux
  - osx
```

trapito:
What i don't understand is why do the tests fail for me but not for the other pull request unless they aren't being executed because there are no changes to any of the python files, I'm reading about those CI tools you linked.

trapito:
BIG changes in the last commit, the whole patch has been completely restructured and it's a lot better

trapito:
It seems that the problem with the test case was in fact that under pypy the sleep operation was finishing before the job expired.

selwin:
I don't have a Windows machine handy to test this at. But if this were to be pulled in, it would have to have good test coverage.

trapito:
How woud we proceed with that? Travis CI doesn't support windows, as theodesp mentioned we could give appveyor a try but i think that's something you'd have to set up (i could set it up in my fork and prepare everything beforehand).

selwin:
@trapito I think a good solution would be to put `WindowsWorker` in a separate repository.

This way you can:
1. Use a different CI provider
2. It can get published on PyPI on its own  schedule whenever maintenance releases are required

Once it's published on PyPI, we'll link to it on the official documentation and README. What do you think?

trapito:
Update : Most of the test cases are enabled for the windows worker (7 have to be skipped) and everything works fine, there's a file for appveyor (appveyor.yml) that does the same as travis.yml (minus 3.6-dev)

The test cases have been slightly modified so temporary file paths aren't hardcoded and the worker class is chosen based on the host operating system.

There's an error in a pypy heroku test but my patches are unrelated to that.

I'll write some docs and samples for the windows specific features, it would be nice if we wrote proper documentation in a readthedocs page.

selwin:
@trapito excellent! Do you want to prepare the github repository for this worker too? We can publish it as `rq-windows-worker` on PyPI or something similar. Once published, we can document and put a link to it here.

trapito:
Honestly i don't see a reason to put this in a different repository the code footprint is small and the same the same tests apply for both workers, also both appveyor and travis can coexist so i don't see the point in it.

selwin:
Sorry about my reluctance in merging in this PR. Yes, I do have a reason :)

If I merge this PR in, I'll have to maintain this code for the long haul, which I'm not comfortable with. I'm not comfortable pulling in code that I can't maintain. I have also never done any development work on Windows before so if something is broken someday, I won't be able to fix it.

trapito:
Gotcha, it kind of sucks but i guess it's understandable, i don't know if publishing it as a stand alone module is feasible, however, if it were then the Worker class constructor modification (to receive the cli_options dict) would be needed to be able to properly build a command line if it were not available (realistically only needed for pytest).

selwin:
I think the module should include its own command, similar to `rq-gevent-worker` https://github.com/zhangliyong/rq-gevent-worker .

However, I'm open to making changes to how `Worker` is instantiated and the CLI to accommodate third party worker classes. Would passing through all cli options as `kwargs` into `Worker.__init__` be sufficient?

selwin:

PR #863, opened 2017-08-08
Update docs/results.md

ushuz:
- ~~Add an example of setting result TTL with `enqueue()`~~
- Typo
- Encourage the use of `enqueue()`

### DIFF
--- a/docs/results.md
+++ b/docs/results.md
@@ -95,11 +95,11 @@ job timed out.
 
 If a job requires more (or less) time to complete, the default timeout period
 can be loosened (or tightened), by specifying it as a keyword argument to the
-`Queue.enqueue()` call, like so:
+`enqueue()` call, like so:
 
 {% highlight python %}
 q = Queue()
-q.enqueue(func=mytask, args=(foo,), kwargs={'bar': qux}, timeout=600)  # 10 mins
+q.enqueue(mytask, args=(foo,), kwargs={'bar': qux}, timeout=600)  # 10 mins
 {% endhighlight %}
 
 You can also change the default timeout for jobs that are enqueued via specific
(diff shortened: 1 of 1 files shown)


docs/results.md
```diff
@@ -91,11 +95,11 @@ job timed out.
 
 If a job requires more (or less) time to complete, the default timeout period
 can be loosened (or tightened), by specifying it as a keyword argument to the
-`Queue.enqueue()` call, like so:
+`enqueue()` call, like so:
 
 {% highlight python %}
 q = Queue()
-q.enqueue(func=mytask, args=(foo,), kwargs={'bar': qux}, timeout=600)  # 10 mins
```
ushuz: `enqueue()` doesn't have `func` keyword argument

docs/results.md
```diff
@@ -108,7 +112,7 @@ high = Queue('high', default_timeout=8)  # 8 secs
 low = Queue('low', default_timeout=600)  # 10 mins
 
 # Individual jobs can still override these defaults
-low.enqueue_call(really_really_slow, timeout=3600)  # 1 hr
```
ushuz: I think here we should use `enqueue()` just like the above snippet as they are in the same section. Besides, `enqueue_call()` usually looks like `enqueue_call(func= really_really_slow)`.

selwin: Yes, agreed. Could you also replace references to `enqueue_call` with `enqueue`? We really want to encourage people to use `enqueue`

ushuz: @selwin I thought the same. I pushed a new commit replacing `enqueue_call()` with `enqueue()`. PTAL.

selwin:
Thanks!

PR #871, opened 2017-08-31
Issue 870 Improved coverage for connections.py and utils.py

theodesp:
Fixes https://github.com/nvie/rq/issues/870


selwin:
Thanks!

Issue #721, opened 2016-06-29
change datetime formatting to include milliseconds


samuelcolvin:
I propose `'%Y-%m-%dT%H:%M:%SZ'` in `utcformat` becomes `'%Y-%m-%dT%H:%M:%S.%fZ'`, note the `%f` to give more exact times.

This is because we have numerous jobs with take around a second and are queued for around a second and without `started_at` and `ended_at` being saved with ms it's impossible to get a good idea of how long their really taking.

I can see from the comments in `utcparse` that something similar was previously the case so I wanted to canvas opinions on whether this was a possibility before submitting a PR?

samuelcolvin:
see [code here](https://github.com/nvie/rq/blob/master/rq/utils.py#L158-L167)

selwin:
Yeah, I think this is a good idea. PR welcome :)

theodesp:
@samuelcolvin are you willing to make a PR for this?

samuelcolvin:
Hi @theodesp, thanks for the prompt, I'd forgotten about this.

I'm submit a PR once #874 is merged.

selwin:
@samuelcolvin #874 has been merged 😊. I personally want to see this supported in RQ because this may be useful in collecting worker/job metrics - something I’ll be working on soon.

PR #874, opened 2017-09-05
remove python 2.6 support

samuelcolvin:
as per discussion #715 this removes python 2.6 support.

There are probably (definitely) lots of improvements possible now the syntax doesn't have to be stretched to support such a massive range of python versions, but this is a start.

**Rationale**: I was thinking about submitting a PR for #721 as someone suggested and realised that the reason I was unkeen was having to remember how an outdated version of python worked and maybe hack around to get python 2.6 installed to test the code. Removing this should make contributing to rq much easier for everyone.

Anyone still running 2.6 has much bigger problems than upgrading rq: they need to concentrate on upgrading to 2.7 and then 3 or scrapping their code-base and starting again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ language: python
 services:
   - redis
 python:
-  - "2.6"
   - "2.7"
   - "3.3"
   - "3.4"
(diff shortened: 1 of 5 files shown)


samuelcolvin:
Personally I think we should drop python 2 completely, but this is a start.

theodesp:
Fine by me

selwin:
Thanks!

PR #877, opened 2017-09-08
Fixed #809 - Added tests for various cli config parameters

### LINKED ISSUE #809: rq worker does not honor REDIS_PASSWORD setting
I have a settings module to configure the worker, and I'm using -c to pass this module to 'rq worker'.

All settings work fine, including the REDIS_URL and QUEUES settings. The REDIS_PASSWORD seems to be ignored though, as the worker fails with 

`redis.exceptions.ResponseError: NOAUTH Authentication required.`

I have listened for incoming connections on my redis server with 'MONITOR', and can confirm that rq never sends the AUTH command, even with the password configured. 

$ docker exec -ti {redacted} pip3 list | grep rq
rq (0.7.1)

theodesp:
Tests show that the config parameters from a file do get passed in the Redis connection pool

### DIFF
--- a/tests/dummy_settings_override.py
+++ b/tests/dummy_settings_override.py
@@ -1,0 +1,8 @@
+# -*- coding: utf-8 -*-
+from __future__ import (absolute_import, division, print_function,
+                        unicode_literals)
+
+REDIS_HOST = "testhost.example.com"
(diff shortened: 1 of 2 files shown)


selwin:
Thanks!

PR #878, opened 2017-09-08
Fixed #731 - Support for deleting Queues

### LINKED ISSUE #731: Support for deleting queues
I couldn't find a way to delete a queue. There's `Queue.empty`, but that leaves an empty queue. I would like to remove the queue completely.

Currently I'm using the following code:

``` python
queue.empty()
redis_conn.srem(rq.Queue.redis_queues_keys, queue._key)
redis_conn.delete(queue._key)
```

That seems to work, but I'd rather not rely on RQ-internals (e.g. `Queue._key`) in my code. It would therefore be nice if that code was available as `Queue.delete`, similar to `Job.delete`.

I can prepare a PR, let me know if you're interested.

theodesp:
Cool

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -111,6 +111,12 @@ class Queue(object):
         script = self.connection.register_script(script)
         return script(keys=[self.key])
 
+    def delete(self):
+        """Deletes the queue and all the associated messages on the queue."""
+        self.empty()
+        self.connection.srem(self.redis_queues_keys, self._key)
+        self.connection.delete(self._key)
+
     def is_empty(self):
         """Returns whether the current queue is empty."""
         return self.count == 0
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -111,11 +111,15 @@ class Queue(object):
         script = self.connection.register_script(script)
         return script(keys=[self.key])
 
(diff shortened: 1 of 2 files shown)


rq/queue.py
```diff
@@ -111,6 +111,12 @@ def empty(self):
         script = self.connection.register_script(script)
         return script(keys=[self.key])
 
+    def delete(self):
+        """Deletes the queue and all the associated messages on the queue."""
+        self.empty()
+        self.connection.srem(self.redis_queues_keys, self._key)
```
selwin: We should pipeline these two calls.

rq/queue.py
```diff
@@ -111,6 +111,12 @@ def empty(self):
         script = self.connection.register_script(script)
         return script(keys=[self.key])
 
+    def delete(self):
```
selwin: Can we add a boolean `delete_jobs` parameter? `queue.empty()` should only be called with `delete_jobs=True`.

theodesp: OK do you want it to have a default parameter value of True or False?

theodesp:
Updated!

selwin:
Thanks!

PR #880, opened 2017-09-09
add milliseconds into timestamps

### LINKED ISSUE #721: change datetime formatting to include milliseconds

samuelcolvin:
fix #721

### DIFF
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -157,16 +157,19 @@ def utcnow():
     return datetime.datetime.utcnow()
 
 
+_TIMESTAMP_FORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
+
+
 def utcformat(dt):
-    return dt.strftime(as_text('%Y-%m-%dT%H:%M:%SZ'))
+    return dt.strftime(as_text(_TIMESTAMP_FORMAT))
 
 
 def utcparse(string):
     try:
-        return datetime.datetime.strptime(string, '%Y-%m-%dT%H:%M:%SZ')
+        return datetime.datetime.strptime(string, _TIMESTAMP_FORMAT)
     except ValueError:
-        # This catches RQ < 0.4 datetime format
-        return datetime.datetime.strptime(string, '%Y-%m-%dT%H:%M:%S.%f+00:00')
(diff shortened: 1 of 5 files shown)


selwin:
@samuelcolvin what are your thoughts about using unix timestamp instead of the current human readable format? It might be slightly more efficient.

samuelcolvin:
I thought about it and indeed the async successor to rq: [arq](https://github.com/samuelcolvin/arq) uses unix timestamps, but there are major disadvantages to unix time stamps and only really one big advantage.

The **big advantage** is that you can store/parse/transfer an int which should be more efficient than a string, but only if the encoding has native support for int packing. rq uses json which encodes ints as strings, thus storing unix time stamps isn't significantly more efficient (either in terms of performance or space) than the current timestamps. arq uses msgpack to encode data in redis which can take advantage of efficient int packing so storing an int is much more efficient than a datetime string.

The **big disadvantages** of unix time stamps are:
* they're bloody hard to make any sense of as humans. For example I give you `1505836763`, make any sense to you? me neither
* There's massive ambiguity in whether unix timestamps are in seconds or ms, the two approaches are used roughly 50-50. This application would require ms, but that's even hard to read, even relatively for a human, for example: could a user realistically have filled in their profile between `1505836763321` and `1505837369987`? not so easy to read.
* in this situation the best approach might be to use seconds since 1970 as a float, but that creates a new set of problems:
  * it's not a standard format at all
  * you have to be careful when formatting floats that you don't get 8 decimal places and end up with massive strings

In summary, I've agonised about unix timestamps vs formatted datetimes in the past a lot, hence this rant. In this situation I don't think it's a good idea.

PR #885, opened 2017-09-30
add gevent worker

jiajunhuang:


selwin:
It would be interesting to have a worker with concurrency support. A couple of questions:
1. Have you run this in production?
2. Could you please add basic test cases for this?

jiajunhuang:
Oh, sorry for that. I've running this in production, I'll add unit-test soon.

jiajunhuang:
I'm sorry that I would like to close this pr.

the way I'm using rq with gevent is I wrote the entrypoint myself, like this:

```python
from gevent import monkey
monkey.patch_all()  # noqa

import multiprocessing
import logging

import gevent
from raven import Client
from raven.transport.http import HTTPTransport
from redis import StrictRedis
from rq import Worker
from rq.worker import WorkerStatus
from rq.contrib.sentry import register_sentry

from itachi.config import config


class GeventWorker(Worker):
    def execute_job(self, job, queue):
        self.set_state(WorkerStatus.BUSY)
        self.log.debug("gonna spawn a greenlet to execute the given job.")
        gevent.spawn(self.perform_job, job, queue).join()
        self.log.debug("greenlet executed.")
        self.set_state(WorkerStatus.IDLE)


def gevent_worker(queues):
    client = Client(config.SENTRY_DSN, transport=HTTPTransport)
    worker = GeventWorker(
        queues=queues,
        connection=StrictRedis.from_url(config.WORKER_BROKER)
    )
    register_sentry(client, worker)
    worker.work()


def start_worker():
    queues = ["to_user", "device", "to_project"]
    workers = []

    for i in queues:
        p = multiprocessing.Process(target=gevent_worker, args=([i],))
        p.start()

    for i in workers:
        i.join()


def setup_root_logger():
    root_logger = logging.getLogger()
    root_logger.setLevel(logging.INFO)


if __name__ == "__main__":
    setup_root_logger()
    start_worker()
```

I think this can help man who wanna use gevent with rq.

cw1427:
Hi @jiajunhuang :  Does this GeventWorker works well in prod env?  I worried about the origin worker would fork a horse process to perform the job and leave the main work process doing the monitor_work_horse.   But when use your Geventwork which just spawn a greenlet for job perform, there is no other process take the worker monitor?  Isn't it a risk or problem?

lowercase00:
@jiajunhuang mind sharing more details on why did you decide to close this? How's been the performance/stability in production environment?

jiajunhuang:
@lowercase00 @cw1427 It is running in production environment, I close this just because I've found the solution. I've not monitor the sub coroutine, you can contribute to it if you are interest on it 😄

lowercase00:
@jiajunhuang any specific reason this couldn't be an "official" gevent worker?

selwin:
I’d be happy to include this in the official RQ package if someone wants to make a PR and maintain this.

lowercase00:
I'm not sure I understand this correctly, but from what I've seen it seems that there are two main issues of adding this the way it is right now:
- The patch has to be made at the beginning of the file, and if gevent is not being used this could lead to issues depending on what the workers are executing.
- As @cw1427 noted, I'm not sure I understood, but this could lead to zumbi workers? Something about this has been mentioned in [this library](https://github.com/zhangliyong/rq-gevent-worker) (it doesn't work anymore).

I imagine there are two possible approaches.

Split workers into files (create a new `workers` folder, and have a folder for each type of worker), and then having the gevent patching being done on the top level of the GeventWorker, and only import the module if GeventWorker class is being used.

Having a top level flag to the worker cli to enable gevent, and then only import gevent if the flag is set.

I honestly don't have the skills to tackle this at the moment as much as I would like to, but I'm willing to try to help if someone leads the design decisions. 

In the mean time, it can be an option to add this example to the documentation on how to run RQ with gevent concurrency. There's already a mention on how to run custom classes, but I imagine it could be helpful to give a small (untested) example on how this could work. In that case, it seems that @jiajunhuang GeventWorker class could be a good resource.

jiajunhuang:
The reason I use a new entrypoint is if you want to use gevent, you have to import and patch it on the top, otherwise modules can't be replaced with gevent's implementation(because of Python will cache modules that already import). So it will not work if someone just simply split worker into a module, once you import worker from `__init__.py`, packages used by `worker.Worker` can't be patched anymore.

lowercase00:
I guess if we added:

```python
if "Gevent" in sys.argv:
    try:
        import gevent.monkey
        gevent.monkey.patch_all()
    except ImportError as error:
        print('GeventWorker requires Gevent to be installed.')
        sys.exit(1)
```

to the top of the `cli.py` file, it would solve the import issue (this is pretty much how Celery does it).
The next challenge would be whether to use `os.fork()` or:

```
p = multiprocessing.Process(target=gevent_worker, args=([i],))
p.start()
```

Maybe this could be achieved my checking which Worker class in being used in the `fork_work_horse()` method?

lowercase00:
I made a first test using the code from [this repo](https://github.com/zhangliyong/rq-gevent-worker) (all credits to @zhangliyong ).
Here's the branch: https://github.com/lowercase00/rq/tree/gevent

PR #752, opened 2016-09-22
 Add at_front option for decorated functions

Pavel-Guseynov:

PR #888, opened 2017-10-10
Add at_front option for decorated functions

karech:
There is already a Pull request for this option, but it hanged up a year ago (#752).


selwin:
Thanks!

PR #892, opened 2017-10-20
Allow passing in meta to decorated jobs

jlucas91:


jlucas91:
Hi @selwin ! I'd love to do what I can to get this merged in. What's the best way for me to do so? Happy to take another approach if you'd prefer. 

Pinging you since I see you on a lot of merges, but LMK if there's someone else I should talk to.

rq/decorators.py
```diff
@@ -44,8 +44,9 @@ def delay(*args, **kwargs):
             else:
                 queue = self.queue
             depends_on = kwargs.pop('depends_on', None)
+            meta = kwargs.pop('meta', None)
```
selwin: Do you mind adding this argument to `__init__`? I just merged in another PR that added another `at_front` argument and realized that this approach could get messy real quick.

If you could merge master into your branch and move `depends_on` kwarg into `__init__` as well, it would be great!

jlucas91: @selwin Sure! Can do. I was a little leery of the approach as well.

Happy to move `depends_on` as well, though that looks like it would be a breaking change?

jlucas91:
@selwin Updated. LMK what you think. 

Is the breaking change here on `depends_on` something we need to be concerned about?

selwin:
With regards to your comment about `depends_on` kwarg, since `depends_on` already made it into a release version of RQ, do you mind overriding it like this?
```python
depends_on = kwargs.pop('depends_on', None)
if not depends_on: # Fall back to option provided by decorator
    depends_on = self.depends_on
```

Please also move `at_front` out of `kwargs` and into `__init__`. We don't need to worry about backwards compatibility for this kwarg since it never made it into a release version.

It would be great if you can make this change by Monday. I plan on releasing 0.9.0 on Tuesday.

jlucas91:
Hey @selwin - Sorry to drop off the map. I'll take a look at this tomorrow AM. Thanks for your help here!

jlucas91:
@selwin Should be good to go. LMK what you think.

jlucas91:
@selwin Bump on this - Any chance we could merge?

selwin:
Thanks!

PR #895, opened 2017-10-23
try latin-1 if utf-8 doesn't work, when decoding exception strings

ndparker:
Hi!

Recently we've had the issue, that an exception contained binary data in its message, which could not be decoded as utf-8. This PR uses latin-1 to decode the exception strings if utf-8 decoding fails.

Thanks for considering!


selwin:
Thanks!

PR #901, opened 2017-11-07
Fixed an issue where `birth` not present in Redis

vanife:
Fixed an issue where worker.refresh() may fail if `birth` is not present in Redis.


selwin:
Thanks!

PR #902, opened 2017-11-09
improve logging in worker.py

samuelcolvin:
Worker logging is currently pretty verbose and naively configured. 

This improves it.

Changes:
* logging `*** Listening on ...` before/after every job moved from INFO to DEBUG
* logging `*** Listening on ...` at INFO on initial load.
* `log_result_lifespan` to prevent, `Result is kept for 500 seconds` being logged for every job
* whoever initially implemented logging in `worker` apparently didn't take the time to look at how python logging works - it specifically allows `log.info('template with var %s', var)` syntax to avoid formatting log statements which are subsequently dropped. I've fixed this on all `DEBUG` log statements, as that will have the most effect since they're generally dropped.
* there was a bunch of white space at the end of lines (is rq not running linting !?) which pycharm automatically deleted.

Currently executing a job with rq causes 4 statements at INFO level. This is currently causing about 1GB of logs each month from rq alone.

I can add a test for `log_result_lifespan` if you're happy with this.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -94,6 +94,7 @@ class Worker(object):
     death_penalty_class = UnixSignalDeathPenalty
     queue_class = Queue
     job_class = Job
+    log_result_lifespan = True
 
     @classmethod
     def all(cls, connection=None, job_class=None, queue_class=None):
@@ -132,7 +133,7 @@ class Worker(object):
                      connection=connection,
                      job_class=job_class,
                      queue_class=queue_class)
-        
+
         worker.refresh()
 
         return worker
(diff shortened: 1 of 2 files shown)


selwin:
Yes please :). Please document `log_result_lifespan` too if possible.

samuelcolvin:
`log_result_lifespan` tests done.

@selwin I hope this is ok, I can't find the code for the docs site so can't add proper documentation, but I've added a comment explaining `log_result_lifespan`. Hopefully that's ok for now.

selwin:
Thanks!

samuelcolvin:
Thanks for merging. Would be great to get this released fairly soon.

It's only a logging improvement and there shouldn't be any braking changes so a minor release should suffice?

selwin:
@samuelcolvin will try to release version 0.10 sometime in the next two weeks. Besides this one, I've merged in my job compression branch (https://github.com/nvie/rq/pull/907) for better memory usage.

I've got patch coming that should fix problems like https://github.com/ui/django-rq/issues/231 when running RQ with hundreds of workers.

PR #906, opened 2017-11-20
Refactor worker.refresh() attribute parsing

nicknovitski:


selwin:
Could you please move the methods into two separate functions `parse_int` and `parse_datetime` so we can use these in other places? Please put these functions in `utils.py`. Thanks :)

nicknovitski:
Sure.  Actually there was a third case, of `float`, so I just made one function for all three.  I'll add tests for it in a bit.

selwin:
@nicknovitski sorry for being nit-picky, but I'd rather have specialized functions such as `parse_date`. It just feels much cleaner to do:
`parse_date(obj.get('enqueued_at'))`
than
`parse_if_present(obj.get('enqueued_at'), parser=utcparse)`

Could you please make the modification?

PR #912, opened 2017-12-21
Add description param to  @job decorator

nlyubchich:
After using `@job` decorator for a function that takes a long string, in RQ worker I got printed all the args/kwargs via `job.get_call_string()`.

To get this overridden, I added `description` argument to the decorator.

I decided not to put this in `delay` method because it's may be currently be used by end user.


selwin:
Thanks!

PR #427, opened 2014-09-21
remove dependents jobs on parent job delete

psihonavt:

PR #916, opened 2018-01-10
Delete dependents of job explicitely

olingerc:
Linked to PRs https://github.com/rq/rq/pull/427 and https://github.com/rq/rq/pull/856
As discussed in both of the above PRs I am introducing a method to explicitly delete dependents of a job.
In this implementation deleting a job by default does not touch the dependents (should we introduce a CANCELED or ORPHANED state?).
In case a user uses `delete_dependents=True`, all dependent jobs are deleted.
Since canceling a job only removes it form the queue and dependents are not in a queue yet anyway that method is not involved.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -582,7 +582,6 @@ class Job(object):
 
     def delete_dependents(self, pipeline=None, remove_from_queue=True):
         """Delete jobs depending on this job."""
-        connection = pipeline if pipeline is not None else self.connection
         for dependent_id in self.dependents_ids:
             try:
                 dependent = Job.fetch(dependent_id, connection=self.connection)
--- a/rq/job.py
+++ b/rq/job.py
@@ -585,8 +585,7 @@ class Job(object):
 
     def delete_dependents(self, pipeline=None, remove_from_queue=True):
         """Delete jobs depending on this job."""
-        # TODO: should we do connection.delete(self.dependents_key) in case
-        # this method is called on its own?
+        connection = pipeline if pipeline is not None else self.connection
         for dependent_id in self.dependent_ids:
             try:
                 job = Job.fetch(dependent_id, connection=self.connection)
@@ -595,6 +594,7 @@ class Job(object):
             except NoSuchJobError:
                 # It could be that the dependent job was never saved to redis
                 pass
+        connection.delete(self.dependents_key)
 
     # Job execution
     def perform(self):  # noqa
--- a/rq/job.py
+++ b/rq/job.py
@@ -579,6 +579,7 @@ class Job(object):
                                    remove_from_queue=remove_from_queue)
 
         connection.delete(self.key)
+        connection.delete(self.dependents_key)
 
     def delete_dependents(self, pipeline=None, remove_from_queue=True):
         """Delete jobs depending on this job."""
@@ -591,8 +592,6 @@ class Job(object):
             except NoSuchJobError:
                 # It could be that the dependent job was never saved to redis
                 pass
-
-        connection.delete(self.dependents_key)
 
     # Job execution
     def perform(self):  # noqa
--- a/rq/job.py
+++ b/rq/job.py
@@ -321,6 +321,7 @@ class Job(object):
         self.ttl = None
         self._status = None
         self._dependency_id = None
+        self._dependents_ids = []
         self.meta = {}
 
     def __repr__(self):  # noqa  # pragma: no cover
@@ -444,6 +445,7 @@ class Job(object):
         self.result_ttl = int(obj.get('result_ttl')) if obj.get('result_ttl') else None  # noqa
         self._status = as_text(obj.get('status') if obj.get('status') else None)
         self._dependency_id = as_text(obj.get('dependency_id', None))
+        self._dependents_ids = self.dependents_ids
         self.ttl = int(obj.get('ttl')) if obj.get('ttl') else None
         self.meta = unpickle(obj.get('meta')) if obj.get('meta') else {}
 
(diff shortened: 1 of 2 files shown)


rq/job.py
```diff
@@ -194,6 +194,12 @@ def dependency(self):
         self._dependency = job
         return job
 
+    @property
+    def dependents_ids(self):
```
selwin: Could you please rename this to `dependent_ids`?

rq/job.py
```diff
@@ -486,6 +494,8 @@ def to_dict(self, include_meta=True):
             obj['status'] = self._status
         if self._dependency_id is not None:
             obj['dependency_id'] = self._dependency_id
+        if self._dependency_id is not None:
```
selwin: Did you mean `dependent_ids`?

rq/job.py
```diff
@@ -438,6 +445,7 @@ def to_date(date_str):
         self.result_ttl = int(obj.get('result_ttl')) if obj.get('result_ttl') else None  # noqa
         self._status = as_text(obj.get('status') if obj.get('status') else None)
         self._dependency_id = as_text(obj.get('dependency_id', None))
+        self._dependents_ids = self.dependents_ids
```
selwin: We shouldn't store `dependent_ids` in Redis? This value won't be accurate since new dependents could be added.

rq/job.py
```diff
@@ -561,9 +577,26 @@ def delete(self, pipeline=None, remove_from_queue=True):
                                             job_class=self.__class__)
             failed_queue.remove(self, pipeline=pipeline)
 
+        if delete_dependents is True:
```
selwin: `if delete_dependents:` is more Pythonic :)

rq/job.py
```diff
@@ -561,9 +577,26 @@ def delete(self, pipeline=None, remove_from_queue=True):
...
         connection.delete(self.key)
         connection.delete(self.dependents_key)
 
+    def delete_dependents(self, pipeline=None, remove_from_queue=True):
+        """Delete jobs depending on this job."""
+        # TODO: should we do connection.delete(self.dependents_key) in case
+        # this method is called on its own?
+        for dependent_id in self.dependents_ids:
+            try:
+                dependent = Job.fetch(dependent_id, connection=self.connection)
```
selwin: Nitpick, could you please rename `dependent` to `job`?

rq/job.py
```diff
@@ -494,8 +492,9 @@ def to_dict(self, include_meta=True):
             obj['status'] = self._status
         if self._dependency_id is not None:
             obj['dependency_id'] = self._dependency_id
-        if self._dependency_id is not None:
-            obj['dependents_ids'] = self._dependents_ids
+        _dependents_ids = list(self.dependent_ids)
```
selwin: Conversion of `dependent_ids` to list should be moved to `def dependent_ids`. The docstring on `dependent_ids` also mentions that it returns `a list of ids`.

rq/job.py
```diff
@@ -494,8 +492,9 @@ def to_dict(self, include_meta=True):
...
+        if len(_dependents_ids) > 0:
```
selwin: `if _dependent_ids` is more Pythonic :)

rq/job.py
```diff
@@ -513,25 +522,31 @@ def save_meta(self):
...
-    def cancel(self):
+    def cancel(self, pipeline=None):
         """Cancels the given job, which will prevent the job from ever being
         ran (or inspected).
 
         This method merely exists as a high-level API call to cancel jobs
         without worrying about the internals required to implement job
         cancellation.
         """
+        # TODO: Dependent jobs are by default in a DEFERRED state. Should we
```
olingerc: @selwin What do you think about the CANCELED state?

selwin: Yes, introducing a canceled state and a corresponding `CanceledJobRegistry` is a good idea. This way, we can keep track of canceled jobs.

This should be implemented in a separate PR though. Feel free to open an issue for this so we don't forget :)

selwin:
@olingerc TravisCI haven't been running builds since this repo was moved. I just updated the integration, do you mind making a change so we can check whether the builds pass on Travis?

rq/job.py
```diff
@@ -486,6 +492,9 @@ def to_dict(self, include_meta=True):
             obj['status'] = self._status
         if self._dependency_id is not None:
             obj['dependency_id'] = self._dependency_id
+        _dependents_ids = self.dependent_ids
```
selwin: `dependent_ids` are stored as a set in `job.dependent_keys` so these don't need to be serialized for storage. I think we can remove these.

selwin:
@olingerc could you please merge the current master branch to your pull request so codecov doesn't complain?

rq/job.py
```diff
@@ -561,9 +570,26 @@ def delete(self, pipeline=None, remove_from_queue=True):
...
             failed_queue.remove(self, pipeline=pipeline)
 
+        if delete_dependents:
+            self.delete_dependents(pipeline=pipeline,
+                                   remove_from_queue=remove_from_queue)
+
         connection.delete(self.key)
         connection.delete(self.dependents_key)
 
+    def delete_dependents(self, pipeline=None, remove_from_queue=True):
```
selwin: One more detail, we don't need the `remove_from_queue` argument because dependent jobs are not yet enqueued.

rq/job.py
```diff
@@ -561,9 +570,26 @@ def delete(self, pipeline=None, remove_from_queue=True):
...
+        """Delete jobs depending on this job."""
+        connection = pipeline if pipeline is not None else self.connection
+        for dependent_id in self.dependent_ids:
+            try:
+                job = Job.fetch(dependent_id, connection=self.connection)
+                job.delete(pipeline=pipeline,
+                           remove_from_queue=remove_from_queue)
```
selwin: Please use `remove_from_queue=False` for this one so RQ doesn't try to remove the job from Redis (we save an extra Redis operation here).

selwin:
Thanks!

PR #923, opened 2018-01-18
Added pytest to dev-dependencies

eshigee:
PyTest is required for running tests, but was not present in dev-requirements.txt


selwin:
Thanks!

PR #929, opened 2018-02-04
update Redis requirement to one specified by readme

rooterkyberian:
Changed Redis version requirement to be the same as one specified by Readme.
https://github.com/rooterkyberian/rq/commit/c6f79784640f95635b5e575e6691caa47350e948


selwin:
Thanks!

PR #939, opened 2018-03-22
Define redis key prefix as class variable

### LINKED ISSUE #920: Best practice to set redis key prefix for all keys used by RQ
Hi,

After quick scan of the source code I determined that while base classes like Queue, Worker etc could be subclassed in order to override `redis_queue_namespace_prefix`, there is no way to control this code (except full method override which is strongly undesired):

```python
def empty(self):
    """Removes all messages on the queue."""
    script = b"""
        local prefix = "rq:job:"
        local q = KEYS[1]
        local count = 0
        while true do
            local job_id = redis.call("lpop", q)
            if job_id == false then
                break
            end

            -- Delete the relevant keys
            redis.call("del", prefix..job_id)
            redis.call("del", prefix..job_id..":dependents")
            count = count + 1
        end
        return count
    """
    script = self.connection.register_script(script)
    return script(keys=[self.key])
```

where prefix is just hardcoded.

Having this how subj can be achieved in best possible way? Thanks.

Looking ahead it would be very helpful to have this as some sort of a setting that will shine with `django-rq` when `KEY_PREFIX` may be used in `CACHES` setting definition.

stj:
Some prefixes were hardcoded in several places. This made it hard to
use custom prefixes via subclasses.

Resolves #920

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -366,7 +366,7 @@ class Job(object):
     @classmethod
     def dependents_key_for(cls, job_id):
         """The Redis key that is used to store job dependents hash under."""
-        return '{0}{0}:dependents'.format(cls.redis_job_namespace_prefix, job_id)
+        return '{0}{1}:dependents'.format(cls.redis_job_namespace_prefix, job_id)
 
     @property
     def key(self):
--- a/rq/job.py
+++ b/rq/job.py
@@ -85,6 +85,7 @@ def get_current_job(connection=None, job_class=None):
 class Job(object):
     """A Job is just a convenient datastructure to pass around job (meta) data.
     """
+    redis_job_namespace_prefix = 'rq:job:'
 
     # Job construction
     @classmethod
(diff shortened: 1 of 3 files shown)


rq/job.py
```diff
@@ -360,12 +361,12 @@ def set_id(self, value):
...
     def key_for(cls, job_id):
         """The Redis key that is used to store job hash under."""
-        return b'rq:job:' + job_id.encode('utf-8')
+        return (cls.redis_job_namespace_prefix + job_id).encode('utf-8')
 
     @classmethod
     def dependents_key_for(cls, job_id):
         """The Redis key that is used to store job dependents hash under."""
-        return 'rq:job:{0}:dependents'.format(job_id)
+        return '{0}{0}:dependents'.format(cls.redis_job_namespace_prefix, job_id)
```
selwin: Could you please fix this typo? Should be `{0}{1}`

stj: Will do. Saw that I did same mistake for `COMPACT_QUEUE`.

Do you know why `key_for` returns a binary while `dependents_key_for` returns string?

selwin:
Thanks!

Issue #915, opened 2018-01-09
worker dies after a (long) inactivity


ELE3E:
Hi I have a setup that has a number of workers listening for on demand jobs. After a while of inactivity, i.e. a list of messages like the one below, the heartbeat signals stop appearing and the worker is not available any more in list of workers (I have an RQ dashboard setup as well).
The time after it stops is not consistent, nor is the number of heartbeats sent before dying.
The process of the worker that died is still there.
How can I check what the root cause of the problem is?

> Sent heartbeat to prevent worker timeout. Next one should arrive within 420 seconds.

selwin:
Is the rqworker process still running? Is there any failed job in the queue that causes the worker to crash?

You should also use a process manager like systemd or supervisor to run RQ in production.

ELE3E:
Yes the rq worker process is still running. It was not executing a job at that time, a long list of heartbeat signals before it stopped.
We are running it on a docker system, as soon as the process exits it will restart automatically.

selwin:
Do you have a job with infinite timeout?

ELE3E:
No I do not, always take the default timeout, but the jobs are usually done in a few seconds

samuelcolvin:
Are there any changes in the memory usage of that container?

Can you see network traffic for that container? If so what is happening

ELE3E:
No network traffic, constant memory usage.

On 10 Jan 2018, at 16:08, Samuel Colvin <notifications@github.com<mailto:notifications@github.com>> wrote:


Closed #915<https://github.com/rq/rq/issues/915>.

samuelcolvin:
You might want to enable `debug` logging to see if it shows any more details (if you haven't already).

Also, could this be the worker stopping [here](https://github.com/rq/rq/blob/afbbe47dbbf776a7729b4e2704363484e0eb6c93/rq/worker.py#L492) perhaps due to hitting `default_worker_ttl` or a problem getting a job from the queue?

ELE3E:
Thanks for the tips. I will put the logging level to debug and also enable the verbose option. Unfortunely, I will have to wait until the worker halts again.
From my logs, I don't see any job around the time the worker halted, so I don't think the problem is getting a job from the queue. 
What is this default_worker_ttl for?

vithickvi:
I am still facing this issue. Have 5 worker servers, if there is no job for a couple of hours then automatically the worker server disappears from rq info. Every time i have to go and restart the supervisor in each worker servers

[...]

PR #945, opened 2018-04-05
add periodic worker heartbeats

### LINKED ISSUE #824: Adding periodic timeout (or heartbeat) to jobs
I had a go at implementing #805, but after some investigation, it looks like that doesn't solve the problem I was aiming to fix.

Main aim is to allow for very long running jobs which do work in smaller chunks. Ideally, I'd like to have an overall job timeout (which might be something like 24 hours), and work chunk timeout (which might be something like 30 seconds).

This would allow for much quicker notification of any issues (e.g. worker dying, job getting stuck, etc.), without having to wait the full 24 hours.

Currently, when a job timeout is set, this results in the worker TTL being increased to at least this value, and also sets up `UnixSignalDeathPenalty` to send a `signal.alarm` if the timeout is reached.

To add the functionality I'd want, I think the following changes would be needed, just wanted to get some feedback before coding them and making a PR:

1. Modify `UnixSignalDeathPenalty` to take both `timeout` and `periodic_timeout` arguments (both optional)
2. Add `periodic_timeout` argument to `Job`
3. Set up the `UnixSignalDeathPenalty` from within `Job` rather than `Worker` when a job is started (means that jobs can cancel/update this, and workers can still access this from their reference to a job)
4. If `timeout` is set, and `periodic_timeout` is not set, everything behaves as it does now
5. If `periodic_timeout` is set, `UnixSignalDeathPenalty` uses `signal.alarm(self.periodic_timeout)`
6. Add a `heartbeat()` method to Job, which can be called by running jobs to indicate the everything is still progressing, and work is being done. This method would a) ensure that `worker.heartbeat()` is called, and b) the current UnixSignalDeathPenalty alarm is cancelled, and a new one set

Any feedback/suggestions would be welcome - there are a fair few changes here, and if they're not suitable, happy to rethink!

### LINKED ISSUE #944: Workers dies unexpectedly - job not marked as failed
Hi,

I'm having issues in my infrastructure that sometimes workers (i.e., their host machine) just randomly die or become permanently unavailable on the network). If the worker is currently processing a job, this causes the job to hang indefinitely.

This is different from the case when the work horse dies, in which case the job is correctly put into the failed queue, so I can manually requeue it later.

Workers already send a heartbeat, but apparently this only works in IDLE state, and not in BUSY.
Right now I have about 100 workers in my list that are BUSY - but in reality they don't exist any more.

Is it possible to restore those hogged jobs from the dead workers?
Is it possible to add a more reasonable heartbeat to the workers themselves (not work horses)?

Are there any strategies to reliably recover from my current predicament?

Kriechi:
fixes #944
fixes #824

Might also affect #915


Kriechi:
@selwin Did you have something like this in mind?

selwin:
Not quite. This implementation polls the OS whether a job has finished every second so workers won't detect finished jobs in realtime. This will significantly lower worker efficiency for use cases where many short duration jobs need to be processed.

It should still use `os.waitpid(self._horse_pid, 0)` so it can detect jobs finishing in realtime, but uses signals to interrupt `waitpid()` (which hangs for an indefinite period of time) to periodically check whether job is still alive.

Job timeouts use a similar mechanism. It sets up an alarm right before job is executed:

https://github.com/rq/rq/blob/master/rq/worker.py#L788

This is how the alarm is setup
https://github.com/rq/rq/blob/master/rq/timeouts.py#L53

Kriechi:
I updated the PR with a new implementation using an alarm, as you suggested.
If you think we are on the right track, I can try to write a test for it.

selwin:
Yes, this PR is on the right track. A few comments:
* We should create a new exception instead of changing `JobTimeoutException`
* `ChildProcessError` needs to be handled (in case jobs finish during heartbeat)

This is a critical part of the code so we'll need a good way to test this.

selwin:
What if the horse has already completed the job, during the short time window in which `worker.heartbeat()` is called? If the job is already complete, the horse PID would no longer exist, right?

selwin:
With regards to the exception, I'm inclined to use a new exception type, so each use case has a specific exception type.

Kriechi:
Please correct me if I'm wrong, but my basic understanding is that all child processes linger around in a Zombie state after exiting. Only once the parent process calls waitpid (or similar syscalls) the process exit information is "harvested" and the child process actually removed.

So even if the horse exits while we are sending the heartbeat, the horse will exist as Zombie until we call waitpid - whenever that happens.

selwin:
Alright, I didn't know that child processes stay around until `waitpid` is called. If that's the case, then we don't need to handle `ChildProcessError `.

Kriechi:
I updated the exception types. Please take a look.

Kriechi:
Just for future reference, here is some explanation about the zombie process & waitpid behaviour:
> A child that terminates, but has not been waited for becomes a "zombie". The kernel maintains a minimal set of information about the zombie process (PID, termination status, resource usage information) in order to allow the parent to later perform a wait to obtain information about the child.

> As long as a zombie is not removed from the system via a wait, it will consume a slot in the kernel process table, and if this table fills, it will not be possible to create further processes. If a parent process terminates, then its "zombie" children (if any) are adopted by init(8), which automatically performs a wait to remove the zombies.

https://www.tutorialspoint.com/unix_system_calls/waitpid.htm

Kriechi:
I added two tests for a slow and a fast job, which should produce different numbers of heartbeats. 

As I didn't want to rewrite the `@slow` decorator, I couldn't make use of `@pytest.mark.parametrize` to simplify the test (would be a lot cleaner and more intuitive). But if you agree, I'm happy to send another PR, once this one has landed.

rq/defaults.py
```diff
@@ -2,5 +2,5 @@
 DEFAULT_QUEUE_CLASS = 'rq.Queue'
 DEFAULT_WORKER_CLASS = 'rq.Worker'
 DEFAULT_CONNECTION_CLASS = 'redis.StrictRedis'
-DEFAULT_WORKER_TTL = 420
+DEFAULT_WORKER_TTL = 30
```
selwin: We shouldn't change `DEFAULT_WORKER_TTL`

rq/worker.py
```diff
@@ -621,9 +621,20 @@ def monitor_work_horse(self, job):
...
                 if e.errno != errno.EINTR:
                     raise
+                # Send a heartbeat to keep the worker alive.
+                self.heartbeat()
 
     def _monitor_work_horse_tick(self, job):
-        _, ret_val = os.waitpid(self._horse_pid, 0)
+        while True:
+            try:
+                with UnixSignalDeathPenalty(self.default_worker_ttl - 15, HorseMonitorTimeoutException):
```
selwin: I think we can hard code this so that heartbeat is always sent every 30 seconds, when the worker is performing a job. Worker keys can then be set to always expire in 45 seconds when job is under way so we'll know that a worker is dead in less than 1 minute's time.

What do you think?

selwin:
In general, I think this PR is on the right track :).

rq/worker.py
Kriechi: Generally, I'm ok with that.
However, the expiration timeout is set in:
`heartbeat()` with `timeout = max(timeout, self.default_worker_ttl)`. 
420 is always bigger than 45.

I'm not sure how I should change this without causing havoc in the rest of the worker...
We could introduce a new `DEFAULT_WORKER_BUSY_TTL = 45` and make it more explicit which timeout should be used when.

Kriechi: I pushed a new commit with a change to the heartbeat timeout.

Kriechi:
It looks like my method of testing `self.assertEqual(mocked.call_count, 7)` does not work on Travis, because on my machine 11 and 7 are the correct call counts.

Any ideas to get stable tests?

selwin:
Try calling `worker.execute_job(job, queue)` directly. This bypasses a good chunk of code. I didn't realize `heartbeat()` is called 7 times just to run a 3 second job. This is an area that we can look into optimizing next.

rq/worker.py
```diff
@@ -621,9 +621,21 @@ def monitor_work_horse(self, job):
                 # which we don't want to catch, so we re-raise those ones.
                 if e.errno != errno.EINTR:
                     raise
+                # Send a heartbeat to keep the worker alive.
+                self.heartbeat()
 
     def _monitor_work_horse_tick(self, job):
-        _, ret_val = os.waitpid(self._horse_pid, 0)
+        while True:
```
selwin: At the moment, we have two nested `while True` loops, one in `monitor_work_horse()` and one in `_monitor_work_horse_tick()`. We should simplify this code by moving the heartbeat and death penalty to `monitor_work_horse()`.

rq/defaults.py
```diff
@@ -3,4 +3,5 @@
 DEFAULT_WORKER_CLASS = 'rq.Worker'
 DEFAULT_CONNECTION_CLASS = 'redis.StrictRedis'
 DEFAULT_WORKER_TTL = 420
+DEFAULT_WORKER_BUSY_TTL = 45
```
selwin: Can we call this `DEFAULT_JOB_MONITORING_INTERVAL` and change it to `30`? Workers can then call `heartbeat(DEFAULT_WORKER_BUSY_TTL + 15)`.

This is easier to explain and document if we choose to make the monitoring interval configurable in the future.

rq/worker.py
Kriechi: `UnixSignalDeathPenalty` should only be used around `os.waitpid`. Otherwise we might end up interrupting `handle_job_failure` or or code paths further down in that function.
I will implement a cleaner loop - that keeps the alarm as close as possible to the waitpid.

Kriechi:
I pushed another change set - I think we are getting closer to a mergeable solution!

Kriechi:
One last question, which might be unrelated:
What happens to a worker process (not the horse), when the worker-key in redis expires?
The the worker process shutdown? Or re-register itself?

As far as I understand it, right now the process is still there and might actually execute jobs, although in redis the worker is already gone...?

selwin:
@Kriechi if worker's key in Redis expires, the worker will keep going. The worker's key should be created again during `heartbeat()`.

rq/worker.py
```diff
@@ -696,7 +701,7 @@ def prepare_job_execution(self, job):
         with self.connection._pipeline() as pipeline:
             self.set_state(WorkerStatus.BUSY, pipeline=pipeline)
             self.set_current_job_id(job.id, pipeline=pipeline)
-            self.heartbeat(timeout, pipeline=pipeline)
+            self.heartbeat(pipeline=pipeline)
```
selwin: When preparing to execute job, `heartbeat()` should be called with TTL of `DEFAULT_JOB_MONITORING_INTERVAL + 15`.

tests/test_worker.py
```diff
@@ -247,6 +248,23 @@ def test_heartbeat(self):
         self.testconn.hdel(w.key, 'birth')
         w.refresh()
 
+    @slow
+    def test_heartbeat_busy(self):
+        """Periodic heartbeats while horse is busy with long jobs"""
+        q = Queue()
+        w = Worker([q])
+
+        for timeout, expected_heartbeats in [(3, 0), (35, 1), (65, 2)]:
```
selwin: This test would take about 100 seconds to run, which is a long time (the CI builds will take a lot of resources). We should expose `Worker.job_monitoring_interval` that allows us to customize this monitoring interval and change it to (for example 2 seconds) during tests.

rq/worker.py
```diff
@@ -610,8 +610,13 @@ def monitor_work_horse(self, job):
...
         while True:
             try:
-                self._monitor_work_horse_tick(job)
+                with UnixSignalDeathPenalty(DEFAULT_JOB_MONITORING_INTERVAL, HorseMonitorTimeoutException):
+                    retpid, ret_val = os.waitpid(self._horse_pid, 0)
                 break
+            except HorseMonitorTimeoutException:
+                # Horse has not exited yet and is still running.
+                # Send a heartbeat to keep the worker alive.
+                self.heartbeat(DEFAULT_JOB_MONITORING_INTERVAL + 15)
```
selwin: Nitpick, I think we can safely change this to `DEFAULT_JOB_MONITORING_INTERVAL + 5` so dead workers will be discovered faster.

selwin:
@Kriechi yes, I think this PR is almost ready. My last round of comments were all about minor nitpicky details. Thanks for your effort, this makes RQ a lot better for running long running tasks.

Kriechi:
Done!

rq/worker.py
```diff
@@ -157,7 +157,7 @@ def find_by_key(cls, worker_key, connection=None, job_class=None,
 
     def __init__(self, queues, name=None, default_result_ttl=None, connection=None,
                  exc_handler=None, exception_handlers=None, default_worker_ttl=None,
-                 job_class=None, queue_class=None):  # noqa
+                 job_class=None, queue_class=None, default_job_monitoring_interval=None):  # noqa
```
selwin: One last nitpick (hopefully ;) ). Please change this to `job_monitoring_interval= DEFAULT_JOB_MONITORING_INTERVAL` .

`default_result_ttl` is prefixed with `default_` because it's used as default if jobs don't specify their own `result_ttl`.

rq/worker.py
```diff
@@ -542,7 +546,7 @@ def heartbeat(self, timeout=0, pipeline=None):
         The effective timeout can never be shorter than default_worker_ttl,
         only larger.
         """
-        timeout = max(timeout, self.default_worker_ttl)
+        timeout = timeout or self.default_worker_ttl
```
selwin: The docstring mentions that `The effective timeout can never be shorter than default_worker_ttl, only larger.` Since this is no longer true, the docstring should be changed.

rq/timeouts.py
```diff
@@ -5,18 +5,31 @@
...
+    timeout value.
+    """
+    pass
+
+
 class BaseDeathPenalty(object):
     """Base class to setup job timeouts."""
 
-    def __init__(self, timeout):
+    def __init__(self, timeout, exception):
```
selwin: `exception` should default to `JobTimeoutException`, this may be used by other custom workers so we should keep this backward compatible.

rq/worker.py
Kriechi: Shouldn't this be handled by https://github.com/rq/rq/pull/945/files/7ca105a2d5cab1551e7ba76863d173591a2db416#diff-1aac421e8c53e895ee40f3b7dc63144fR186 ?
I followed the same "style" as for `default_worker_ttl`...

rq/worker.py
Kriechi: Done.

rq/timeouts.py
Kriechi: Done.

rq/worker.py
Kriechi: Sorry, the commit got rewritten - and the diff lost. Here is a working link: https://github.com/rq/rq/pull/945/files#diff-1aac421e8c53e895ee40f3b7dc63144fR186

selwin: Feel free to change that "style" as well :). Default argument should be specified directly ini `__init__`.

Kriechi:
I pushed another commit, which improves the default worker option handling.
Since I already touched parts of the CLI option parsing (because it was missing defaults), I added a new option for the job monitoring interval.

rq/worker.py
```diff
@@ -155,9 +155,11 @@ def find_by_key(cls, worker_key, connection=None, job_class=None,
...
         return worker
 
-    def __init__(self, queues, name=None, default_result_ttl=None, connection=None,
-                 exc_handler=None, exception_handlers=None, default_worker_ttl=None,
-                 job_class=None, queue_class=None, default_job_monitoring_interval=None):  # noqa
+    def __init__(self, queues, name=None, default_result_ttl=DEFAULT_RESULT_TTL,
+                 connection=None, exc_handler=None, exception_handlers=None,
+                 default_worker_ttl=DEFAULT_WORKER_TTL, job_class=None,
+                 queue_class=None,
+                 default_job_monitoring_interval=DEFAULT_JOB_MONITORING_INTERVAL):  # noqa
```
selwin: Could you please change the keyword `default_job_monitoring_interval` to `job_monitoring_interval`?

rq/cli/cli.py
```diff
@@ -217,6 +220,7 @@ def worker(cli_config, burst, logging_level, name, results_ttl,
                                          connection=cli_config.connection,
                                          default_worker_ttl=worker_ttl,
                                          default_result_ttl=results_ttl,
+                                         default_job_monitoring_interval=job_monitoring_interval,
```
selwin: Could you please change the keyword `default_job_monitoring_interval` to `job_monitoring_interval`?

Kriechi: Done.

rq/worker.py
Kriechi: Done.

selwin:
I have reviewed this PR and it seems good to me. Will let it sit for a bit and review it again sometime this weekend before I merge it in just in case I missed something (don't want to rush this since this is an important change).

selwin:
@Kriechi if you have the time, could you please try running RQ with this patch a test run on your test environment with long running jobs?

dralley:
@selwin We've just run into this same issue, I will test it tomorrow and report back here

rq/worker.py
```diff
@@ -483,7 +481,7 @@ def work(self, burst=False, logging_level="INFO"):
                         self.log.info('Stopping on request')
                         break
 
-                    timeout = None if burst else max(1, self.default_worker_ttl - 60)
+                    timeout = None if burst else max(1, self.default_worker_ttl - 15)
```
dralley: This is perhaps an ignorant question, since I'm new to this codebase, but why hardcode this value at all?   Would there be a reason not to do something like:

```timeout = None if burst else max(1, int(self.default_worker_ttl / 2) - 1)```

**Comment by dralley** [comment]
This patch worked well in my testing.  Thank you @Kriechi !

**Comment by selwin** [comment]
Thanks @Kriechi, I merged the PR in. I'll let it sit in master for a few weeks so people get a chance to play with it before making a new release.

**Comment by dralley** [comment]
@selwin would you consider doing a development release (not a full release obviously) as-is?  Our project is in alpha / early beta, and this issue is a blocker our for switching from Celery to RQ.  

https://packaging.python.org/tutorials/distributing-packages/#pre-release-versioning

**Comment by selwin** [comment]
@dralley I've never done any alpha/beta releases before. You could try installing RQ using the current master's git hash. Alternatively, send me a PR modifying `setup.py` and I'll release a beta version on PyPI.

**Comment by frbry** [comment]
Hi @Kriechi, @selwin, I'm having some trouble with this change. Currently if the job takes longer than `self.default_job_monitoring_interval + 5` to complete, the worker Redis key expires, never to be created again. Perhaps `heartbeat()` doesn't get called? Any ideas?

**Comment by selwin** [comment]
@frbry are you sure you're running the latest version of RQ? I manually tried this but was unable to replicate this.

**Comment by frbry** [comment]
@selwin I'm using `rq==0.12.0`. I don't know if it makes any difference but I'm enqueuing jobs via django_rq (`django-rq==1.2.0`).

I tried switching to `rq==0.9.2` and the issue went away.

This is how I reproduce the issue:

Start the worker.
```
(project) ➜  workspace git:(debug) ✗ heroku local orders-worker
[OKAY] Loaded ENV .env File as KEY=VALUE Format
1:40:14 PM orders-worker.1 |  16:40:14 Registering birth of worker vagrant-ubuntu-trusty-64.11369
1:40:14 PM orders-worker.1 |  16:40:14 RQ worker u'rq:worker:vagrant-ubuntu-trusty-64.11369' started, version 0.12.0
1:40:14 PM orders-worker.1 |  16:40:14 *** Listening on orders...
1:40:14 PM orders-worker.1 |  16:40:14 Sent heartbeat to prevent worker timeout. Next one should arrive within 420 seconds.
1:40:14 PM orders-worker.1 |  16:40:14 Cleaning registries for queue: orders
1:40:14 PM orders-worker.1 |  16:40:14 *** Listening on orders...
1:40:14 PM orders-worker.1 |  16:40:14 Sent heartbeat to prevent worker timeout. Next one should arrive within 420 seconds.
```

The worker appears in Redis (note TTL):
worker created

Enqueue a job:
```
In [1]: import django_rq
In [2]: queue = django_rq.get_queue('orders')
In [3]: queue.enqueue('project.tasks.test_job')
Out[3]: Job(u'd94f3f47-8b1b-4900-b752-31d9b672d1c7', enqueued_at=datetime.datetime(2018, 9, 14, 13, 43, 25, 593797))
```

Worker starts on the job:
```
1:43:25 PM orders-worker.1 |  16:43:25 orders: project.tasks.test_job() (d94f3f47-8b1b-4900-b752-31d9b672d1c7)
1:43:25 PM orders-worker.1 |  16:43:25 Sent heartbeat to prevent worker timeout. Next one should arrive within 420 seconds.
1:43:25 PM orders-worker.1 |  16:43:25 Sent heartbeat to prevent worker timeout. Next one should arrive within 35 seconds.
```

35 seconds passes, the job is still running, there is no heartbeat in the logs anymore. Worker disappears in Redis. The job finishes in the background. Now there are two different things happening after this point:
1. If the job completes within 420 seconds, heartbeat comes back and the worker reappears in Redis.
2. if the job takes longer than 420 seconds (with custom queue default timeout, say, 6000), worker does not come back.

Let me know if you can replicate any of this.

frbry:
‘heroku local orders-worker’ basically creates a new process with this command: ‘newrelic-admin run-program python manage.py rqworker orders --worker-class rq.SimpleWorker’.

selwin:
Ah ok, this seems to be a bug in `SimpleWorker`. Since the worker itself performs the job, it should send heartbeats with much longer ttl, similar to the previous RQ version.

PR #947, opened 2018-04-15
support workder name override from config file

houqp:


selwin:
Thanks for the PR. Would you mind adding this new config option to the docs? The file is under `/docs/docs/workers.md`

houqp:
Sorry that I missed your message, I have updated the doc

PR #954, opened 2018-04-30
Issue 872

theodesp:
Updated PR. Indicate infinite TTL


selwin:
Thanks for the PR. Could you remove the `INFINITE_TTL` from constants.py? Not a big fan of this change, would rather compare the result to `-1` directly.

selwin:
Thanks!

Issue #921, opened 2018-01-12
Python 3.7 compatibility


mattrobenolt:
In quick testing, there's at least one issue with 3.7 and rq:

https://github.com/rq/rq/blob/master/rq/queue.py#L61

`async` will become a reserved word and this results in a `SyntaxError`.

samuelcolvin:
pull request welcome :smile:

mattrobenolt:
Yup, I probably will. :)

Do you have a preference for how to do this? It'll be an API change since we can't use `async` as a kwarg anymore. What I was thinking is this:

* Replace with `sync` and make this work like the exact opposite of `async`. So the default just being `sync=False`. This is a breaking API change and would probably require a 0.11 bump? Not sure what your versioning scheme is.
* We can try to maintain backwards compat by accepting `**kwargs` then checking for `kwargs.get('async')`.

I'd prefer not to maintain the backwards compat, but this might be a common enough use case for people that it'd be better to? Your call. I'm fine with implementing either and throwing it up as  PR.

In the meantime, not sure if there are any other 3.7 issues, since I couldn't even get it to run due to this being a SyntaxError, but I don't anticipate there being other issues.

mattrobenolt:
Yeah, we can do something like:

```python
if 'async' in kwargs:
    raise ValueError('use sync now')
```

Or whatever is preferred. We just can't leave it around as-is without a SyntaxError.

selwin:
Agree with @samuelcolvin:
1. Change `async` to `sync`
2. Catch `async` in kwargs and give a helpful error message.

We can then release version 1.0 which is backwards incompatible with earlier versions. With this version, I want to also convert `FailedQueue` to `FailedJobRegistry` which scales a lot better to a large number of jobs.

mattrobenolt:
Cool, I'll poke at this tonight probably and see if there are any other issues with python 3.7 after this is fixed. I'll make sure the test suite all passes.

theodesp:
@mattrobenolt any progress on this?

mattrobenolt:
I have not yet, unfortuantely.

mikob:
I've not done "2. catch async in kwargs and give a helpful error message" but is this the right direction so far?

samuelcolvin:
I made a comment on the PR, but perhaps better to comment here:

Surely it would be clearer to use `asynchronous` or `async_` to keep things the same rather than switching from `async=True` to `sync=False`?

selwin:
[...]

PR #963, opened 2018-06-06
Changed Queue param 'async' to 'sync' to be Python 3.7 compatible.

mikob:
Got a start on this one.
#921

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ your typically lengthy or blocking function:
 import requests
 
 def count_words_at_url(url):
-    """Just an example function that's called async."""
+    """Just an example function that's called asynchronously."""
     resp = requests.get(url)
     return len(resp.text.split())
 ```
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -65,7 +65,7 @@ job function.
     and marked as `failed`. Its default unit is second and it can be an integer or a string representing an integer(e.g.  `2`, `'2'`). Furthermore, it can be a string with specify unit including hour, minute, second(e.g. `'1h'`, `'3m'`, `'5s'`).
 * `result_ttl` specifies the expiry time of the key where the job result will
   be stored
-* `ttl` specifies the maximum queued time of the job before it'll be cancelled. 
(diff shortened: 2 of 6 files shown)


Florents-Tselai:
I'd prefer to have an `async_` which follows the "general" (?) convention of adding an `_` in the end. That's the case for other packages that say have a `class_` variable defined (e.g. SQLAlchemy comes to mind).

Switching to `sync` / `synchronous` would be awkward as the point of having a queue system is to have asynchronous tasks :) 
Just my two cents.

selwin:
@mikob do you mind changing the `synchronous` keyword to `asynchronous`? Sorry about this.

Also, could you please open this PR against [v1.0](https://github.com/rq/rq/tree/v1.0) branch?

Hopefully I'll be able to finish working on https://github.com/rq/rq/tree/failed-job-registry in around a month's time and we'll release 1.0 then.

mikob:
@selwin, would you reconsider "sync"? Here's my thinking:

 "Synchronous" is the root word, and "a-synchronous" is just a negation of that. It's harder to fathom a double negation (eg. "async=false" is a double negative) Hence it makes more sense to go with `sync`, and it's also easier to spell out then `synchronous` and more palatable than the lazy fix `async_` IMO. 

The standard node library goes with "sync" such as with `fs.syncReadFile` since async is the default. In our case here, a job queue, async is also the default.

mikob:
@selwin have you reconsidered using "sync" instead -- as per my reasoning above?

selwin:
I'm really 50/50 on this topic.

Using `asynchronous` would provide an easier upgrade path for existing users. On the other hand, if I were designing this library from scratch on Python 3.7, I'd probably choose to use `sync` for the reason you mentioned; `asynchronous` is a bit tricky to spell for non native English speakers like me.

Let's just keep using `sync` since this will be better for the long term. Could you please reopen this PR against the `v1.0` branch?

mikob:
reopened on `v1.0` branch https://github.com/rq/rq/pull/970

PR #967, opened 2018-06-08
Windows Clarification

W7PEA:
You can use WSL and run RQ and Redis on Windows by running them in a bash shell within windows.


selwin:
Thanks!

PR #970, opened 2018-06-27
Changed Queue param 'async' to 'sync' to be Python 3.7 compatible.

mikob:


selwin:
Thanks!

selwin:
@mikob sorry about this, after I merged in this PR into the v1.0 branch, I saw https://github.com/rq/rq/pull/977 which introduces the keyword `is_async` which:
1. Provides an easier upgrade path (this won't break existing codes that uses positional arguments to specify `async`)
2. Is easy to spell and still sounds good

I think we'll go with `is_async` for RQ v1.0.

Thanks a lot for your help.

mikob:
@selwin no problem! Thanks for the update!

PR #971, opened 2018-07-02
Rename argument named `async` to `asynk`

Flimm:
The name `async` is not allowed in Python 3.7.

Fixes https://github.com/rq/rq/issues/921


selwin:
Decided to go with `is_async` proposed by @chevell in https://github.com/rq/rq/pull/977. Thanks!

Issue #972, opened 2018-07-03
"async" function argument causing SyntaxError since Python3.7


Madi7:
In rq/queue.py:

    def __init__(self, name='default', default_timeout=None, connection=None,
                 async=True, job_class=None):
        ...

Python3.7's release note:

    Backwards incompatible syntax changes:

        async and await are now reserved keywords.

blueyed:
Fixed in https://github.com/rq/rq/commit/c2b939d2df7df2bd0a6e5a73823a07baecc17895.

regan-karlewicz:
Any idea when there will be a release to fix this?

selwin:
Version 0.12 is now on PyPI https://pypi.org/project/rq/

PR #975, opened 2018-07-04
Make Python 3.7 compatibility and keep backward compatibility

kitsuyui:
- Close: https://github.com/rq/rq/issues/972
- Keep async keyword for older Python by using ** keyword arguments.
  - Thanks to https://github.com/rq/rq/issues/921#issuecomment-358096956
- Add sync keyword argument that means the inverse of async.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
-  - "3.6-dev"
+  - "3.7-dev"
   - "pypy"
 install:
   - pip install -e .
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -389,17 +389,17 @@ class TestJob(RQTestCase):
         assert get_failed_queue(self.testconn).count == 0
 
     def test_job_access_within_synchronous_job_function(self):
-        queue = Queue(**{'async': False})
+        queue = Queue(sync=True)
         queue.enqueue(fixtures.access_self)
 
     def test_job_async_status_finished(self):
-        queue = Queue(**{'async': False})
+        queue = Queue(sync=True)
         job = queue.enqueue(fixtures.say_hello)
         self.assertEqual(job.result, 'Hi there, Stranger!')
         self.assertEqual(job.get_status(), JobStatus.FINISHED)
 
     def test_enqueue_job_async_status_finished(self):
-        queue = Queue(**{'async': False})
+        queue = Queue(sync=True)
         job = Job.create(func=fixtures.say_hello)
         job = queue.enqueue_job(job)
         self.assertEqual(job.result, 'Hi there, Stranger!')
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -389,17 +389,17 @@ class TestJob(RQTestCase):
         assert get_failed_queue(self.testconn).count == 0
 
     def test_job_access_within_synchronous_job_function(self):
-        queue = Queue(async=False)
+        queue = Queue(**{'async': False})
         queue.enqueue(fixtures.access_self)
 
     def test_job_async_status_finished(self):
-        queue = Queue(async=False)
+        queue = Queue(**{'async': False})
         job = queue.enqueue(fixtures.say_hello)
         self.assertEqual(job.result, 'Hi there, Stranger!')
         self.assertEqual(job.get_status(), JobStatus.FINISHED)
 
     def test_enqueue_job_async_status_finished(self):
-        queue = Queue(async=False)
+        queue = Queue(**{'async': False})
         job = Job.create(func=fixtures.say_hello)
         job = queue.enqueue_job(job)
         self.assertEqual(job.result, 'Hi there, Stranger!')
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -389,17 +389,17 @@ class TestJob(RQTestCase):
         assert get_failed_queue(self.testconn).count == 0
 
     def test_job_access_within_synchronous_job_function(self):
-        queue = Queue(sync=True)
+        queue = Queue(asynchronous=False)
         queue.enqueue(fixtures.access_self)
 
     def test_job_async_status_finished(self):
-        queue = Queue(sync=True)
+        queue = Queue(asynchronous=False)
         job = queue.enqueue(fixtures.say_hello)
         self.assertEqual(job.result, 'Hi there, Stranger!')
         self.assertEqual(job.get_status(), JobStatus.FINISHED)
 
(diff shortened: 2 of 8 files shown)


kitsuyui:
This also works fine with my Python 3.7.0 environment.

```
$ py.test
=============================================== test session starts ===============================================
platform darwin -- Python 3.7.0, pytest-3.6.2, py-1.5.4, pluggy-0.6.0
rootdir: /Users/yui/Documents/Develop/repo/github.com/rq/rq, inifile:
collected 214 items

tests/test_cli.py .................                                                                         [  7%]
tests/test_connection.py .....                                                                              [ 10%]
tests/test_decorator.py ...........                                                                         [ 15%]
tests/test_helpers.py .                                                                                     [ 15%]
tests/test_job.py ................................................                                          [ 38%]
tests/test_queue.py ..................................................                                      [ 61%]
tests/test_registry.py .................                                                                    [ 69%]
tests/test_sentry.py .                                                                                      [ 70%]
tests/test_utils.py ........                                                                                [ 73%]
tests/test_worker.py ............................................ssssssss..                                 [ 99%]
tests/test_worker_registration.py ..                                                                        [100%]

====================================== 206 passed, 8 skipped in 7.13 seconds ======================================
```

samuelcolvin:
surely this is more confusing, better to raise an informative error if `async` is included?

samuelcolvin:
Or at least a deprecation warning, then remove in a future release.

kitsuyui:
@samuelcolvin 
Ya, sorry for bringing up completed topics again.
I don't feel good that breaking backward compatiblity.
I will add more commit to raise deprecation warning.

samuelcolvin:
:+1: I think that makes sense.

kitsuyui:
I did it.

jd:
Missing fixes #972 (so I add it here for cross linking)

kitsuyui:
I added the links of prior PR and co-author statuses into commit message.
Thanks you for @mikob and @Flimm.

samuelcolvin:
why have you added `sudo` and `dist`? I don't think it's necesary.

tests/test_job.py
```diff
@@ -389,17 +389,17 @@ def test_job_access_within_job_function(self):
         assert get_failed_queue(self.testconn).count == 0
 
     def test_job_access_within_synchronous_job_function(self):
-        queue = Queue(async=False)
+        queue = Queue(**{'async': False})
```
samuelcolvin: best to switch these to use `sync=True`

kitsuyui: Thank you, I will fix this!

kitsuyui:
@samuelcolvin 

> why have you added `sudo` and `dist`? I don't think it's necesary.

I have read Flimm's commit https://github.com/rq/rq/pull/971/files
And I also have read these articles:
https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System
https://docs.travis-ci.com/user/languages/python/#Development-releases-support

So I had wondered these dist and sudo combination is needed.
Sorry but I didn't inspect that is real. It may be overdoing. I will rewind it.

samuelcolvin:
see [here](https://github.com/samuelcolvin/pydantic/blob/master/.travis.yml#L5-L7)

I think all you need is to add `3.7-dev`

kitsuyui:
@samuelcolvin 
Thank you very much.
Test passed. Just adding 3.7-dev is enough as you say.

samuelcolvin:
no problem, @selwin will need to review and merge.

rodocontegni:
Won't this break backwards compatibility for code that uses positional arguments instead of kwargs when instantiating a `Queue`?

E.g.:
```python

q = Queue('default', None, Redis(), True)
```

chevell:
Changing the keyword to mean the inverse is a bad idea, please do not use this approach as it is almost certainly going to break peoples' code.

Why not simply change it to `is_async`?

kitsuyui:
@rodocontegni and @chevell 
Thank you.
ya, I have overlooked about the case when async is used as non-keyword arguments.
I'm willing to adapt to the case and add tests with more attentively.

kitsuyui:
I finally tried with replacing asynchronous.
And I added the tests to cover the all patterns.

Like this:

```python
def test_arguments_for_backward_compatibility(self):
    # test when arguments not specified
    self.assertEqual(Queue()._async, True)
    # test async/asynchronous as non keyword argument usage
    self.assertEqual(Queue('default', None, None, True)._async, True)
    self.assertEqual(Queue('default', None, None, False)._async, False)
    # test asynchronous for backward compapatibility
    self.assertEqual(Queue(asynchronous=True)._async, True)
    self.assertEqual(Queue(asynchronous=False)._async, False)
    # test async for backward compapatibility and it raises DeprecationWarning
    with warnings.catch_warnings(record=True) as w:
        warnings.simplefilter('always')
        self.assertEqual(Queue(**{'async': True})._async, True)
        assert len(w) == 1
        assert issubclass(w[-1].category, DeprecationWarning)
        warnings.resetwarnings()
    with warnings.catch_warnings(record=True) as w:
        warnings.simplefilter('always')
        self.assertEqual(Queue(**{'async': False})._async, False)
        assert len(w) == 1
        assert issubclass(w[-1].category, DeprecationWarning)
        warnings.resetwarnings()
    # nasty edge cases that specified both
    with self.assertRaises(ValueError) as context:
        Queue(asynchronous=True, **{'async': True})
    with self.assertRaises(ValueError) as context:
        Queue(asynchronous=True, **{'async': False})
    with self.assertRaises(ValueError) as context:
        Queue(asynchronous=False, **{'async': True})
    with self.assertRaises(ValueError) as context:
        Queue(asynchronous=False, **{'async': False})
```

kitsuyui:
So the Codecov Report becamee green from red.

kitsuyui:
@selwin @samuelcolvin
Sorry about my churning.
Would you please re-review this?

.travis.yml
```diff
@@ -8,7 +8,7 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
-  - "3.6-dev"
+  - "3.7-dev"
```
jd: This sounds like the dev branch of 3.7, not 3.7 final.

kitsuyui: I think 3.7-dev is the latest version when without specifying dist at current.

- https://github.com/travis-ci/travis-ci/issues/9815
- https://github.com/rq/rq/pull/975#issuecomment-402719107

selwin:
Decided to go with `is_async` keyword proposed by @chevell here: https://github.com/rq/rq/pull/977 . Thanks!

PR #977, opened 2018-07-06
Replace 'async' keyword with 'is_async' for Queue objects

chevell:
This PR provides Python 3.7 compatibility by changing the Queue object's `async` keyword to `is_async`. For backwards compatibility, `**kwargs` is checked for the `async` keyword and will continue to work, but will raise a warning.

All tests and documentation references are updated to use `is_async` instead of `async`.

Some other PR's are suggesting reversing the keyword (i.e. `sync`) which will break any code that uses ordered keywords rather than named keywords, so this PR is intended to provide a safer alternative.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
-  - "3.6-dev"
+  - "3.7-dev"
   - "pypy"
 install:
   - pip install -e .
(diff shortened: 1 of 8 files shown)


chevell:
I didn't realise travis doesn't support the stable 3.7 release yet so I'm removing 3.7 from .travis.yml and leaving 3.7-dev. Otherwise, all tests pass.

selwin:
Thanks!

Issue #959, opened 2018-05-29
changing the rq.worker logger date time format


shikharsg:
Hi,

I was wondering if I could change the logging settings of `rq.worker` here: https://github.com/rq/rq/blob/master/rq/logutils.py#L10. Is there any way to do this without changing the code in rq/logutils.py?

I'm asking because I want the text in line number 4 below to be logged to my log file. While line 5,6,8 is going to my log file, line 4 is not. I also need to change the format from time to full date time.

```
08:54:21 RQ worker 'rq:worker:my-vm.3203' started, version 0.10.0
08:54:21 *** Listening on q1...
08:54:21 Cleaning registries for queue: q1
08:54:26 q1: test_task.test_task('test') (b9c59efb-ae21-47db-9f19-112aa2b81169)
2018-05-29 08:54:26,733 [   INFO] [root] startedtest
2018-05-29 08:54:39,288 [   INFO] [mymodule] testdone
08:56:19 q1: Job OK (b9c59efb-ae21-47db-9f19-112aa2b81169)
2018-05-29 08:56:19,373 [   INFO] [rq.worker] q1: Job OK (b9c59efb-ae21-47db-9f19-112aa2b81169)
08:56:19 Result is kept for 500 seconds
```

Thanks.

selwin:
If you use RQ's `rq worker` script, there's currently no easy way to customize this. I think we should modify RQ's worker script to accept a `--date-format` argument is reasonable.

We should also standardize the datetime format for all log messages emitted by RQ (for example lines 1, 2, 3 have different date formats than line 8).

Would be awesome if you could open a pull request for this.

shikharsg:
Thanks for having a look at this. I have generated a PR now:: #979

selwin:
Fixed in https://github.com/rq/rq/pull/979

PR #979, opened 2018-07-14
added logging formatting options

shikharsg:
The format of the worker logs cannot be changed as of now. This pull request attempts to change that. See issue [#959](https://github.com/rq/rq/issues/959).
Not sure if these changes are the most elegant solution. Maybe we should allow the user to pass in a log handler to the worker? Much like how you can pass in an exception handler to a worker.


selwin:
This PR looks good. Could you also document these two options in this file? https://github.com/rq/rq/blob/master/docs/docs/workers.md

rq/logutils.py
```diff
@@ -5,14 +5,16 @@
 import logging
 
 from rq.utils import ColorizingStreamHandler
+from rq.defaults import (DEFAULT_LOGGING_FORMAT,
+                         DEFAULT_LOGGING_DATE_FORMAT)
 
 
-def setup_loghandlers(level):
+def setup_loghandlers(level, datefmt=DEFAULT_LOGGING_DATE_FORMAT,
```
selwin: One last minor detail, could you please change the `datefmt` to `date_format` and `logfmt` to `log_format`? This should keep the variable naming consistent across RQ's code base.

shikharsg: Sorry my bad. Fixed now

selwin:
Thanks!

PR #982, opened 2018-07-25
add is_async property to queue

perobertson:
When creating a queue we are allowed to specify if its async or not, but there is no way to know if a current queue instance is async or not without knowing the internals of the code.

This just exposes that setting as a property on the queue.


selwin:
Thanks!

PR #985, opened 2018-07-30
Update docs to use print function vs print statement

alysivji:
I was working my way through the documentation and had to change a few print statements to print functions for the examples to run in Python 3.

As most new projects begin with Python 3, I thought it would be a good idea to update the examples in the documentation.

Please let me know if I missed anything. Thanks!


selwin:
Thanks!

PR #991, opened 2018-09-18
Update/add flag for description logging

finnci:
Add ability to disable logging of all job descriptions.

I've often found that job descriptions in this case contain sensitive information, or simply are verbose and not something we want to be filling up our logs with.

This is a pretty simple solution to that, with some tests.

### DIFF
--- a/docs/docs/workers.md
+++ b/docs/docs/workers.md
@@ -63,7 +63,7 @@ In addition to `--burst`, `rq worker` also accepts these arguments:
 * `--connection-class`: Redis connection class to use, defaults to `redis.StrictRedis`.
 * `--log-format`: Format for the worker logs, defaults to `'%(asctime)s %(message)s'`
 * `--date-format`: Datetime format for the worker logs, defaults to `'%H:%M:%S'`
-* `--log_job_description`: Toggle logging of a job description, defaults to True.
+* `--disable-job-desc-logging`: Toggle logging of a job description, defaults to True.
 
 ## Inside the worker
 
--- a/docs/docs/workers.md
+++ b/docs/docs/workers.md
@@ -63,7 +63,7 @@ In addition to `--burst`, `rq worker` also accepts these arguments:
 * `--connection-class`: Redis connection class to use, defaults to `redis.StrictRedis`.
 * `--log-format`: Format for the worker logs, defaults to `'%(asctime)s %(message)s'`
 * `--date-format`: Datetime format for the worker logs, defaults to `'%H:%M:%S'`
-
+* `--log_job_description`: Toggle logging of a job description, defaults to True.
 
 ## Inside the worker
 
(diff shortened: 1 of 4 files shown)


selwin:
Hey there, this is a useful feature, could you please add this to the docs?

rq/cli/cli.py
```diff
@@ -180,6 +180,7 @@ def info(cli_config, interval, raw, only_queues, only_workers, by_queue, queues,
 @click.option('--results-ttl', type=int, default=DEFAULT_RESULT_TTL , help='Default results timeout to be used')
 @click.option('--worker-ttl', type=int, default=DEFAULT_WORKER_TTL , help='Default worker timeout to be used')
 @click.option('--job-monitoring-interval', type=int, default=DEFAULT_JOB_MONITORING_INTERVAL , help='Default job monitoring interval to be used')
+@click.option('--log_job_description', is_flag=True, help='Toggle description logging in worker.')
```
selwin: Since job description is logged by default, it's more intuitive if we reverse this flag.

Let's rename this to `--disable-job-desc-logging`? Other ideas welcome :)

docs/docs/workers.md
```diff
@@ -63,7 +63,7 @@ In addition to `--burst`, `rq worker` also accepts these arguments:
 * `--connection-class`: Redis connection class to use, defaults to `redis.StrictRedis`.
 * `--log-format`: Format for the worker logs, defaults to `'%(asctime)s %(message)s'`
 * `--date-format`: Datetime format for the worker logs, defaults to `'%H:%M:%S'`
-
+* `--disable-job-desc-logging`: Toggle logging of a job description, defaults to True.
```
selwin: Could you please update the explanation to something like `Turn off job description logging` or something similar? It's only a flag, so there's no default value.

rq/cli/cli.py
```diff
@@ -180,6 +180,7 @@ def info(cli_config, interval, raw, only_queues, only_workers, by_queue, queues,
 @click.option('--results-ttl', type=int, default=DEFAULT_RESULT_TTL , help='Default results timeout to be used')
 @click.option('--worker-ttl', type=int, default=DEFAULT_WORKER_TTL , help='Default worker timeout to be used')
 @click.option('--job-monitoring-interval', type=int, default=DEFAULT_JOB_MONITORING_INTERVAL , help='Default job monitoring interval to be used')
+@click.option('--disable-job-desc-logging', is_flag=True, help='Toggle description logging in worker.')
```
selwin: Can you change the help text to `Turn off description logging`? I don't think `Toggle` is a suitable term in this particular context.

finnci:
@selwin updated as suggested, apologies for the delay.

selwin:
Thanks!

Issue #1003, opened 2018-10-23
The future of RQ and Sentry


untitaker:
Our [new SDK for Sentry will soon have a RQ integration released](https://github.com/getsentry/sentry-python/pull/139). I wonder if it makes sense to deprecate the integration in RQ itself, as it's effectively unmaintained anyway

selwin:
Would be happy to :). Is there any docs about this new integration? The only thing I can find on docs.sentry.io is https://docs.sentry.io/clients/python/integrations/rq/ which is already deprecated.

untitaker:
If there's consensus for this within RQ's core dev group I'll prepare a PR to add a similar deprecation notice to RQ's docs.

selwin:
No objections from me. @samuelcolvin any thoughts?

With the new Sentry SDK, can we keep the API the same at `rqworker --sentry-dsn="https://<key>:<secret>@sentry.io/<project>"`?

untitaker:
I opened a PR at I opened a PR at https://github.com/rq/rq/pull/1005

selwin:
I read the docs you provided. If we were to deprecate the `--sentry-dsn` argument, the setup would be a bit more complicated in that users would need to add this bit of code in their config file ([docs here](http://python-rq.org/docs/workers/)).

```python
import sentry_sdk
sentry_sdk.init("https://<key>@sentry.io/<project>")
```

Wouldn't it be better if we directly call that bit of code in `worker.py` if `--sentry-dsn` is passed in? If for some reason this is discouraged, could you please add this workaround in your pull request?

I have another question, does Sentry's new SDK work well with RQ's default `Worker` class? I'm asking because Raven's async transport doesn't work reliably with with RQ's default `Worker` class.

untitaker:
@selwin You are going to call this for your webapp anyway, which I assume shares its config code with the workers. The expectation is that most code will look like this:

```python
sentry_sdk.init(DSN, integrations=[RqIntegration(), FlaskIntegration(), ...])
```

We'd like to have this interface consistent across integrations. What exactly should I add to the PR?

I will look into fixing the transport issue but we want to avoid having to add multiple transports to the new SDK

selwin:
While I'm sure a good portion of RQ users are web developers, there are cases where RQ is not used with a web framework (not sure how big of a percentage this is).

I think the best way forward would be to:
1. Support `--sentry-dsn` for RQ's worker script, this would keep backward compatibility for existing RQ users
[...]

PR #1005, opened 2018-10-24
Deprecate sentry integration

untitaker:
See #1003


samuelcolvin:
@selwin surely best to just remove the integration directly?

selwin:
@samuelcolvin see my comment [here](
https://github.com/rq/rq/issues/1003#issuecomment-433084754)

selwin:
Fixed in #1045. Docs: https://python-rq.org/patterns/sentry/

PR #1009, opened 2018-10-27
Replacing FailedQueue with FailedJobRegistry

selwin:
Highlights of this pull request:

- `FailedQueue` has been replaced with `FailedJobRegistry`:
  * `get_failed_queue()` function has been removed. Please use `FailedJobRegistry(queue=queue)` instead.
  * `move_to_failed_queue()` has been removed.
  * RQ now provides a mechanism to automatically cleanup failed jobs. By default, failed jobs are kept for 1 year.

- RQ's custom job exception handling mechanism has also changed slightly:
  * RQ's default exception handling mechanism (moving jobs to `FailedQueueRegistry`) can be disabled by doing `Worker(disable_default_exception_handler=True)`.
  * Custom exception handlers are no longer executed in reverse order.


selwin:
Fixed in https://github.com/rq/rq/pull/1059

PR #1015, opened 2018-11-15
Pinned redis-py dep under 3.0

MarechJ:
Fixes #1014

PR #1016, opened 2018-11-16
modify zadd calls for redis-py 3.0

dash-rai:
redis-py 3.0 changes the zadd interface that accepts a single
mapping argument that is expected to be a dict.
https://github.com/andymccurdy/redis-py#mset-msetnx-and-zadd

This makes rq incompatible with redis-py <3.0. If we require compatibility with older versions, we can add a shim in compat/connections.py


marcy-buccellato:
YES! Thank you @dash-rai . This is blocking all of our new releases. I was planning to fork RQ and fork Django RQ just to hardcode the redis version here to 12.10.6, but this is much more elegant. Looking forward to this fix.

MarechJ:
Good one @dash-rai , I'll close mine which was the quick fix  #1015

samuelcolvin:
LGTM.

Would be good to get this merged and released, @selwin could you have a look at it?

selwin:
This looks good.

Do you mind removing https://github.com/rq/rq/blob/master/rq/compat/connections.py#L29 in this patch too? Since `Redis == StrictRedis` in redis 3.0, we no longer need to monkey patch `connection._zadd` and other incompatible methods.

dash-rai:
> Do you mind removing https://github.com/rq/rq/blob/master/rq/compat/connections.py#L29 in this patch too? Since `Redis == StrictRedis` in redis 3.0, we no longer need to monkey patch `connection._zadd` and other incompatible methods.

Sure, I'll do that.

dash-rai:
@selwin Updated the PR with the changes.

selwin:
Merged into master. I'll wait a week to see if there's any blocking bugs before releasing cutting a release. I'll also give related projects a heads up about this upcoming change.

yanir3:
Thanks @selwin. Can we get an ETA on the release to pypi?

terencehonles:
@yanir3 @samuelcolvin looks like it's been pushed (just now) https://pypi.org/project/rq/#history

samuelcolvin:
Great. Thanks so much.

PR #1022, opened 2018-12-06
added worker option: --logging-file, allow config file not project trree

kgk:
1.  Allow a custom logging config to be loaded from a project file.
2.  Parse a config file that is not in the project tree (any not loadable by importlib).


selwin:
Instead of requiring a different configuration file just for logging, we should modify RQ to read logging configurations from RQ's current settings file.

Could you modify this PR to read `LOGGING` variable from RQ's configuration file and provide some examples in `/docs`?

kgk:
Not sure I fully follow your comment.    What would be in the contents of the LOGGING variable?  It seems that you would prefer the logging configuration to be in the setting files .. so would it contain logging.dictConfig .  I would be happy to add that as an option. 

But for other projects that already have logging setup contained in a file, logging.fileConfig is more convenient.      In our projects, we use a single config file  to organize the logging and root logging handlers (syslog, socket, etc) and then expect all components to simply read that file.    This way I can setup a syslog root logger with a default format and use it consistently.

selwin:
Yeah, the `LOGGING` variable would contain `logging.dictConfig`.

>But for other projects that already have logging setup contained in a file, logging.fileConfig is more convenient. In our projects, we use a single config file to organize the logging and root logging handlers (syslog, socket, etc) and then expect all components to simply read that file. This way I can setup a syslog root logger with a default format and use it consistently.

Noted. Could you please add the following:
1. Document this option in https://github.com/rq/rq/blob/master/docs/docs/workers.md#worker-arguments
2. Adding some tests so our test coverage doesn't decrease

Adding the capability to read `LOGGING` variable would be nice to have, but can be done in a separate PR.

selwin:
Could you please document this option and test this?

selwin:

PR #1026, opened 2019-01-02
Add uuid to worker self.name to prevent collisions in docker setups

### LINKED ISSUE #497: ValueError exception in register_birth()
May I suggest that we provide a exception class detail to reflect this existing ValueError?
There are 2 possible ValueErrors in Worker class, and this variant is recoverable.
It would be great to have a specific Exception class to reflect this scenario, and let the caller intercept and deal with it should it be encountered.

def register_birth(self):
        ...
            raise ValueError('There exists an active worker named \'%s\' '
                             'already.' % (self.name,))

expl0ratory:
fixes #497

Real-world repros are challenging, but this occurs in docker-compose set ups, as well as in production on my kubernetes clusters.

There was not an explicit test for the property, but the functionality is covered in many of the worker tests.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -11,6 +11,7 @@ import socket
 import sys
 import time
 import traceback
+import uuid
 import warnings
 from datetime import timedelta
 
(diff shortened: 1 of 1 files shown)


selwin:
Could you modify this PR to change the worker name to just uuid string and store `hostname` and `PID` information [properly in Redis](https://github.com/rq/rq/blob/master/rq/worker.py#L552)?

selwin:
I have made the necessary changes in v1.0 branch here: https://github.com/rq/rq/commit/7097c7eb9028e2ba47209d03b95f0aa2364a59c8

Issue #997, opened 2018-10-13
total_working_time attribute of Worker class is incorrect


tbaumeist:
The total_working_time is supposed to capture the complete execution time of all jobs in that worker. However, only the microseconds of each execution is being captured. You are losing seconds and days that have passed. https://github.com/rq/rq/blob/master/rq/worker.py#L591 in def increment_total_working_time is only getting the microseconds from the timedelta object.  `job_execution_time.microseconds` https://docs.python.org/3/library/datetime.html#datetime.timedelta

Solutions: either works
1.` job_execution_time.total_seconds() * 1000`
2. `(job_execution_time.microseconds + (job_execution_time.seconds + job_execution_time.days * 24 * 3600) * 10**6) / 10**6`

See: https://stackoverflow.com/questions/18426882/python-time-difference-in-milliseconds-not-working-for-me for more background on how timedelta works.

selwin:
Fixed in https://github.com/rq/rq/pull/1047

PR #1028, opened 2019-01-07
Fix total_working_time attribute for Worker #997

Florents-Tselai:


selwin:
Could you please fix the failing test?

selwin:
Fixed in https://github.com/rq/rq/pull/1047

PR #1030, opened 2019-01-19
Improve worker logging.

tds333:
Switch some messages from warn to info because it is normal requested behavior.
Also for log messages use % formatting and parameters. It is faster and consistent with other logs in the file.


selwin:
Thanks for the PR. Could you please fix the merge conflicts?

tds333:
I have resolved the conflicts and merged back master into the branch.

PR #1036, opened 2019-01-30
fix: worker exception exit code

Chyroc:


selwin:
Do you mind explaining what bug this PR solves?

PR #1042, opened 2019-02-17
Fix #1040 queue default timeout bug.

### LINKED ISSUE #1040: enqueue_job uses wrong timeout value
The enqueue_job method of the Queue class uses the wrong timeout value.

If a Queue was constructed with a new timeout, internally stored at _default_timeout, this should be used. But enqueue_job uses the wrong global default self.DEFAULT_TIMEOUT if job hat no timeout set.

tds333:
Fix #1040 and add tests for queue default_timeout.


selwin:
Thanks!

PR #1055, opened 2019-03-16
queue.enqueue() no longer accepts `timeout` argument

selwin:


docs/docs/index.md
```diff
@@ -82,7 +83,7 @@ In the last case, it may be advantageous to instead use the explicit version of
 q = Queue('low', connection=redis_conn)
 q.enqueue_call(func=count_words_at_url,
                args=('http://nvie.com',),
-               timeout=30)
+               job_timeout=30)
```
WeatherGod: This appears to be incorrect? It doesn't seem like the call signature for `enqueue_call()` was changed?

PR #1064, opened 2019-04-08
Give better titles to documentation pages

mlissner:
Having the same titles on all my tabs is making me crazy. Hopefully this will fix it!


selwin:
Thanks!

PR #1075, opened 2019-04-30
FIX: ttl can also have zero value.

shubhamdipt:
BUG: TTL = 0 was getting converted to None.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -438,7 +438,7 @@ class Job(object):
         self.failure_ttl = int(obj.get('failure_ttl')) if obj.get('failure_ttl') else None  # noqa
         self._status = obj.get('status') if obj.get('status') else None
         self._dependency_id = as_text(obj.get('dependency_id', None))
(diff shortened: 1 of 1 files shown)


selwin:
I'm not sure why anyone would want to set `job.ttl` to 0. This would mean the job expires right away.

shubhamdipt:
In any case, zero should not be converted to None. Either zero should be documented as invalid input or it should stay as zero.

selwin:
I think we should disallow 0 TTL. Could you please modify this PR accordingly? Possibly document this as well. Thanks!

shubhamdipt:
In my opinion, we should still allow it as it covers a use case of cleaning up the job immediately. The value 0 does carry relevance in such cases.

selwin:
I think you're confusing `result_ttl` (how long a job result is stored) with `ttl` (how long a job is valid for).

When workers find a job that has expired, it will be discarded without being executed.

PR #1082, opened 2019-05-06
Fix usage example of queue.enqueue()

macintoshpie:
queue.enqueue() has arg `f`, not `func` (like enqueue_call).


selwin:
Can we get rid of `f` altogether?

macintoshpie:
I'm assuming you mean for the example, so I removed it there.

selwin:
Yes. Thanks!

PR #1083, opened 2019-05-06
Fix docs on job attributes

macintoshpie:
Job objects no longer have `.status` attribute - correct docs to use the `.get_status()` function.


selwin:
Thanks!

Issue #1085, opened 2019-05-07
Bug: inf timeout value in StartedJobRegistry


macintoshpie:
There is a bug in the worker that doesn't handle the infinite case for timeout, when `job.timeout == -1`.

This is can be fixed in the `prepare_job_execution()` function of the worker by checking if the timeout is negative before adding it to the StartedJobRegistry.

selwin:
Fixed in https://github.com/rq/rq/pull/1086

PR #1086, opened 2019-05-07
Fix timeout adding job to StartedJobRegistry

macintoshpie:
See issue #1085 

Fixed by adding a check for negative `job.timeout` value before adding the job to the `StartedJobRegistry`.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -699,7 +699,10 @@ class Worker(object):
         """Performs misc bookkeeping like updating states prior to
         job execution.
         """
-        timeout = (job.timeout or 180) + 60
+        if job.timeout < 0:
+          timeout = job.timeout
+        else:
+          timeout = (job.timeout or 180) + 60
 
         if heartbeat_ttl is None:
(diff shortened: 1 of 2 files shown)


macintoshpie:
Also, it's unclear to me if it's valid to have a timeout of 0. If so, then a check for `None` needs to be added.

rq/worker.py
```diff
@@ -699,7 +699,8 @@ def prepare_job_execution(self, job, heartbeat_ttl=None):
         """Performs misc bookkeeping like updating states prior to
         job execution.
         """
-        timeout = (job.timeout or 180) + 60
+        base_timeout = job.timeout or 180
```
selwin: Sorry for the nitpick, but can we make this simpler to read by explicitly checking for `-1`?
```python
if job.timeout == -1:
    timeout = -1
else:
    timeout = job.timeout or 180
```

macintoshpie: I think that's good. Will update the pr

PR #1089, opened 2019-05-16
Update index.md to add "import time"

christopherpickering:
added "import time"

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -31,6 +31,7 @@ simply do this:
 from rq import Queue
 from redis import Redis
 from somewhere import count_words_at_url
+import time
 
 # Tell RQ what Redis connection to use
 redis_conn = Redis()


selwin:
Thanks!

christopherpickering:
Hi Selwin,

No problem, thanks for the great instructions!

Hey I have a question on the topic that I wasn’t able to solve-

I have a Django website. When users come on one of the apps I need to login to an external website, gather some info, and return it to the user.

Sometimes the info gathered is 20-30 pieces of information.

Right now I’m logging in to the external site (unique session) for each piece of information.

When I have 100 people in the site searching at the same time the logins to the external site get to be a lot!

I’m trying to figure out how to use the same session (with python requests module) for all requests while the session is active- to save time and ease the load on the external site. And also multithread the external requests as they are now through nginx.

I was trying to us a rq worker to handle the session but didn’t have a good way to get the responses back to the user.

Have you faced something like this before with rq?

Thanks!
Christopher

PR #1090, opened 2019-05-28
Multi dependencies v1.0

olingerc:
Hi,

I recently took some time to merge the multi-dependencies feature I have been using for years now into the current 1.0 branch. I am prepared to help getting this ready but I know it will take a lot of work.

 My branch works but of course there are code stretches that feel a bit hacky.

One pain point I had, was the fact that if multiple workers listen on the same queue, it might happen that during dependency en-queuing the same job gets taken by two workers which in fact duplicates the job.

There is a test case for this in the test_queue.py "test_multi_deps_enqueued_only_once". The test fails in one in 1000 cases so the problem is really difficult to pinpoint.

Also, we will need to squash all my "WIP" commits at some point ;-)


selwin:
Hey there, since this will be a long road, I suggest getting this PR in piece by piece, over multiple releases.

Over the past few years, the job dependency feature has been stable so merging this in should be do-able if we do this carefully.

The first few steps could be:
1. Open a PR that changes `job._dependency_id` to `job._dependency_ids` (a list)? In the first step, the list will only contain a single `dependency_id` and the way dependency_id is stored in Redis shouldn't change.
2. Open a PR that changes `dependency_ids` to be stored in a Redis set (while still supporting a single dependency)

We'll decide on the next few steps should come in after the 1 and 2 are merged in.

PR #1094, opened 2019-06-03
add the ability to have the worker stop executing after a max amount of jobs

perobertson:
This adds the ability to the worker to gracefully exit after a maximum number of jobs have been processed. The behaviour is similar to `--burst` except instead of processing until all queues are empty, it will exit after the limit is reached.

This will be useful in situations where the queue is known to be busy for a long time and you want to stop part way through, push our a deploy of your application, and then have the worker start up again with the updated job code.

Let me know if you would prefer different naming or coding conventions.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -435,7 +435,7 @@ class Worker(object):
             self.set_state(before_state)
 
     def work(self, burst=False, logging_level="INFO", date_format=DEFAULT_LOGGING_DATE_FORMAT,
-             log_format=DEFAULT_LOGGING_FORMAT):
+             log_format=DEFAULT_LOGGING_FORMAT, job_limit=None):
         """Starts the work loop.
 
         Pops and performs all jobs on the current list of queues.  When all
@@ -446,7 +446,7 @@ class Worker(object):
         """
         setup_loghandlers(logging_level, date_format, log_format)
         self._install_signal_handlers()
-        did_perform_work = False
+        completed_jobs = 0
         self.register_birth()
         self.log.info("RQ worker %r started, version %s", self.key, VERSION)
         self.set_state(WorkerStatus.STARTED)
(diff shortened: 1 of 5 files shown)


selwin:
Overall, this PR looks good. Could you please rename the CLI argument to `--max-jobs` and `max_jobs` as the worker argument?

selwin:
One more thing, please document this argument here: https://github.com/rq/rq/blob/master/docs/docs/workers.md

perobertson:
@selwin all fixed up. I had debated calling it `max_jobs` at the start so thanks for pointing that out.

rq/worker.py
```diff
@@ -477,7 +477,11 @@ def work(self, burst=False, logging_level="INFO", date_format=DEFAULT_LOGGING_DA
                     self.execute_job(job, queue)
                     self.heartbeat()
 
-                    did_perform_work = True
+                    completed_jobs += 1
+                    if max_jobs is not None:
+                        if completed_jobs >= max_jobs:
+                            self.log.info("RQ worker %r done, quitting", self.key)
```
selwin: Could you please change the message to `Worker %s: finished executing <MAX_JOBS> jobs, quitting...` to make it more consistent with the other messages?

perobertson: Yup, I can make that change. However it looks like there is a bit of inconsistency already. 

These 2 messages use `RQ Worker` with the prefix before the name.
https://github.com/rq/rq/blob/2baef02dbd9d1522cd9653f0086e4162d502ef29/rq/worker.py#L451
https://github.com/rq/rq/blob/2baef02dbd9d1522cd9653f0086e4162d502ef29/rq/worker.py#L473

Then there is this message for the exceptional case.
https://github.com/rq/rq/blob/2baef02dbd9d1522cd9653f0086e4162d502ef29/rq/worker.py#L490-L493

What do you think about using:
```suggestion
                            self.log.info("RQ Worker %r finished executing %d jobs, quitting", self.key, completed_jobs)
```

And changing the exception message to:
```python
self.log.error(
    'RQ Worker %r: found an unhandled exception, quitting...',
    self.key, exc_info=True
)
```

perobertson: Ive made that change in the meantime.

selwin:
Thanks!

selwin:
Thanks!

eriktelepovsky:
Hello. Does the worker automatically restart itself when it quits after max_jobs value is fulfilled?

PR #1095, opened 2019-06-03
CI - test against py3.8

perobertson:
py3.7 has been out for a bit. It makes sense to have the tests start checking out if 3.8 will work.


selwin:
Thanks!

PR #1096, opened 2019-06-03
Fixes disable-job-desc-logging flag on CLI

janierdavila:
PR #991 introduced this useful flag but it seems to have missed
passing this flag to the worker. This simple commit fixes that
issue.


selwin:
Thanks!

PR #1108, opened 2019-06-16
Replace procname with setproctitle

### LINKED ISSUE #1106: Replace procname with setproctitle
The current rq implementation uses [procname](https://github.com/lericson/procname) for better `ps -ef` output, but the procname doesn't support Python3 actually, according to [this](https://github.com/lericson/procname/issues/3). 
When importing `procname` under python3, it throws `ImportError`: 
```
>>> import procname
  File "<stdin>", line 1, in <module>
ImportError: /home/johnj/.pyenv/versions/general/lib/python3.7/site-packages/procname.cpython-37m-x86_64-linux-gnu.so: undefined symbol: Py_InitModule3
```

We should use [setproctitle](https://github.com/dvarrazzo/py-setproctitle) instead.

j178:
Fix #1106

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -43,7 +43,7 @@ from .version import VERSION
 from .worker_registration import clean_worker_registry, get_keys
 
 try:
-    from procname import setprocname
+    from setproctitle import setproctitle as setprocname
 except ImportError:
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

PR #1110, opened 2019-06-16
Added checks for 0 ttl

selwin:


rq/queue.py
```diff
@@ -231,10 +231,19 @@ def enqueue_call(self, func, args=None, kwargs=None, timeout=None,
...
+            timeout = self._default_timeout
+        elif timeout == 0:
+            raise ValueError('0 timeout is not allowed. Use -1 for infinite timeout')
+
         result_ttl = parse_timeout(result_ttl)
         failure_ttl = parse_timeout(failure_ttl)
+
         ttl = parse_timeout(ttl)
+        if ttl is not None and ttl <= 0:
+            raise ValueError('Job ttl must be greater than 0')
```
mscienski: Other than this parameter check, was the job updated to no longer accept less than 0 ttl?

The docs still reflect that a negative ttl will keep the job on the queue indefinitely:

https://python-rq.org/docs/
> ttl specifies the maximum queued time of the job before it’s discarded. If you specify a value of -1 you indicate an infinite job ttl and it will run indefinitely

PR #1111, opened 2019-06-18
Fix #469 - shutdown worker if it has become orphaned

### LINKED ISSUE #469: heartbeat after key expires
The solution to #138 seems to have created another issue that only occurs in erroneous circumstances. We were running RQ dashboard on a VM and noted that the worker list would disappear randomly even though the workers were still there and still performing jobs on the queue. I was able to isolate the cause to the expiration of the redis key being used in heartbeats and there being no logic to recreate the key once that happens. 

My workaround is to simply register the birth and set the state, like this:

```
def heartbeat(self, timeout=0):
    timeout = max(timeout, self.default_worker_ttl)
    if not self.connection.expire(self.key, timeout):
        # re-register birth if it finds the heartbeat entry missing
        self.register_birth()
        self.set_state(self._state)
    self.log.debug('Sent heartbeat to prevent worker timeout. '
                   'Next one should arrive within {0} seconds.'.format(timeout))
```

stevenkampen:
This forces a worker to quit if the corresponding worker key has been removed or expired. Otherwise the worker continues to execute jobs but in a broken state (e.g. worker key is empty save for `last_heartbeat`).

Tests do not pass at this stage. Haven't looked closely.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -29,7 +29,7 @@ from .connections import get_current_connection, push_connection, pop_connection
 from .defaults import (DEFAULT_RESULT_TTL,
                        DEFAULT_WORKER_TTL, DEFAULT_JOB_MONITORING_INTERVAL,
                        DEFAULT_LOGGING_FORMAT, DEFAULT_LOGGING_DATE_FORMAT)
-from .exceptions import DequeueTimeout, ShutDownImminentException
+from .exceptions import DequeueTimeout, ShutDownImminentException, OrphanedWorkerError
 from .job import Job, JobStatus
 from .logutils import setup_loghandlers
 from .queue import Queue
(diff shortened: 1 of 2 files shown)


selwin:
I agree with the general approach. However, existing tests would need to pass, and new tests added before I can merge this in.

rq/worker.py
```diff
@@ -550,7 +557,9 @@ def heartbeat(self, timeout=None, pipeline=None):
         timeout = timeout or self.default_worker_ttl
         connection = pipeline if pipeline is not None else self.connection
         connection.expire(self.key, timeout)
-        connection.hset(self.key, 'last_heartbeat', utcformat(utcnow()))
+        result = connection.hset(self.key, 'last_heartbeat', utcformat(utcnow()))
```
selwin: This is most probably sent within a Redis pipeline so you'll need to check the result after `pipeline.execute()` is called.

iamxiaozhen:
the commit is merged to branch master or not?

PR #1125, opened 2019-08-05
Store worker's RQ and Python versions

eoranged:
Mostly for displaying purposes (in RQ dashboard).
Also allows to detect partial upgrades when some workers are running old RQ/Python version.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -161,7 +161,7 @@ class Worker(object):
                  queue_class=None, log_job_description=True,
                  job_monitoring_interval=DEFAULT_JOB_MONITORING_INTERVAL,
                  disable_default_exception_handler=False,
-                 prepare_for_work=True):  # noqa
+                 prepare_for_work=True, version=VERSION):  # noqa
         if connection is None:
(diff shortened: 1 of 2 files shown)


selwin:
These info are useful to have, but I don't think we should pass RQ's version through `__init__`. Can you modify this PR to import RQ's version number from `version.py`?

eoranged:
@selwin Unfortunately this won't work, because this way the version will be overwritten on every read.
E.g.: worker uses version 1.0, and we are reading worker object from 2.0. In this case worker object will be initialized with value 2.0, which is incorrect.

selwin:
We can make it so that it always sets `version` to `VERSION` on `__init__` (which it defaults to anyway) and it gets overwritten on `worker.refresh()`.

eoranged:
> We can make it so that it always sets `version` to `VERSION` on `__init__` (which it defaults to anyway) and it gets overwritten on `worker.refresh()`.

Ok It makes sense. Updated.

selwin:
Thanks!

PR #1126, opened 2019-08-05
Fix unreliable test

eoranged:
Also make error message more useful in case of future failures.

Stumbled upon this during preparation of previous PR, but was unable to reproduce It. Adjusted limits a little bit.


selwin:
Thanks!

PR #1130, opened 2019-08-14
Add failure_ttl on job decorator

pax0r:
Without it you cannot specify custom timeout for failed jobs created using decorator


selwin:
Thanks!

PR #1136, opened 2019-09-18
Edit jobs doc for clarity

joncros:
Edits to the "jobs" page of the documentation to make things clearer for new users. In particular, when I first read the docs I saw a job is retrieved using the job id. But it wasn't immediately clear just from this page where one gets the job id. The page seemed to suggest using get_current_job(), but in fact that only works from within the job function.

Some of this does duplicate information from index.md in the interest of lessening the amount of jumping from page to page needed to get the concepts.


selwin:
This is an excellent addition to the docs. Thanks!

PR #1140, opened 2019-09-25
add doc page for job registries

joncros:
Adds a page documenting the types of job registries. I did debate whether to insert this on an existing page such as the Queues page. The Queues page has gotten fairly large, though. Let me know, @selwin, if you have a preference.

### DIFF
--- a/docs/docs/job_registries.md
+++ b/docs/docs/job_registries.md
@@ -27,8 +27,8 @@ job = queue.enqueue(count_words_at_url, 'http://nvie.com')
 # get StartedJobRegistry by queue
 registry = StartedJobRegistry(queue=queue)
 
-# or get StartedJobRegistry by connection
-registry = StartedJobRegistry(connection=conn)
+# or get StartedJobRegistry by queue name and connection
+registry2 = StartedJobRegistry(name='my_queue', connection=conn)
 
 # sleep for a moment while job is taken off the queue
 time.sleep(0.1)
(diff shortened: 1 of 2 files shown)


docs/docs/job_registries.md
```diff
@@ -0,0 +1,45 @@
+---
+title: "RQ: Job Registries"
+layout: docs
+---
+
+Each queue maintains a set of Job Registries:
+* `StartedJobRegistry` Holds currently executing jobs. Jobs are added right before they are
```
selwin: Nitpick: can we make the language here more consistent? Each point should start with `holds ...`

docs/docs/job_registries.md
```diff
@@ -0,0 +1,45 @@
...
+
+conn = Redis()
+queue = Queue(connection=conn)
+job = queue.enqueue(count_words_at_url, 'http://nvie.com')
+
+# get StartedJobRegistry by queue
+registry = StartedJobRegistry(queue=queue)
+
+# or get StartedJobRegistry by connection
+registry = StartedJobRegistry(connection=conn)
```
selwin: We should also add `name='queue_name'`. Without the `name` argument, it looks like `StartedJobRegistry` is an object that keeps track of all started jobs in this Redis connection.

docs/docs/job_registries.md
joncros: Sure, I've added a commit to do this.

docs/docs/job_registries.md
joncros: Got it. I've pushed a commit to address this.

selwin:
Thanks!

Issue #369, opened 2014-06-26
Customisable args/kwargs serializer


selwin:
It would be nice to have an option to use JSON in addition to the default `pickle` serializer for performance and security reasons,

f0cker:
Is this in progress anywhere? This is vital imho, without it any applications built using RQ are likely to be vulnerable to remote code execution.

rpkak:
Should this issue still be open? I think it is done in #1219 and #1381.

selwin:
Yep, this is already implemented.

PR #1141, opened 2019-09-28
rq: Configurable serializer support

### LINKED ISSUE #973: More versatile pickling with dill
[dill][d] extend's Python's pickling capabilities to handle things like lambdas and instance methods.

It would be nice if RQ were able to use this, if it's available on the system.  As far as I know, it's a drop-in replacement for cPickle.

[d]: https://github.com/uqfoundation/dill

woodruffw:
This is a WIP approach to configurable serializer support.

Instead of making the serializer an attribute of every class/instance that might potentially need to use it, this introduces `utils.serializer` and `utils.configure_serializer`. During early initialization, applications can do something like this:

```python
import rq
import dill

rq.configure_serializer(dill)
```

A "serializer" is any object that has at least `dumps` and `loads`.

This also introduces a more generic exception, `DeserializationError`, and aliases the old `UnpickleError` to it for backwards compatibility.

I'd appreciate any and all feedback on this approach!

Closes #973.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -395,7 +395,7 @@ class Job(object):
             rv = self.connection.hget(self.key, 'result')
             if rv is not None:
                 # cache the result
-                self._result = loads(rv)
+                self._result = serializer.loads(rv)
         return self._result
 
     """Backwards-compatibility accessor property `return_value`."""
--- a/rq/job.py
+++ b/rq/job.py
@@ -5,27 +5,16 @@ from __future__ import (absolute_import, division, print_function,
 import inspect
 import warnings
 import zlib
-from functools import partial
 from uuid import uuid4
 
 from rq.compat import as_text, decode_redis_hash, string_types, text_type
 
 from .connections import resolve_connection
-from .exceptions import NoSuchJobError, UnpickleError
+from .exceptions import DeserializationError, NoSuchJobError
 from .local import LocalStack
-from .utils import (enum, import_attribute, parse_timeout, str_to_date,
-                    utcformat, utcnow)
+from .utils import (enum, serializer, import_attribute, parse_timeout,
+                    str_to_date, utcformat, utcnow)
 
-try:
-    import cPickle as pickle
-except ImportError:  # noqa  # pragma: no cover
-    import pickle
-
-
-# Serialize pickle dumps using the highest pickle protocol (binary, default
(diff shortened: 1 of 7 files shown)


rq/job.py
```diff
@@ -41,18 +30,18 @@
 UNEVALUATED = object()
 
 
-def unpickle(pickled_string):
-    """Unpickles a string, but raises a unified UnpickleError in case anything
-    fails.
+def deserialize(serialized):
```
selwin: Nitpick: `serialized_string`

selwin:
In general, I like this PR, but I'm not a fan of using a global `serializer` object. Could you please change this PR such that the serializer object has to be passed in to `Queue` and `Worker`'s init methods?
```python
Worker(connection=connection, serializer=serializer)
Queue(connection=connection, serializer=serializer)
```

`serializer` argument should default to `pickle_serializer`.

woodruffw:
Thanks!

I agree re: the global; that was my original approach because I wasn't sure how best to trickle the serializer into e.g. jobs created via `Job.create`. But I should be able to do it via an attribute on the queue, so that's fine.

I'll make the changes in a bit!

woodruffw:
@selwin Any thoughts on how to handle class methods like `Queue.from_queue_key`?

We won't have access to a serializer instance in that context, so the retrieved `Queue` will be incorrectly reconstructed with the default serializer, even if originally specified with `dill` or another.

selwin:
Sorry for the late reply, I think `from_queue_key` should also accept a `serializer` argument.

selwin:
As an added note, you should probably do your work against the [job-scheduling branch](https://github.com/rq/rq/tree/job-scheduling), this branch contains quite a few changes and I plan on merging this into master soon (I hope).

nzjrs:
This would also address #369 I belive

woodruffw:
> This would also address #369 I belive

I think so, yes. Sorry for the lack of activity here -- I haven't had much free time to work on this.

Anybody who feels like taking it over should do so -- I probably won't be able to continue it any time in the next month or so.

selwin:
No worries 👍🏻

selwin:
Fixed in https://github.com/rq/rq/pull/1219

woodruffw:
Woohoo!

PR #1145, opened 2019-10-16
Fix dependent jobs enqueueing

van-ess0:
If Queue was subclassed and enqueue_job changed, these changes should affect on dependent jobs' enqueueing too.

```
class MyQueue(Queue):
    def enqueue_job(self, *args, **kwargs):
        logger.info('Job was enqueued')
        return super().enqueue_job(*args, **kwargs)
```


selwin:
Thanks!

PR #279, opened 2013-10-25
Support for multiple job dependencies

selwin:
I started with @jchia's awesome [initial pull request](https://github.com/nvie/rq/pull/277) and reworked some parts to make it more readable (hopefully).

As far as I'm concerned, this PR is mostly complete except for some pipelining that we need to do for efficiency reasons.

Would be great if @nvie and @jchia can do another quick review on this PR so we can get this merged and get 0.4 out the door.

PR #1147, opened 2019-10-20
Multi Dependency Support [Internal API Changes]

thomasmatecki:
Hi!

After reviewing issues & PRs, it seems dependence upon multiple jobs is not implemented, still on the table, and perhaps even useful to many. If I've misinterpreted, please forgive me :bowing_man: .

In an attempt to make some headway toward _multiple dependence_, I've tried to address https://github.com/rq/rq/pull/1090#issuecomment-497528595.

The intent of this PR is simply to lay groundwork, avoiding any API or functional changes.

For reference: https://github.com/rq/rq/pull/856


thomasmatecki:
oh my. I have just noticed #279. @selwin @nvie - is this work likely to be reopened? I'm happy to pull on this thread for a bit to try to get in completed (:smiley_cat: though it seems many have attempted this before).

selwin:
This feature is something I'd like to get into RQ, now that RQ's job dependency feature is solid.

This PR looks good, but I think we should create a `_dependency_id` property to keep things backwards compatible for people who use third party packages like django-rq and rq-dashboard.

thomasmatecki:
:+1: @selwin `_dependency_id` added in: d00b6d872ac454b4064f38b0c6debae39472c0ec

selwin:
Thanks!

PR #1155, opened 2019-10-28
Multi Dependency Support - Registration & Enqueue Call

thomasmatecki:
Internal API changes to support multiple dependencies.

Successor to  #1147.

* Store all of a job's _dependencies_ in a redis set. Delete that set when a job is deleted.
* Add `Job#fetch_dependencies` method - which return all jobs a job is dependent upon and optionally _WATCHES_ all dependency ids.
* Use `Job#fetch_dependencies` in `Queue#call_enqueue`. `fetch_dependencies` now sets WATCH and raises `InvalidJobDependency`, rather than `call_enqueue`.

`Queue` and `Job` public APIs still expect single ids of jobs for `depends_on` but internally register them in a way that could support multiple jobs being passed as dependencies.

Next up: updates to `Queue#enqueue_dependents`.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -411,11 +411,10 @@ class Job(object):
 
         Returned jobs will use self's connection, not the pipeline supplied.
         """
-        if pipeline is None:
-            pipeline = pipeline or self.connection
+        connection = pipeline if pipeline is not None else self.connection
 
         if watch and self._dependency_ids:
-            pipeline.watch(*self._dependency_ids)
+            connection.watch(*self._dependency_ids)
 
         jobs = self.fetch_many(self._dependency_ids, connection=self.connection)
 
@@ -700,6 +699,8 @@ class Job(object):
         elif ttl > 0:
             connection = pipeline if pipeline is not None else self.connection
             connection.expire(self.key, ttl)
+            connection.expire(self.dependents_key, ttl)
+            connection.expire(self.dependencies_key, ttl)
 
     @property
     def failed_job_registry(self):
--- a/rq/job.py
+++ b/rq/job.py
@@ -400,7 +400,7 @@ class Job(object):
 
     @property
     def dependencies_key(self):
-        return '{0}{1}:dependencies'.format(self.redis_job_namespace_prefix, self.id)
+        return '{0}:{1}:dependencies'.format(self.redis_job_namespace_prefix, self.id)
 
     def fetch_dependencies(self, watch=False, pipeline=None):
         """
@@ -415,10 +415,7 @@ class Job(object):
         if watch and self._dependency_ids:
             pipeline.watch(*self._dependency_ids)
 
-        jobs = self.fetch_many(
-            self._dependency_ids,
-            connection=self.connection
-        )
+        jobs = self.fetch_many(self._dependency_ids, connection=self.connection)
 
         for i, job in enumerate(jobs):
             if not job:
@@ -625,9 +622,7 @@ class Job(object):
         if delete_dependents:
             self.delete_dependents(pipeline=pipeline)
 
-        connection.delete(self.key)
-        connection.delete(self.dependents_key)
-        connection.delete(self.dependencies_key)
+        connection.delete(self.key, self.dependents_key, self.dependencies_key)
 
     def delete_dependents(self, pipeline=None):
         """Delete jobs depending on this job."""
--- a/rq/job.py
+++ b/rq/job.py
@@ -140,8 +140,10 @@ class Job(object):
             job._dependency_ids = [depends_on.id if isinstance(depends_on, Job) else depends_on]
         return job
 
-    def get_status(self):
-        self._status = as_text(self.connection.hget(self.key, 'status'))
+    def get_status(self, refresh=True):
+        if refresh:
+            self._status = as_text(self.connection.hget(self.key, 'status'))
+
         return self._status
 
     def set_status(self, status, pipeline=None):
--- a/rq/job.py
+++ b/rq/job.py
@@ -298,7 +298,12 @@ class Job(object):
 
     @classmethod
     def fetch_many(cls, job_ids, connection=None):
-        """Bulk version of Job.fetch"""
+        """
+        Bulk version of Job.fetch
+
+        For any job_ids which a job does not exist, the corresponding item in
+        the returned list will be None.
+        """
         with connection.pipeline() as pipeline:
             for job_id in job_ids:
                 pipeline.hgetall(cls.key_for(job_id))
@@ -400,9 +405,9 @@ class Job(object):
     def fetch_dependencies(self, watch=False, pipeline=None):
         """
         Fetch all of a job's dependencies. If a pipeline is supplied, and
-        watch is true, the set WATCH on all the keys of all dependencies.
+        watch is true, then set WATCH on all the keys of all dependencies.
 
-        Returned jobs will NOT use the pipeline supplied.
+        Returned jobs will use self's connection, not the pipeline supplied.
         """
         if pipeline is None:
             pipeline = pipeline or self.connection
@@ -410,21 +415,15 @@ class Job(object):
         if watch and self._dependency_ids:
             pipeline.watch(*self._dependency_ids)
 
-        results = [
-            pipeline.hgetall(self.key_for(job_id))
-            for job_id in self._dependency_ids
-        ]
+        jobs = self.fetch_many(
+            self._dependency_ids,
+            connection=self.connection
+        )
 
-        jobs = []
-        for i, job_id in enumerate(self._dependency_ids):
-            if results[i]:
-                job = self.__class__(job_id)
-                job.restore(results[i])
-                jobs.append(job)
-            else:
-                # If the dependency does not exist, raise an
-                # exception to avoid creating an orphaned job.
-                raise InvalidJobDependency('Job {0} does not exist'.format(job_id))
+        for i, job in enumerate(jobs):
+            if not job:
+                raise InvalidJobDependency('Job {0} does not exist'.format(self._dependency_ids[i]))
+
         return jobs
 
 
--- a/rq/job.py
+++ b/rq/job.py
@@ -297,7 +297,7 @@ class Job(object):
         return job
 
     @classmethod
-    def fetch_many(cls, job_ids, connection=None):
+    def fetch_many(cls, job_ids, connection):
         """
         Bulk version of Job.fetch
 
--- a/rq/job.py
+++ b/rq/job.py
@@ -419,7 +419,7 @@ class Job(object):
 
         for i, job in enumerate(jobs):
             if not job:
-                raise InvalidJobDependency('Job {0} does not exist'.format(self._dependency_ids[i]))
+                raise NoSuchJobError('Dependency {0} does not exist'.format(self._dependency_ids[i]))
 
         return jobs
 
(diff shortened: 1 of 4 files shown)


olingerc:
@thomasmatecki A big thank you that you are taking this up again. I've been using my own branch with multi-dependencies for a long time now in production but never found the time to integrate it back into master.

thomasmatecki:
@olingerc 👍 glad to contribute - would love any thoughts on the approach I'm taking.

rq/job.py
```diff
@@ -393,6 +393,41 @@ def dependents_key(self):
...
+        Returned jobs will NOT use the pipeline supplied.
+        """
+        if pipeline is None:
+            pipeline = pipeline or self.connection
+
+        if watch and self._dependency_ids:
+            pipeline.watch(*self._dependency_ids)
+
+        results = [
+            pipeline.hgetall(self.key_for(job_id))
```
selwin: Since we need to evaluate results right away, we shouldn't use `pipeline` here, I think we should use `self.connection`.

Also, I think it's about time we have an operation to fetch jobs in bulk. If you have the time, a separate PR implementing `Job.fetch_many(self.connection, [job_id_1, job_id_2])` would be much appreciated. Otherwise, it's also ok :)

thomasmatecki: @selwin it looks as though `Job.fetch_many` already exists :smiley: 

https://github.com/rq/rq/blob/dfc753332489af85ea2f9c9188771f8a2d61e400/rq/job.py#L300

Good thinking re: `self.connection` vs. `pipeline` - passing `self`'s connection to `fetch_many` in dfc753332489af85ea2f9c9188771f8a2d61e400 seems to do the trick.

One other thing - the `fetch_many` method used a default value of `None` for connection, but  will always raise an `AttributeError` of `NoneType` is _something_ is not passed. I've removed this default value in d366b788ecef38709507b4730cc516eab437adfb because it seems more correct to enforce the necessity of `connection` in the method's signature.

selwin:
@thomasmatecki I'll try to take a look at the revised PR this weekend.

rq/job.py
```diff
@@ -393,6 +398,35 @@ def dependents_key(self):
         """The Redis key that is used to store job dependents hash under."""
         return self.dependents_key_for(self.id)
 
+    @property
+    def dependencies_key(self):
+        return '{0}{1}:dependencies'.format(self.redis_job_namespace_prefix, self.id)
```
selwin: The key should be `{0}:{1}:dependencies`

rq/job.py
```diff
@@ -393,6 +398,35 @@ def dependents_key(self):
...
+
+        Returned jobs will use self's connection, not the pipeline supplied.
+        """
+        if pipeline is None:
+            pipeline = pipeline or self.connection
+
+        if watch and self._dependency_ids:
+            pipeline.watch(*self._dependency_ids)
+
+        jobs = self.fetch_many(
```
selwin: Nitpick: these three lines is short enough, can you please make this into one liner?

rq/job.py
```diff
@@ -593,6 +627,7 @@ def delete(self, pipeline=None, remove_from_queue=True,
 
         connection.delete(self.key)
```
selwin: Can you change this into:
```python
connection.delete(self.key, self.dependents_key, self.dependencies_key)
```

tests/test_job.py
```diff
@@ -665,3 +694,82 @@ def test_key_for_should_return_prefixed_job_id(self):
...
+
+        dependent_job.register_dependency()
+        dependent_job.save()
+
+        dependency_job.delete()
+
+        with self.assertRaises(InvalidJobDependency):
+            dependent_job.fetch_dependencies(pipeline=self.testconn)
+
+    def test_fetch_dependencies_raises_if_dependency_deleted(self):
```
selwin: It seems like this testcase is a duplicate of the [previous one](https://github.com/rq/rq/pull/1155/files#diff-5cbe52c4c7165713f09d3ae38168277eR729)

rq/job.py
thomasmatecki: In 7c690087a9f4d287d400e8b489eb851e94fedfc0"
https://github.com/rq/rq/blob/7c690087a9f4d287d400e8b489eb851e94fedfc0/rq/job.py#L625

tests/test_job.py
thomasmatecki: :+1: removed

rq/job.py
thomasmatecki: In 7c690087a9f4d287d400e8b489eb851e94fedfc0:
https://github.com/rq/rq/blob/7c690087a9f4d287d400e8b489eb851e94fedfc0/rq/job.py#L418

rq/job.py
thomasmatecki: In  7c690087a9f4d287d400e8b489eb851e94fedfc0:

https://github.com/rq/rq/blob/7c690087a9f4d287d400e8b489eb851e94fedfc0/rq/job.py#L403

rq/job.py
```diff
@@ -393,6 +398,32 @@ def dependents_key(self):
...
+            pipeline = pipeline or self.connection
+
+        if watch and self._dependency_ids:
+            pipeline.watch(*self._dependency_ids)
+
+        jobs = self.fetch_many(self._dependency_ids, connection=self.connection)
+
+        for i, job in enumerate(jobs):
+            if not job:
+                raise InvalidJobDependency('Job {0} does not exist'.format(self._dependency_ids[i]))
```
selwin: One last comment, I'm just wondering whether we should raise a `NoSuchJobError('Dependency {0} does not exist')` instead of `InvalidJobDependency`. I could go either way. Thoughts?

rq/queue.py
```diff
@@ -252,33 +252,35 @@ def enqueue_call(self, func, args=None, kwargs=None, timeout=None,
...
+                            watch=True,
+                            pipeline=pipe
+                        )
+
+                        pipe.watch(job.dependencies_key)
+
+                        pipe.multi()
+
+                        for dependency in dependencies:
+                            if dependency.get_status() != JobStatus.FINISHED:
```
selwin: `dependency.get_status()` makes a round trip to Redis just to get the job's status. We should check `dependency._status` instead.

rq/queue.py
```diff
@@ -252,33 +252,35 @@ def enqueue_call(self, func, args=None, kwargs=None, timeout=None,
...
+                                job.set_status(JobStatus.DEFERRED)
```
selwin: This should be `job.set_status(JobStatus.DEFERRED, pipeline=pipe)`

rq/queue.py
```diff
@@ -252,33 +252,35 @@ def enqueue_call(self, func, args=None, kwargs=None, timeout=None,
...
+                                job.register_dependency(pipeline=pipe)
+                                job.save(pipeline=pipe)
+                                job.cleanup(ttl=job.ttl, pipeline=pipe)
+                                pipe.execute()
```
selwin: I think we can call `pipe.execute()` outside of the for loop to write everything all at once.

rq/job.py
thomasmatecki: I think `NoSuchJobError` is the way to go. It seems a bit more consistent with what is raised other places in `job.py` and also seems to more precisely describe _what went wrong_.

Just to call out: In the environments in which _I_ use RQ in production, this problem will surface in logs and alerting - raising an exception named `InvalidJobDependency` as opposed to `NoSuchJobError` isn't going to provide any context I won't already have from a stack trace. To that end, I only see any place _else_ `InvalidJobDependency` is used. Perhaps it could be removed?

rq/queue.py
thomasmatecki: @selwin do you mind taking another glance? The way I read lines 275 - 282 is:

- Iterate over all of `job`'s dependencies
- IF any _one_ dependency is not `FINISHED`, THEN: register dependencies in redis, set the job to `DEFERRED`, `save`, `cleanup` and `return`(_short circuiting_ `enqueue_call`).

Moving it outside side of the loop seems like it would necessitate changing the _short circuit_ structure the method.

rq/queue.py
thomasmatecki: In 6f685b8b01945feea624bab98aa99ed01e467cdf

rq/queue.py
thomasmatecki: in f9b8f92a532035077f89a59f77d6cb79e5eadee9

rq/queue.py
thomasmatecki: In 58edab45ccfb042ae6a61a119901ce2dadddfbdd - I have updated `get_status`, adding a `refresh` parameter: 

https://github.com/rq/rq/blob/58edab45ccfb042ae6a61a119901ce2dadddfbdd/rq/job.py#L143

This feels a little bit like interface bloat, but also let's us avoid exposure of `Job`'s private `_status` attribute:

https://github.com/rq/rq/blob/58edab45ccfb042ae6a61a119901ce2dadddfbdd/rq/queue.py#L277

selwin: `get_status(refresh=True)` is a wonderful idea 👍

rq/job.py
selwin: Ok.

rq/queue.py
selwin: Yes, you're absolutely right. That was my mistake.

rq/job.py
```diff
@@ -591,8 +624,7 @@ def delete(self, pipeline=None, remove_from_queue=True,
         if delete_dependents:
             self.delete_dependents(pipeline=pipeline)
 
-        connection.delete(self.key)
-        connection.delete(self.dependents_key)
+        connection.delete(self.key, self.dependents_key, self.dependencies_key)
```
selwin: `dependencies_key`, along with `dependents_key` should be expired in [job.cleanup()](https://github.com/rq/rq/blob/master/rq/job.py#L655)

rq/job.py
```diff
@@ -393,6 +400,32 @@ def dependents_key(self):
...
+
+    def fetch_dependencies(self, watch=False, pipeline=None):
+        """
+        Fetch all of a job's dependencies. If a pipeline is supplied, and
+        watch is true, then set WATCH on all the keys of all dependencies.
+
+        Returned jobs will use self's connection, not the pipeline supplied.
+        """
+        if pipeline is None:
+            pipeline = pipeline or self.connection
```
selwin: Should this be `pipeline = self.connection`?

rq/job.py
```diff
@@ -393,6 +400,32 @@ def dependents_key(self):
...
+        return '{0}:{1}:dependencies'.format(self.redis_job_namespace_prefix, self.id)
+
+    def fetch_dependencies(self, watch=False, pipeline=None):
+        """
+        Fetch all of a job's dependencies. If a pipeline is supplied, and
+        watch is true, then set WATCH on all the keys of all dependencies.
+
+        Returned jobs will use self's connection, not the pipeline supplied.
+        """
+        if pipeline is None:
```
selwin: I think we should use `connection = pipeline if pipeline is not None else self.connection` convention that's used throughout this library.

rq/job.py
thomasmatecki: :+1: 71a427279f90c5ab566ba741467ad2413994bc75

rq/job.py
thomasmatecki: :+1: 71a427279f90c5ab566ba741467ad2413994bc75

rq/job.py
thomasmatecki: :+1: 71a427279f90c5ab566ba741467ad2413994bc75

selwin:
Thanks!

PR #1161, opened 2019-11-23
Added `delete_job` argument to registry.remove()`

selwin:


selwin:
Will try to publish a release before Christmas. Thanks

Issue #323, opened 2014-03-14
Timeout SIGALRM not enough to interrupt on some kind of worker


V-E-O:
Hi,

My RQ task has some native ext library that do not respect SIGALRM when timeout reached. So non-responding worker hang for a while long beyond the timeout I expect.

The parent process rqworker is still waiting for return from pipeline, but the timeout is reached, that worker is expected to be free not busy. Some weird status happen in RQ dashboard, the worker may be disappearing or waiting for blank queue.

Can we have an enhancement to let the parent rqworker monitor the timeout? If the timeout is reached, try send some signal or just kill the child.

Best Regards,
V.E.O

aroberts:
I'm also hitting this issue when the alarm should fire during a SQLAlchemy commit().  The alarm never fires, and I end up with jobs running way past the timeout.

selwin:
Some libraries or tasks may catch SIGALRM so it never reaches the worker. I think having the parent process monitor the forked process is a decent approach. This way, tasks will still be killed on time even if it misbehaves or becomes unresponsive. What do you think @nvie ?

hhuuggoo:
I've had luck overriding execute_job and replacing the call to waitpid with this function.   I can submit a PR if we like this approach

import os
import threading
import time
import signal

def waitpid_timeout(pid, timeout=5.0):
    result = []
    def helper():
        result.append(os.waitpid(pid, 0))
    st = time.time()
    t = threading.Thread(target=helper)
    t.start()
    t.join(timeout=timeout)
    if t.isAlive():
        os.kill(pid, signal.SIGKILL)
        t.join()
    return result[0]

if __name__ == "__main__":
    import subprocess
    p = subprocess.Popen("python -c 'import time; time.sleep(1)'", shell=True)
    st = time.time()
    print waitpid_timeout(p.pid)
    ed = time.time()
    print ed - st
    p = subprocess.Popen("python -c 'import time; time.sleep(10)'", shell=True)
    st = time.time()
    print waitpid_timeout(p.pid)
    ed = time.time()
    print ed - st

aroberts:
[...]

PR #1169, opened 2019-12-11
Add a hard kill from the parent process with a 10% increased timeout …

mr-trouble:
…in case the forked process gets stuck and cannot stop itself.

#323

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -625,6 +625,8 @@ class Worker(object):
         either executes successfully or the status of the job is set to
         failed
         """
+
+        ret_val = None
         while True:
             try:
                 job.started_at = job.started_at or utcnow()
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -627,6 +627,7 @@ class Worker(object):
         """
         while True:
             try:
+                job.started_at = job.started_at or utcnow()
                 with UnixSignalDeathPenalty(self.job_monitoring_interval, HorseMonitorTimeoutException):
                     retpid, ret_val = os.waitpid(self._horse_pid, 0)
                 break
@@ -634,6 +635,12 @@ class Worker(object):
                 # Horse has not exited yet and is still running.
                 # Send a heartbeat to keep the worker alive.
                 self.heartbeat(self.job_monitoring_interval + 5)
+
+                # Kill the job from this side if something is really wrong (interpreter lock/etc).
+                if (utcnow() - job.started_at).total_seconds() > (job.timeout * 1.1):
+                    self.kill_horse()
+                    break
+
             except OSError as e:
                 # In case we encountered an OSError due to EINTR (which is
                 # caused by a SIGINT or SIGTERM signal during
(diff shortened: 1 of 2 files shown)


rq/worker.py
```diff
@@ -625,15 +625,24 @@ def monitor_work_horse(self, job):
...
                 with UnixSignalDeathPenalty(self.job_monitoring_interval, HorseMonitorTimeoutException):
                     retpid, ret_val = os.waitpid(self._horse_pid, 0)
                 break
             except HorseMonitorTimeoutException:
                 # Horse has not exited yet and is still running.
                 # Send a heartbeat to keep the worker alive.
                 self.heartbeat(self.job_monitoring_interval + 5)
+
+                # Kill the job from this side if something is really wrong (interpreter lock/etc).
+                if (utcnow() - job.started_at).total_seconds() > (job.timeout * 1.1):
```
selwin: Instead of adding a 10% time multiplier, I think it's safe to add just 1 or 2 seconds to the timeout. The default job timeout is 180 so having a 10% tolerance would be 18 seconds, which is way too long.

rq/worker.py
```diff
@@ -625,15 +625,24 @@ def monitor_work_horse(self, job):
         either executes successfully or the status of the job is set to
         failed
         """
+
+        ret_val = None
         while True:
             try:
+                job.started_at = job.started_at or utcnow()
```
selwin: `job.started_at` should be set outside of the `while` loop, correct? Otherwise `job.started_at` will keep increasing everytime the parent checks whether horse is still alive.

rq/worker.py
mr-trouble: Sounds good. I'll retest with 1 or 2 seconds.

rq/worker.py
mr-trouble: I'll test this, but I think that the `or` will only fire the `utcnow()` method once because `job.started_at` is `None`. After this, its a useless assignment. I'll move it outside the loop anyways because its better form. Thanks for the review.

tests/test_worker.py
```diff
@@ -1062,6 +1062,34 @@ def test_work_horse_death_sets_job_failed(self):
...
+        w.fork_work_horse(job, queue)
+        job.timeout = 5
+        w.job_monitoring_interval = 1
+        now = utcnow()
+        w.monitor_work_horse(job)
+        job_status = job.get_status()
+        fudge_factor = 1
+        total_time = w.job_monitoring_interval + 5 + fudge_factor
+        self.assertTrue((utcnow() - now).total_seconds() < total_time)
+        self.assertEqual(job_status, JobStatus.FAILED)
```
selwin: Nitpick: `self.assertEqual(job.get_status(), JobStatus.FAILED)`

selwin:
I plan on releasing a new RQ version before this weekend. It would be great if you can get this PR in shape by then so it can be part of the next release.

mr-trouble:
> I plan on releasing a new RQ version before this weekend. It would be great if you can get this PR in shape by then so it can be part of the next release.

Sorry, was vacationing for the last 20 days or so. Looks like I've missed the release, but I've updated according to your review. Happy new year!

selwin:
Thanks!

PR #1170, opened 2019-12-15
[Multi-Dependencies] Enqueuing of Dependents

thomasmatecki:
successor to #1155 

This PR updates `Queue#enqueue_dependents` such that only those _dependents_ for which all _dependencies_ are complete, will be enqueued. It includes:

- New method on `Job#get_dependencies_statuses` to retrieve the status of all of a jobs _dependencies_.
- A modification to `Queue#enqueue_dependents` (as described above).
- A modification to `Worker#handle_job_success` so that the status of **ALL** successful jobs is persisted to `FINISHED` in Redis.
- A small fix(?) to `Job#set_status` so that it uses a pipeline when appropriate.

Notwithstanding any bugs, I believe the next step here is to modify `Job` such that it accepts more than one dependency(in `depends_on`.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -9,6 +9,7 @@ from functools import partial
 from uuid import uuid4
 
 from rq.compat import as_text, decode_redis_hash, string_types, text_type
+
 from .connections import resolve_connection
 from .exceptions import NoSuchJobError
 from .local import LocalStack
@@ -740,12 +741,7 @@ class Job(object):
         return [Job.key_for(as_text(_id))
                 for _id in dependencies]
 
-    def dependencies_are_met(
-        self,
-        pipeline=None,
-        exclude=None
-
-    ):
+    def dependencies_are_met(self, exclude_job_id=None, pipeline=None):
         """Returns a boolean indicating if all of this jobs dependencies are _FINISHED_
 
         If a pipeline is passed, all dependencies are WATCHed.
@@ -755,7 +751,6 @@ class Job(object):
         `FINISHED` may not be yet set in redis, but said job is indeed _done_ and this
         method is _called_ in the _stack_ of it's dependents are being enqueued.
         """
-        exclude = exclude or []
 
         pipe = pipeline if pipeline is not None else self.connection
 
@@ -786,6 +781,6 @@ class Job(object):
         return all(status == JobStatus.FINISHED or not created_at
                    for dependency_id, created_at, status
                    in dependencies_statuses
-                   if dependency_id not in exclude)
+                   if not (exclude_job_id and dependency_id == exclude_job_id))
 
 _job_stack = LocalStack()
--- a/rq/job.py
+++ b/rq/job.py
@@ -763,11 +763,11 @@ class Job(object):
         if exclude_job_id:
             dependencies_ids.discard(exclude_job_id)
 
-        dependencies_statuses = [
-            connection.hget(self.key_for(key), 'status')
-            for key
-            in dependencies_ids
-        ]
+        with connection.pipeline() as pipeline:
+            for key in dependencies_ids:
+                pipeline.hget(self.key_for(key), 'status')
+
+            dependencies_statuses = pipeline.execute()
 
         return all(
             status.decode() == JobStatus.FINISHED
--- a/rq/job.py
+++ b/rq/job.py
@@ -744,9 +744,8 @@ class Job(object):
         pipe = pipeline if pipeline is not None else self.connection
 
         if watch:
-            pipe.watch(self.dependencies_key)
-            pipe.watch(*[self.redis_job_namespace_prefix + as_text(_id)
-                        for _id in pipe.smembers(self.dependencies_key)])
+            pipe.watch(*[Job.key_for(_id)
+                         for _id in self.connection.smembers(self.dependencies_key)])
 
         sort_by = self.redis_job_namespace_prefix + '*->ended_at'
         get_field = self.redis_job_namespace_prefix + '*->status'
@@ -761,6 +760,5 @@ class Job(object):
         ]
 
         return dependencies_statuses
-
 
 _job_stack = LocalStack()
--- a/rq/job.py
+++ b/rq/job.py
@@ -724,7 +724,7 @@ class Job(object):
             connection.sadd(dependents_key, self.id)
             connection.sadd(self.dependencies_key, dependency_id)
 
-    def dependencies_finished(
+    def dependencies_are_met(
             self,
             pipeline=None
     ):
--- a/rq/job.py
+++ b/rq/job.py
@@ -21,6 +21,7 @@ try:
 except ImportError:  # noqa  # pragma: no cover
     import pickle
 
+
 # Serialize pickle dumps using the highest pickle protocol (binary, default
 # uses ascii)
 dumps = partial(pickle.dumps, protocol=pickle.HIGHEST_PROTOCOL)
@@ -103,13 +104,11 @@ class Job(object):
             job._func_name = '{0}.{1}'.format(func.__module__, func.__name__)
         elif isinstance(func, string_types):
             job._func_name = as_text(func)
-        elif not inspect.isclass(func) and hasattr(func,
-                                                   '__call__'):  # a callable class instance
+        elif not inspect.isclass(func) and hasattr(func, '__call__'):  # a callable class instance
             job._instance = func
             job._func_name = '__call__'
         else:
-            raise TypeError(
-                'Expected a callable or a string, but got: {0}'.format(func))
+            raise TypeError('Expected a callable or a string, but got: {0}'.format(func))
         job._args = args
         job._kwargs = kwargs
 
@@ -124,8 +123,7 @@ class Job(object):
 
         # dependency could be job instance or id
         if depends_on is not None:
-            job._dependency_ids = [
-                depends_on.id if isinstance(depends_on, Job) else depends_on]
+            job._dependency_ids = [depends_on.id if isinstance(depends_on, Job) else depends_on]
         return job
 
     def get_status(self, refresh=True):
@@ -411,8 +409,8 @@ class Job(object):
         if watch and self._dependency_ids:
             connection.watch(*self._dependency_ids)
 
-        jobs = [job for
-                job in self.fetch_many(self._dependency_ids, connection=self.connection)
+        jobs = [job
+                for job in self.fetch_many(self._dependency_ids, connection=self.connection)
                 if job]
 
         return jobs
@@ -471,10 +469,8 @@ class Job(object):
             except Exception as e:
                 self._result = "Unserializable return value"
         self.timeout = parse_timeout(obj.get('timeout')) if obj.get('timeout') else None
-        self.result_ttl = int(obj.get('result_ttl')) if obj.get(
-            'result_ttl') else None  # noqa
-        self.failure_ttl = int(obj.get('failure_ttl')) if obj.get(
-            'failure_ttl') else None  # noqa
+        self.result_ttl = int(obj.get('result_ttl')) if obj.get('result_ttl') else None  # noqa
+        self.failure_ttl = int(obj.get('failure_ttl')) if obj.get('failure_ttl') else None  # noqa
         self._status = as_text(obj.get('status')) if obj.get('status') else None
 
         dependency_id = obj.get('dependency_id', None)
@@ -791,6 +787,5 @@ class Job(object):
                    for dependency_id, created_at, status
                    in dependencies_statuses
                    if dependency_id not in exclude)
-
 
 _job_stack = LocalStack()
--- a/rq/job.py
+++ b/rq/job.py
@@ -4,12 +4,12 @@ from __future__ import (absolute_import, division, print_function,
 
 import inspect
 import warnings
-import zlib
 from functools import partial
 from uuid import uuid4
 
-from rq.compat import as_text, decode_redis_hash, string_types, text_type
+import zlib
 
+from rq.compat import as_text, decode_redis_hash, string_types, text_type
 from .connections import resolve_connection
 from .exceptions import NoSuchJobError
 from .local import LocalStack
@@ -752,35 +752,28 @@ class Job(object):
         method is _called_ in the _stack_ of it's dependents are being enqueued.
         """
 
-        pipe = pipeline if pipeline is not None else self.connection
+        connection = pipeline if pipeline is not None else self.connection
 
         if pipeline is not None:
-            pipe.watch(*self.dependency_ids)
+            connection.watch(*self.dependency_ids)
 
-        sort_by = self.redis_job_namespace_prefix + '*->ended_at'
-        get_fields = (
-            '#',
-            self.redis_job_namespace_prefix + '*->created_at',
-            self.redis_job_namespace_prefix + '*->status'
-        )
+        dependencies_ids = {_id.decode()
+                            for _id in connection.smembers(self.dependencies_key)}
 
-        # As a minor optimization to more quickly tell if all dependencies
-        # are _FINISHED_, sort dependencies by the `ended_at` timestamp so
-        # those jobs which are not yet finished are at the start of the
-        # list. Sorting here lexographically works because these dates are
-        # stored in an ISO 8601 format, so lexographic order is the same as
-        # chronological order.
+        if exclude_job_id:
+            dependencies_ids.discard(exclude_job_id)
+
         dependencies_statuses = [
-            tuple(map(as_text, result))
-            for result in pipe.sort(name=self.dependencies_key, by=sort_by,
-                                    get=get_fields, alpha=True,
-                                    groups=True, )
+            connection.hget(self.key_for(key), 'status')
+            for key
+            in dependencies_ids
         ]
 
-        # if `created_at` is None, then this has been deleted!
-        return all(status == JobStatus.FINISHED or not created_at
-                   for dependency_id, created_at, status
-                   in dependencies_statuses
-                   if not (exclude_job_id and dependency_id == exclude_job_id))
+        return all(
+            status.decode() == JobStatus.FINISHED
+            for status
+            in dependencies_statuses
+            if status
+        )
 
 _job_stack = LocalStack()
--- a/rq/job.py
+++ b/rq/job.py
@@ -13,9 +13,9 @@ from rq.compat import as_text, decode_redis_hash, string_types, text_type
 from .connections import resolve_connection
 from .exceptions import NoSuchJobError
 from .local import LocalStack
+from .serializers import resolve_serializer
 from .utils import (enum, import_attribute, parse_timeout, str_to_date,
                     utcformat, utcnow)
-from .serializers import resolve_serializer
 
 try:
     import cPickle as pickle
@@ -736,9 +736,9 @@ class Job(object):
             connection.sadd(self.dependencies_key, dependency_id)
 
     @property
-    def dependencies_job_ids(self):
+    def dependency_ids(self):
         dependencies = self.connection.smembers(self.dependencies_key)
-        return [Job.key_for(as_text(_id))
+        return [Job.key_for(_id.decode())
                 for _id in dependencies]
 
     def dependencies_are_met(self, exclude_job_id=None, pipeline=None):
@@ -755,7 +755,7 @@ class Job(object):
         pipe = pipeline if pipeline is not None else self.connection
 
         if pipeline is not None:
-            pipe.watch(*self.dependencies_job_ids)
+            pipe.watch(*self.dependency_ids)
 
         sort_by = self.redis_job_namespace_prefix + '*->ended_at'
         get_fields = (
--- a/rq/job.py
+++ b/rq/job.py
@@ -5,10 +5,10 @@ from __future__ import (absolute_import, division, print_function,
 import inspect
 import warnings
 import zlib
+from functools import partial
 from uuid import uuid4
 
 from rq.compat import as_text, decode_redis_hash, string_types, text_type
-
 from .connections import resolve_connection
 from .exceptions import NoSuchJobError
 from .local import LocalStack
@@ -16,6 +16,15 @@ from .utils import (enum, import_attribute, parse_timeout, str_to_date,
                     utcformat, utcnow)
 from .serializers import resolve_serializer
 
+try:
+    import cPickle as pickle
+except ImportError:  # noqa  # pragma: no cover
+    import pickle
+
+# Serialize pickle dumps using the highest pickle protocol (binary, default
+# uses ascii)
+dumps = partial(pickle.dumps, protocol=pickle.HIGHEST_PROTOCOL)
+loads = pickle.loads
 
 JobStatus = enum(
     'JobStatus',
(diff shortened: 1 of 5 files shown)


rq/job.py
```diff
@@ -733,4 +732,43 @@ def register_dependency(self, pipeline=None):
             connection.sadd(dependents_key, self.id)
             connection.sadd(self.dependencies_key, dependency_id)
 
+    def get_dependencies_statuses(
```
selwin: I'm not sure whether we should have this API, or we should just expose a `job.has_unmet_dependencies` that simply returns a boolean indicating whether or not all dependencies are finished.

rq/job.py
```diff
@@ -733,4 +732,43 @@ def register_dependency(self, pipeline=None):
...
+        As a minor optimization allowing callers to more quickly tell if all
+        dependencies are _FINISHED_, the returned list is sorted by the
+        `ended_at` timestamp, so those jobs which are not yet finished are at
+        the start of the list.
+        """
+
+        pipe = pipeline if pipeline is not None else self.connection
+
+        if watch:
+            pipe.watch(self.dependencies_key)
```
selwin: Do we need to watch `dependencies_key`? If I'm not mistaken, this key contains a list of dependencies, which should not change after the job is first enqueued.

thomasmatecki: ah yes - perhaps a bit too defensive here. Probably better to not make the unnecessary call.

thomasmatecki: @selwin in f624ab78432299c8e021604b40636675df2da341

rq/job.py
thomasmatecki: @selwin I've changed such that `Job` now has a method `dependencies_finished`. I'm not in love with the name, but including a _negative_ seemed to make a few things more convoluted; e.g:
```
[ ... for ... in 
if not job.has_unmet_dependencies]
```

Originally I thought exposing `get_dependencies_statuses` might be of use to other projects,(like `django_rq`) but I think you're suggestion is better, because it keeps the loop in `Queue#enqueue_dependents` cleaner.

rq/job.py
```diff
@@ -733,4 +732,38 @@ def register_dependency(self, pipeline=None):
...
+    ):
+        """Returns a boolean indicating if all of this jobs dependencies are _FINISHED_
+
+        If a pipeline is passed, all dependencies are WATCHed.
+        """
+
+        pipe = pipeline if pipeline is not None else self.connection
+
+        if pipeline is not None:
+            pipe.watch(*[Job.key_for(as_text(_id))
```
selwin: I feel like we should have a convenience function to get all dependencies' ID, something like `job.dependencies_ids`?

rq/job.py
```diff
@@ -733,4 +732,38 @@ def register_dependency(self, pipeline=None):
             connection.sadd(dependents_key, self.id)
             connection.sadd(self.dependencies_key, dependency_id)
 
+    def dependencies_finished(
```
selwin: Is `job.dependencies_are_met` a better name? There could be cases where one of the dependencies have finished and are no longer in Redis becase it's cleaned up right away.

rq/job.py
```diff
@@ -733,4 +732,38 @@ def register_dependency(self, pipeline=None):
...
+        # list. Sorting here lexographically works because these dates are
+        # stored in an ISO 8601 format, so lexographic order is the same as
+        # chronological order.
+        dependencies_statuses = [
+            (as_text(_id), as_text(status))
+            for _id, status in pipe.sort(name=self.dependencies_key, by=sort_by,
+                                         get=['#', get_field], alpha=True, groups=True, )
+        ]
+
+        return all(status == JobStatus.FINISHED
```
selwin: We also need to account for situations where jobs are already finished but are no longer in Redis. In this case we can assume that the job is done.

rq/worker.py
```diff
@@ -783,6 +783,7 @@ def handle_job_success(self, job, queue, started_job_registry):
                     # if dependencies are inserted after enqueue_dependents
                     # a WatchError is thrown by execute()
                     pipeline.watch(job.dependents_key)
+                    job.set_status(JobStatus.FINISHED, pipeline=pipeline)
```
selwin: Is there a reason why `job.set_status` is moved here? If the job is deleted right away, this operation doesn't need to be called.

rq/job.py
thomasmatecki: Yes much appreciated - with https://github.com/rq/rq/pull/1170#discussion_r363012688, I see now that I have not been correctly considering jobs with `ttl=0`.

olingerc:
@thomasmatecki. I see that your work on multi-dependencies is progressing nicely! Great job. 
I’m not sure if you’ve attacked tests already. In my multi-dependencies branch there is a certain test case at the end of tests/test_queue.py (writing on phone so no easy way to find the link). in my implementation it would fail 1 in 10000 files. Basically if 2 workers listening to the same queue would become free at the same time and try to enqueue a dependent job, said job would be duplicated.
I will be happy to test your code whenever you feel it is ready. I’m looking very much forward to it!

willyketch:
Nice features! Great job, Congrats !

I would like to test this relevant feature. 

Please, could you tell me approximately when this feature multi dependencies job) will be available ? 

Thanks a lot for answering,

rq/job.py
selwin: Just a reminder that this comment this needs to be addressed

rq/worker.py
selwin: This also needs to be addressed.

thomasmatecki:
@olingerc @willyketch I believe this is quite close to be test-ready, but haven't had the time get it over the line in the last few months. I will endeavor to get something done this weekend!

rq/worker.py
thomasmatecki: @selwin certainly - I have not had a chance to come back to this for some months.

The reason this was moved was tied to a change you commented on above - addressing https://github.com/rq/rq/pull/1170#discussion_r363012688 should make this change unnecessary.

willyketch:
Hi Thomas,

@thomas , Thanks for answering. When this PR will be close, I will test it
promptly.

Bests,

Willy

willyketch:
Hi Thomas,

Thanks for answering. When PR will be closed, I will test this relevant
feature promptly.

Thanks,

Best Regards,

Willy

rq/job.py
thomasmatecki: changed to `dependencies_are_met` in 406d06106744c91cd945509c71a5e8bceac730fa

rq/worker.py
thomasmatecki: Moved back in c416694125755f58fbe2b771ee42894f6f201962 - you are correct; I've done my best to explain the approach taken in the commit message

> When a job with dependents is _successful_ it's dependents are enqueued. Only if the FINISHing job's `result_ttl` is non-zero is the change in status persisted in Redis - that is, when each dependent job is enqueued, the _FINISHing_ job (,triggering the enqueueing,) has an _outdated_ status in redis. This avoids redundant call because if `result_ttl=0` then the job is deleted then deleted in `Job#cleanup`.

> In order to enqueue the dependents, we therefore _exclude_ the FINISHing job from the check if each dependents' dependencies have been met.

rq/job.py
thomasmatecki: agreed! will do.

thomasmatecki: done in e1674e79d12aef7fb16f91b29b0f9be03c551a7a

rq/job.py
thomasmatecki: Addressed in d1ce1b26f8bc60b76f92d806add01096f3c29b31

thomasmatecki:
@selwin I think I've addressed all your comments, albeit with a few things that seem a bit hacky. I'm curious of your thoughts, principally with two things:

* The addition of an `exclude` parameter allows us to _exclude_ the job that  finished successful from the check if dependencies are met.

https://github.com/thomasmatecki/rq/blob/d1ce1b26f8bc60b76f92d806add01096f3c29b31/rq/job.py#L764

If `result_ttl` is zero, then this job's status is not explicitly set to `FINISHED`. In this case, the job is deleted, *including* it's set of dependencies.

* We're using a the `created_at` key in the results of a redis `sort` to determine if a job has been deleted.

https://github.com/thomasmatecki/rq/blob/d1ce1b26f8bc60b76f92d806add01096f3c29b31/rq/job.py#L804

Because a job can be saved without a status and `sort` will return `None` for those keys of a hash that doesn't exist. A job _should_ always have a `created_at` key it's hash, so it being `None` is a pretty clear indicator is has been deleted.

Like I said, both of these approaches seem to _work_ but feel like a bit of a hack. There is _probably_ a refactor that moves much of what is done in `Queue#enqueue_dependents` to a similarly named method of `Job` and moves things around a bit in `Worker#handle_job_success`, BUT it probably a significant amount of shuffling things around.

Cheers~

rq/queue.py
```diff
@@ -458,8 +457,19 @@ def enqueue_dependents(self, job, pipeline=None):
                 if pipeline is None:
                     pipe.watch(dependents_key)
 
-                dependent_jobs = [self.job_class.fetch(as_text(job_id), connection=self.connection)
-                                  for job_id in pipe.smembers(dependents_key)]
+                dependent_job_ids = [as_text(_id)
+                                     for _id in pipe.smembers(dependents_key)]
+
+                dependent_jobs = [
```
selwin: Can we rename this to `jobs_to_enqueue`? Previously all dependent jobs had to be enqueued because there's no multi dependency feature, now we have to do an extra check.

rq/job.py
```diff
@@ -750,4 +748,58 @@ def register_dependency(self, pipeline=None):
             connection.sadd(dependents_key, self.id)
             connection.sadd(self.dependencies_key, dependency_id)
 
+    @property
+    def dependencies_job_ids(self):
+        dependencies = self.connection.smembers(self.dependencies_key)
+        return [Job.key_for(as_text(_id))
+                for _id in dependencies]
+
+    def dependencies_are_met(
```
selwin: Nitpick: make the args one line.

rq/job.py
```diff
@@ -750,4 +748,58 @@ def register_dependency(self, pipeline=None):
             connection.sadd(dependents_key, self.id)
             connection.sadd(self.dependencies_key, dependency_id)
 
+    @property
+    def dependencies_job_ids(self):
```
selwin: Should this be `dependency_ids`?

rq/job.py
```diff
@@ -750,4 +748,58 @@ def register_dependency(self, pipeline=None):
...
+        dependencies = self.connection.smembers(self.dependencies_key)
+        return [Job.key_for(as_text(_id))
```
selwin: I think we can drop `as_text`, since Python 2.7 has been discontinued. Let's new codes Py 3 only from now on.

rq/job.py
selwin: The order of arguments should be `self`, `exclude_job_id`, `pipeline`

rq/job.py
```diff
@@ -750,4 +748,58 @@ def register_dependency(self, pipeline=None):
...
+
+        # As a minor optimization to more quickly tell if all dependencies
+        # are _FINISHED_, sort dependencies by the `ended_at` timestamp so
+        # those jobs which are not yet finished are at the start of the
+        # list. Sorting here lexographically works because these dates are
+        # stored in an ISO 8601 format, so lexographic order is the same as
+        # chronological order.
+        dependencies_statuses = [
+            tuple(map(as_text, result))
+            for result in pipe.sort(name=self.dependencies_key, by=sort_by,
```
selwin: I think we should exclude excluded job before this Redis call so we don't need to fetch this key from Redis.

rq/queue.py
```diff
@@ -458,8 +457,19 @@ def enqueue_dependents(self, job, pipeline=None):
...
+                    dependent_job for dependent_job
+                    in self.job_class.fetch_many(
+                        dependent_job_ids,
+                        connection=self.connection
+                    ) if dependent_job.dependencies_are_met(
+                        pipeline=pipe,
+                        exclude={job.id}
```
selwin: Does exclude need to be a set? I think it can be a single `job_id`

selwin:
> * The addition of an `exclude` parameter allows us to _exclude_ the job that  finished successful from the check if dependencies are met.
It does feel a bit hackish, but since I think it's still ok.

> * We're using a the `created_at` key in the results of a redis `sort` to determine if a job has been deleted.
I'm not familiar with `redis.sort()` but I think it'd be easier to use something like `fetch_statuses(job_ids)` that returns all job statuses. If job is no longer there, that function would just return `None`.

willyketch:
Hello everyone,
Please, do you have some news about this review?

Can i already test multidepencies job with this PR?

Thanks for job already done,

Bests,

selwin:
There’s just one point left in my last review. Once @thomasmatecki addresses that, I can merge this PR in.

thomasmatecki:
@selwin Got it! I will try and find some time this weekend!

willyketch:
@thomasmatecki @selwin Thanks for answering. I am waiting your go to test your nice feature! 
Thanks again for this job,
Bests

willyketch:
Hello everyone,
Please, can I test this multi-dependency feature ? 
Thanks for answering,
Bests

rq/job.py
thomasmatecki: :heavy_check_mark: in 24d2a399164adc6214df71f1507b84d05b8af789

rq/queue.py
thomasmatecki: :heavy_check_mark: in 24d2a399164adc6214df71f1507b84d05b8af789

rq/job.py
thomasmatecki: 0672cd00c6ab5268f83887e66990726904cbe1f9

rq/job.py
thomasmatecki: 0672cd00c6ab5268f83887e66990726904cbe1f9

rq/queue.py
thomasmatecki: 0672cd00c6ab5268f83887e66990726904cbe1f9

rq/job.py
```diff
@@ -725,4 +735,52 @@ def register_dependency(self, pipeline=None):
...
+                                    get=get_fields, alpha=True,
+                                    groups=True, )
+        ]
```
selwin, line 778: I think we can simplify this bit. We should just fetch all job `statuses` in one pipeline and check if they are all `finished` or `None` (has been deleted).

thomasmatecki: 👌 will do --- agreed, this is rather wonky.

thomasmatecki: done in 0b528dae4b763ee8f1f7af9d3a5fd85b301fa6e6 :heavy_check_mark:

rq/job.py
thomasmatecki: 0b528dae4b763ee8f1f7af9d3a5fd85b301fa6e6

thomasmatecki:
@selwin I think all comments have been addressed.

rq/job.py
```diff
@@ -725,4 +735,45 @@ def register_dependency(self, pipeline=None):
...
+            connection.watch(*self.dependency_ids)
+
+        dependencies_ids = {_id.decode()
+                            for _id in connection.smembers(self.dependencies_key)}
+
+        if exclude_job_id:
+            dependencies_ids.discard(exclude_job_id)
+
+        dependencies_statuses = [
+            connection.hget(self.key_for(key), 'status')
```
selwin, line 767: Hopefully one last comment, can we pipeline these calls so we only make 1 trip to Redis?

willyketch:
@selwin @thomas Hello, Nice job. Almost finished!!
I will be happy to test this feature when it will be ready, and give you some feedbacks.

Bests,

rq/job.py
thomasmatecki: 33e4bea

thomasmatecki: @selwin :heavy_check_mark:

wketcha:
Hello @thomasmatecki @selwin Congrats for your your work!! Now, available in 1.4.2. 

I am trying multidepencies jobs to test, but i have some issues (probably syntax issues):

This is my tests:
In **function.py** i define :
def Hello_World():
    print("Hello World")
def Hello_Boy():
    print("Hello Boy")
def Hello_Girl():
    print("Hello Girl")
------------------------------------------
in **cli.py** , i get
from redis import Redis
from rq.queue import Queue
from function import Hello_World, Hello_Boy, Hello_Girl

def main():
    redis_conn = Redis()
    queue = Queue('test', connection=redis_conn)
    job_1 = queue.enqueue(Hello_World)
    job_2 = queue.enqueue(Hello_Boy)
    job_3 = queue.enqueue(Hello_Girl, depends_on=[job_1, job_2])
    print("JOB_3", job_3)

if __name__ == "__main__":
    main()
--------------------------------------------------------------
**Output:** 
I get this error : **redis.exceptions.DataError: Invalid input of type: 'list'. Convert to a bytes, string, int or float first.** 
Please, could you help me how making Job_3 depending to Job_1 and Job_2 ?

Thanks!
Bests

thomasmatecki:
@wketcha the interface to enqueue a job has not yet been updated to accept multiple dependencies.

https://github.com/rq/rq/blob/23d43874b675a1cfc06684e9298b5079c488459c/rq/job.py#L123

One could "work around it" like this:

```python
def main():
    redis_conn = Redis()
    queue = Queue('test', connection=redis_conn)
    job_1 = queue.enqueue(Hello_World)
    job_2 = queue.enqueue(Hello_Boy)
    job_3 = Job.create(Hello_Girl, connection=redis_conn)
    job_3._dependency_ids = [job_1.id, job_2.id]
    queue.enqueue_job(job_3)
```

Admittedly, very clunky though. The work to allow the jobs to be enqueued in the manner above, isn't too involved. I'm happy to take it on when I have the time.

wketcha:
Hi @thomasmatecki,

thanks for answering ! Your work around works, Thanks for answering

Bests,

MW3000:
Hello @thomasmatecki and @selwin. I tried the following:

```
def main():
    redis_connection = Redis()
    queue = Queue(connection=redis_connection)
    download_job = queue.enqueue(tasks.download)
    transform_job_1 = queue.enqueue(tasks.transform1, depends_on=download_job)
    transform_job_2 = queue.enqueue(tasks.transform2, depends_on=download_job)
    upload_job_1 = queue.enqueue(tasks.upload1, depends_on=transform_job_1)
    upload_job_2 = queue.enqueue(tasks.upload2, depends_on=transform_job_2)
    clean_job = Job.create(tasks.cleanup, connection=redis_connection)
    clean_job._dependency_ids = [upload_job_1.id, upload_job_2.id]
    queue.enqueue_job(clean_job)
```

And did expect the jobs beeing executed in an order like this:
```
download_job
transform_job1
transform_job2
upload_job1
upload_job2
clean_job
```

Instead what I get is the following order:
```
download_job
clean_job
transform_job1
transform_job2
upload_job1
upload_job2
```

Is there something wrong with my code, or is this a bug, or were my expectations wrong?

wketcha:
Hello @MW3000 @thomasmatecki @selwin ,

I get the same issues than Martin. Indeed, I have almost the same use case, 

```
def main():
    redis_conn = Redis()
    queue = Queue('test', connection=redis_conn)
    job_1 = queue.enqueue(Hello_World)
    job_2 = queue.enqueue(Hello_Boy, depends_on=job_1)
    job_3 = Job.create(Hello_Girl, connection=redis_conn)
    job_3._dependency_ids = [job_1.id, job_2.id]
    job_3 = queue.enqueue_job(job_3)
```
I get None when fetching results (from Hello_World() and Hello_Boy() ) in my function Hello_Girl(), where as i should get in theory results regarding dependencies. 
When i debug deeper, i noticed that 

- status of job_1 = queued, 
- status of job_2 = **defered**,
-  status of job_3 queued. When printing job_3 i get :  
```[Job('de28376a-4c6e-427b-83f8-de72ceffc564', enqueued_at=datetime.datetime(2020, 6, 2, 14, 12, 51, 75632)), Job('e6505171-6de5-46fa-a700-4d44e7835feb', enqueued_at=**None**)] ```  We have None for enqueue_at for Job_2.

One workaround i did which works is : 

```
def main():
    redis_conn = Redis()
    queue = Queue('test', connection=redis_conn)
    job_1 = queue.enqueue(Hello_World)
    job_2 = Job.create(Hello_Boy, connection=redis_conn) 
    job_2._dependency_ids = [job_1.id]
    job_2 = queue.enqueue_job(job_3)
    job_3 = Job.create(Hello_Girl, connection=redis_conn)
    job_3._dependency_ids = [job_1.id, job_2.id]
    job_3 = queue.enqueue_job(job_3)
```

jobs status i get are : 

-  status of job_1 = queued, 

- status of job_2 = queued, 

- status of job_3 = queued. When printing job_3 i get :
  ```[Job('de28376a-4c6e-427b-83f8-de72ceffc564', enqueued_at=datetime.datetime(2020, 6, 2, 14, 12, 51, 75632)), Job('e6505171-6de5-46fa-a700-4d44e7835feb', enqueued_at=enqueued_at=datetime.datetime(2020, 6, 2, 14, 12, 51, 80114))] ``` 

Now I get expected results in my function Hello_Girl().

**My conclusion**: I think that until now, multi-dependencies jobs works perfectly when all jobs get **queued status**, not deferred.

**Comment by wketcha** [comment]
Hello @thomasmatecki @selwin 

I get one issue when working with multi dependencies rq job.

My use case is: 

```
def main():
    redis_conn = Redis()
    queue = Queue('test', connection=redis_conn)
    job_1 = queue.enqueue(Hello_World)
    job_2 = Job.create(Hello_Boy, connection=redis_conn) 
    job_2._dependency_ids = [job_1.id]
    job_2 = queue.enqueue_job(job_3)
    job_3 = Job.create(Hello_Girl, connection=redis_conn)
    job_3._dependency_ids = [job_1.id, job_2.id]
    job_3 = queue.enqueue_job(job_3)
    job_3_clone = queue.fetch_job(job_3.id)
```

**My output is** : 

_job_1, job_2, job_3, job_3_clone_
```
job_1:  <Job 4fee9879-f962-4efe-9132-6698ed1e421a: Hello_World(0)>
job_2:  <Job 58b3c07c-5c21-495c-b5a6-f6466649dcb3: Hello_Boy(1)>
job_3:  <Job 3b8f8e68-9b6c-4216-adff-2324f35d70a5: Hello_Girl()>
job_3_clone:  <Job 3b8f8e68-9b6c-4216-adff-2324f35d70a5: Hello_Girl()>
```
_Dependencies of job_3, job_3_clone:_
```
job3 depend on **2 jobs** :  [
Job('637afc7d-087d-46d9-b391-b966fcecf26f', enqueued_at=datetime.datetime(2020, 6, 9, 19, 41, 37, 348447)),
Job('05ef9122-2af9-4e35-9a86-8e719ba3bf8e', enqueued_at=datetime.datetime(2020, 6, 9, 19, 41, 37, 350601))]

jobs3 clone depend on only **1 job**:  [Job('637afc7d-087d-46d9-b391-b966fcecf26f', enqueued_at=datetime.datetime(2020, 6, 9, 19, 41, 37, 348447))]
```
My issue is I **don't get the same dependencies between job_3 and job_3_clone**.  Please, this is a well known issue ?

Thanks for answering

wketcha:
Hello @thomasmatecki @selwin ,

sorry to give you  some feedbacks about multi-dependenies feature. It's a nice feature, and i am testing your work for my project, that's why i post some comments or some questions. 

Regarding my previous Post just above, when :

-   I ran only **one worker** to execute the three jobs (job_1, job_2, job_3) with job_3 which depends on [job_1, job_2], all works fine. I get expected job_3 result.

- I ran **three workers** to execute these three jobs, i get failure. I noticed that, three jobs run at the same time in parallel on the three workers, and all get  "started" as Job status. Where as, job_3 should be processed only if job_1 and job_2 are finished. That's why i get failure. This issue is a well  known issue ?  
Please, do you have one workaround to address my issue ?

Thanks in advance, for answering

martin-benson:
I've hit the same issue as @wketcha. I've used the "workaround" patten described by @thomasmatecki and am running multiple workers against the queue and see that the jobs run out of order with respect to dependencies. Is there an adaptation to the workaround that will make it work where multiple workers operate on the same queue? Thanks in advance for any assistance!!

PR #1171, opened 2019-12-17
Add example for unix://

### LINKED ISSUE #1152: Easy Documentation fix!
On the [various pages](http://python-rq.org/docs/workers/) where you use specify ```--url``` from the command line, it would be great if you add in a mention of ```unix:///var/run/redis/redis.sock``` so that users can easily figure out how to connect an rq worker to a unix socket. 

For whatever reason, ```redis:///var/run/redis/redis.sock``` works fine in a flask app but not on the rq worker cli.

goldstar611:
Fix #1152 by adding in an example for connecting to unix sockets from the command line

### DIFF
--- a/docs/docs/workers.md
+++ b/docs/docs/workers.md
@@ -57,7 +57,7 @@ just to scale up your workers temporarily during peak periods.
 
 In addition to `--burst`, `rq worker` also accepts these arguments:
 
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

Issue #1177, opened 2020-01-10
Overriding logging with 1.2.0


frenchja:
There appears to be a change in logging behavior between 1.1.0 and 1.2.0 that I'm having trouble overriding.  Previously, I was able to get the workers to output JSON output by passing in a CustomWorker that used this logging configuration:

```python
from pythonjsonlogger import jsonlogger

import logging
import progressbar
import sys


class CustomJsonFormatter(jsonlogger.JsonFormatter):
    def add_fields(self, log_record, record, message_dict):
        super(CustomJsonFormatter, self).add_fields(log_record, record, message_dict)
        log_record["@timestamp"] = log_record["asctime"]

        if log_record.get("level"):
            log_record["level"] = log_record["level"].upper()
        else:
            log_record["level"] = record.levelname


logger = logging.getLogger()
logHandler = logging.StreamHandler(sys.stdout)
logger.setLevel(logging.INFO)
logHandler.setLevel(logging.INFO)
FORMAT = (
    "%(asctime)s %(name)s %(levelname)s "
    + "%(message)s %(pathname)s %(exc_info)s %(funcName)s %(lineno)d %(thread)d "
    + "%(process)d %(processName)s %(threadName)s"
)
formatter = CustomJsonFormatter(FORMAT)
logHandler.setFormatter(formatter)
logger.addHandler(logHandler)
```

However, now each worker emits two logs, one correctly formatted to stdout and another using the default Python formatter to stderr.  What's the preferred way to override this and emit one log output again?

selwin:
Fixed in https://github.com/rq/rq/pull/1178

levrik:
@selwin I'm on RQ 1.3.0 and I'm at least still facing the issue of duplicate logs.

PR #1178, opened 2020-01-17
fix: use correct logging handler for scheduler

FlorianPerucki:
The current way the logger is configured in the scheduler impacts other loggers in our project. This probably relates to https://github.com/rq/rq/issues/1177 too.

This PR uses RQ's `setup_loghandlers` instead of calling `logging.basicConfig` directly.


selwin:
Thanks!

Issue #1179, opened 2020-01-17
Wrong job status if job is scheduled


coolhacker170597:
Looks like when creating a scheduled job, the job status is not set to "scheduled". Think this thing may cause new problems when deleting jobs from scheduled registry

PR #1181, opened 2020-01-20
Add job status setting in enqueue_at (and in enqueue_in) methods

coolhacker170597:
Update tests for this change
Closes: #1179

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -248,14 +248,14 @@ class Queue(object):
     def create_job(self, func, args=None, kwargs=None, timeout=None,
                    result_ttl=None, ttl=None, failure_ttl=None,
                    description=None, depends_on=None, job_id=None,
-                   meta=None):
+                   meta=None, status=JobStatus.QUEUED):
         """Creates a job based on parameters given."""
         timeout = parse_timeout(timeout) or self._default_timeout
 
         job = self.job_class.create(
             func, args=args, kwargs=kwargs, connection=self.connection,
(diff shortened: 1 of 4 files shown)


rq/queue.py
```diff
@@ -387,6 +387,7 @@ def enqueue_at(self, datetime, func, *args, **kwargs):
         job = self.create_job(func, *args, **kwargs)
         registry = ScheduledJobRegistry(queue=self)
         with self.connection.pipeline() as pipeline:
+            job.set_status(JobStatus.SCHEDULED, pipeline=pipeline)
```
selwin: Instead of changing job status to scheduled here, we should create the job with `scheduled` as status in line 387 so we don't incur an extra Redis hit.

selwin:
I plan on publishing a bugfix sometime in the next two weeks. It would be good if we can get this PR in soon.

rq/queue.py
coolhacker170597: create_job func has no status param, and inside create_job job creating via job.create 'queued' status was hardcoded. So i will add new arg named 'status' into create_job which use 'queued' status as default. Thats will be ok?

coolhacker170597:
> I plan on publishing a bugfix sometime in the next two weeks. It would be good if we can get this PR in soon.

Yeah, this PR should help us with scheduled registry integration into RQ-Dashboard

selwin:
Thanks!

PR #1187, opened 2020-02-05
Respect timeout -1 in worker hard kill

### LINKED ISSUE #1186: Jobs with -1 timeout failing with "work-horse terminated unexpectedly; waitpid returned None"
Since 1.2.1, jobs with a timeout value of `-1` have been failing, with the log message
```
WARNING 2020-02-05 20:51:01,504 (rq.worker.monitor_work_horse): Moving job to FailedJobRegistry (work-horse terminated unexpectedly; waitpid returned None)
```

I think what might be happening is the new hard kill on heartbeat isn't respecting `-1` to mean "no timeout" or "infinite timeout": https://github.com/rq/rq/blob/0080b223fece617ceb33cb22701950b21dba28a1/rq/worker.py#L690-L693

I think on the first heartbeat, it performs the arithmetic `job.timeout + 1`, which'll be `0`, and since the elapsed time will always be higher than that, it immediately kills the work horse.

theY4Kman:
This PR alters the hard kill performed in the worker on heartbeat: timeout values of `-1` are considered "no timeout" or "infinite timeout", and no hard kill is performed.

Fixes #1186

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -688,7 +688,7 @@ class Worker(object):
                 self.heartbeat(self.job_monitoring_interval + 5)
 
                 # Kill the job from this side if something is really wrong (interpreter lock/etc).
-                if (utcnow() - job.started_at).total_seconds() > (job.timeout + 1):
+                if job.timeout != -1 and (utcnow() - job.started_at).total_seconds() > (job.timeout + 1):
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

PR #1191, opened 2020-02-13
Store job exception as a result.

### LINKED ISSUE #1190: When a job throws an exception, the exception should be stored as a result.
The documentation explain that when a job throws an exception, it should be stored as its result and pickled to the caller.

Currently the job.result property returns None as the exception is never stored.

luisbc92:
When a job throws an exception, the exception should be stored in the result as per the documentation.

This commit fixes #1190


selwin:
Fixed in https://github.com/rq/rq/commit/bf93e11f2f98f526039f49341114b8c42fc05b17

Issue #1193, opened 2020-02-15
kill_horse function will create a zombie processes


wevsty:
rq version：1.2.2
When I run rq worker for a long time, I always find some zombie processes.

like this
```bash
xubuntu   16813  0.0  0.0      0     0 ?        Z    04:19   0:00 [python] <defunct>
xubuntu   16814  0.0  0.0      0     0 ?        Z    04:19   0:00 [python] <defunct>
xubuntu   16816  0.0  0.0      0     0 ?        Z    04:19   0:00 [python] <defunct>
```
After debugging, I think the problem lies in the kill_horse function.
```python
    def kill_horse(self, sig=SIGKILL):
        """
        Kill the horse but catch "No such process" error has the horse could already be dead.
        """
        try:
            os.kill(self.horse_pid, sig)
        except OSError as e:
            if e.errno == errno.ESRCH:
                # "No such process" is fine with us
                self.log.debug('Horse already dead')
            else:
                raise
```
os.waitpid() is not called after executing os.kill(),This led to the zombie process.

wevsty:
The fix code has been merged. This issue should be fixed in the next release.

PR #1194, opened 2020-02-15
fix kill_horse will cause zombie processes

wevsty:
fix issue #1193

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -376,6 +376,8 @@ class Worker(object):
         """
         try:
             os.kill(self.horse_pid, sig)
+            self.log.debug('Horse killed')
+            os.waitpid(self.horse_pid, 0)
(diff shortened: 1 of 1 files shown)


rq/worker.py
```diff
@@ -376,6 +376,8 @@ def kill_horse(self, sig=SIGKILL):
         """
         try:
             os.kill(self.horse_pid, sig)
+            self.log.debug('Horse killed')
```
selwin: Can you add the PID number of the killed process in the log message? Also please change the severity level to log.info.

wevsty: > 
> 
> Can you add the PID number of the killed process in the log message? Also please change the severity level to log.info.

Ok, I have modified the code.

selwin:
Thanks!

PR #1201, opened 2020-02-25
Pass job ID to error handlers

### LINKED ISSUE #1192: Pass job_id in Worker.handle_exception extra argument
In `Worker.handle_exception`, the call to `self.log.error` passes various details of the job (namely `func_name`, `args`, `kwargs` and `origin`) in the `extra` `dict`. This is useful because any log handlers that receive the exception's `LogRecord` will not be able to get access to the job via `get_current_job()`, because at that point the job context will be unavailable.

In my use case, I have Django logging set up to email the site admins on exceptions (including job exceptions). The built-in Django `AdminEmailHandler` mails exception details including a full stack trace to the site admins, but I have extended `AdminEmailHandler` to provide extra functionality (for example, certain types of exception that occur in data import jobs can be handled by data analysts making changes to ETL configurations and then re-running the job, and those exceptions should go to those data analysts instead of to the overall site admins). Sometimes the stack trace in question can contain sensitive information, so it would be useful for the email being sent to include a link to the failed job on the Django admin site instead of including the full stack trace (because access to the failed job queue on Django admin can be better controlled). This also provides quick immediate access from the email to the failed job, both for troubleshooting and requeueing (we can have 100s of failed jobs at any given time, and locating them via the full list of failed jobs on the admin can sometimes be time-consuming).

To do this, however, we need the job ID, which as I mentioned is not readily available while the log handler is executing. Is there another way to get the job ID besides `get_current_job()` that I am not aware of, or is a PR which adds `job_id` as another entry in the `extra` argument to the `self.log.error` call in `Worker.handle_exception` the best way forward? I am happy to submit the PR if so.

smaccona:
The worker handles exceptions in the job outside of the job's own context, so an exception handler / logger cannot call `get_current_job()` to obtain the job ID. The job ID can be used to locate the job in the failed job registry, which allows useful behaviors such as linking to a failed job on a dashboard in an error report.

Closes #1192.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -933,6 +933,7 @@ class Worker(object):
             'arguments': job.args,
             'kwargs': job.kwargs,
             'queue': job.origin,
+            'job_id': job.id,
         })
 
         for handler in self._exc_handlers:


selwin:
Thanks!

PR #1219, opened 2020-03-18
Implement Customizable Serializer Support

solababs:
Based of inspiration gotten from https://github.com/rq/rq/pull/1141

This is an approach to create Queues and Jobs with customized serializer objects.
The serializer object should have at least dumps and loads methods.
If no serializer is specified, it defaults to pickle

Usage:

```
import dill
import json, pickle

job = Job(connection=connection, serializer=dill)
queue = Queue(connection=connection, serializer=dill)

queue = Queue(connection=connection, serializer=pickle)  # pickle serializer
queue = Queue(connection=connection, serializer=json)
```

Closes issue https://github.com/rq/rq/issues/369

I'd appreciate any and all feedback on this approach!

### DIFF
--- a/docs/docs/jobs.md
+++ b/docs/docs/jobs.md
@@ -84,7 +84,7 @@ job = Job.create(count_words_at_url,
 
 ## Job / Queue Creation with Custom Serializer
 
-When creating a job or queue, you can pass in a custom serializer that will be used for serializing / de-serializing messages.
+When creating a job or queue, you can pass in a custom serializer that will be used for serializing / de-serializing job arguments.
 Serializers used should have at least `loads` and `dumps` method.
 The default serializer used is `pickle`
 
--- a/docs/docs/jobs.md
+++ b/docs/docs/jobs.md
@@ -82,6 +82,20 @@ job = Job.create(count_words_at_url,
           })
 ```
 
+## Job / Queue Creation with Custom Serializer
+
+When creating a job or queue, you can pass in a custom serializer that will be used for serializing / de-serializing messages.
+Serializers used should have at least `loads` and `dumps` method.
+The default serializer used is `pickle`
+
+```python
+import json
+from rq import Job, Queue
+
+job = Job(connection=connection, serializer=json)
+queue = Queue(connection=connection, serializer=json)
+```
+
 ## Retrieving a Job from Redis
 
 All job information is stored in Redis. You can inspect a job and its attributes
--- a/rq/serializers.py
+++ b/rq/serializers.py
@@ -1,37 +1,20 @@
 import pickle
 
-from .exceptions import DeserializationError
-
-
-class MySerializer(object):
-    """MySerializer Class that uses the user specified serializer module."""
-    def __init__(self, serializer):
-        self.serializer = serializer
-
-    def dumps(self, data):
-        try:
-            serialized_data = self.serializer.dumps(data)
-        except Exception as e:
-            raise DeserializationError("Unserializable return value", data, e)
-
-        return serialized_data
-
-    def loads(self, serialized_string):
-        try:
-            obj = self.serializer.loads(serialized_string)
-        except Exception as e:
-            raise DeserializationError('Could not deserialize', serialized_string, e)
-
-        return obj
+from .compat import string_types
+from .utils import import_attribute
 
 
 def resolve_serializer(serializer):
     """This function checks the user defined serializer for ('dumps', 'loads') methods
     It returns a default pickle serializer if not found else it returns a MySerializer
     The returned serializer objects implement ('dumps', 'loads') methods
+    Also accepts a string path to serializer that will be loaded as the serializer
     """
     if not serializer:
         return pickle
+
+    if isinstance(serializer, string_types):
+        serializer = import_attribute(serializer)
 
     default_serializer_methods = ('dumps', 'loads')
 
@@ -39,4 +22,4 @@ def resolve_serializer(serializer):
         if not hasattr(serializer, instance_method):
             raise NotImplementedError('Serializer should have (dumps, loads) methods.')
 
-    return MySerializer(serializer)
+    return serializer
--- a/rq/serializers.py
+++ b/rq/serializers.py
@@ -1,0 +1,80 @@
+from abc import ABCMeta
+from functools import partial
+from .exceptions import UnpickleError, DeserializationError
+
+try:
+    import cPickle as pickle
+except ImportError:  # noqa  # pragma: no cover
+    import pickle
+
+
+class BaseSerializer(object):
+    __metaclass__ = ABCMeta
+
+    def serialize(self, value):
+        raise NotImplementedError('This method must be implemented')
+
+    def deserialize(self, value):
+        raise NotImplementedError('This method must be implemented')
+
+
+class DefaultSerializer(BaseSerializer):
+    """DefaultSerializer Class that uses the pickle module."""
+    def __init__(self):
+        self.serializer = pickle
+
+    def serialize(self, data):
+        try:
+            serialized_data = partial(pickle.dumps, protocol=pickle.HIGHEST_PROTOCOL)(data)
+        except Exception as e:
+            raise UnpickleError('Unpickleable return value', data, e)
+
+        return serialized_data
+
+    def deserialize(self, pickled_string):
+        try:
+            obj = self.serializer.loads(pickled_string)
+        except Exception as e:
+            raise UnpickleError('Could not unpickle', pickled_string, e)
+
+        return obj
+
+
+class CustomSerializer(BaseSerializer):
+    """CustomSerializer Class that uses the user specified serializer module."""
+    def __init__(self, serializer):
+        self.serializer = serializer
+
+    def serialize(self, data):
+        try:
+            serialized_data = self.serializer.dumps(data)
+        except Exception as e:
+            raise DeserializationError("Unserializable return value", data, e)
+
+        return serialized_data
+
+    def deserialize(self, serialized_string):
+        try:
+            obj = self.serializer.loads(serialized_string)
+        except Exception as e:
+            raise DeserializationError('Could not deserialize', serialized_string, e)
+
+        return obj
+
+
+def resolve_serializer(serializer):
+    """This function checks the user defined serializer for ('dumps', 'loads') methods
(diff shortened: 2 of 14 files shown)


solababs:
CC @selwin

selwin:
Thanks so much for working on this! This is a good PR, but I'd like the API to be simplified a bit. A serializer object should just be something that implements two methods, `dumps` and `loads`.

So this is how it would be used:
```python
import json, pickle

queue = Queue(connection=connection, serializer=pickle)  # pickle serializer
queue = Queue(connection=connection, serializer=json)  # json serializer
```

I believe this method would also work with `dill`, out of the box since it already implements [dumps](https://dill.readthedocs.io/en/latest/dill.html#dill._dill.dumps) and [loads](https://dill.readthedocs.io/en/latest/dill.html#dill._dill.loads).

A custom serializer can be declared as such:
```python
class MySerializer(object):

    @classmethod
    def dumps(cls, data):
    	# ...
    
    @classmethod
    def loads(cls, data):
    	# ...
```

solababs:
> Thanks so much for working on this! This is a good PR, but I'd like the API to be simplified a bit. A serializer object should just be something that implements two methods, `dumps` and `loads`.
> 
> So this is how it would be used:
> 
> ```python
> import json, pickle
> 
> queue = Queue(connection=connection, serializer=pickle)  # pickle serializer
> queue = Queue(connection=connection, serializer=json)  # json serializer
> ```
> 
> I believe this method would also work with `dill`, out of the box since it already implements [dumps](https://dill.readthedocs.io/en/latest/dill.html#dill._dill.dumps) and [loads](https://dill.readthedocs.io/en/latest/dill.html#dill._dill.loads).
> 
> A custom serializer can be declared as such:
> 
> ```python
> class MySerializer(object):
> 
>     @classmethod
>     def dumps(cls, data):
>     	# ...
>     
>     @classmethod
>     def loads(cls, data):
>     	# ...
> ```

Got it, would update it

solababs:
@selwin see updated.

rq/serializers.py
```diff
@@ -0,0 +1,80 @@
...
+    for the ('dumps', 'loads') methods respectively
+    """
+    if not serializer:
+        return DefaultSerializer()
+
+    default_serializer_methods = ('dumps', 'loads')
+
+    for instance_method in default_serializer_methods:
+        if not hasattr(serializer, instance_method):
+            return DefaultSerializer()
```
selwin: This should return an `NotImplementedError`.

rq/serializers.py
```diff
@@ -0,0 +1,80 @@
...
+    __metaclass__ = ABCMeta
+
+    def dumps(self, value):
+        raise NotImplementedError('This method must be implemented')
+
+    def loads(self, value):
+        raise NotImplementedError('This method must be implemented')
+
+
+class DefaultSerializer(BaseSerializer):
```
selwin: I don't think we need this class. Our default serializer is just `pickle`

rq/serializers.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def resolve_serializer(serializer):
+    """This function checks the user defined serializer for ('dumps', 'loads') methods
+    It returns a DefaultSerializer if not found else it returns a MySerializer
+    The returned serializer objects implement ('dumps', 'loads') methods
+    for the ('dumps', 'loads') methods respectively
+    """
+    if not serializer:
+        return DefaultSerializer()
```
selwin: This can be:
```python
def resolve_serializer(serializer):
    if not serializer:
        return pickle
```

tests/test_job.py
```diff
@@ -2,6 +2,8 @@
 from __future__ import (absolute_import, division, print_function,
                         unicode_literals)
 
+import dill
```
selwin: For tests, just use `json` so we don't need to pull in an external library just for this.

rq/serializers.py
solababs: I agree. 
Also remember that our default serializer `pickle` also implements

```
        try:
            serialized_data = partial(pickle.dumps, protocol=pickle.HIGHEST_PROTOCOL)(data)
        except Exception as e:
            raise UnpickleError('Unpickleable return value', data, e)
```

rq/serializers.py
solababs: Comment above applies here also.

What do you think?

rq/serializers.py
selwin: Would this not be backwards compatible in certain instances? One scenario that comes to mind would be if you have workers in different Python versions with different pickling protocols.

rq/serializers.py
```diff
@@ -0,0 +1,80 @@
+from abc import ABCMeta
+from functools import partial
+from .exceptions import UnpickleError, DeserializationError
+
+try:
+    import cPickle as pickle
+except ImportError:  # noqa  # pragma: no cover
+    import pickle
```
selwin: There's only `pickle` in Python 3, so I think we can eliminate the `import cPickle` bit.

tests/test_serializers.py
```diff
@@ -0,0 +1,45 @@
...
+try:
+    import unittest
+except ImportError:
+    import unittest2 as unittest  # noqa
+
+is_py2 = sys.version[0] == '2'
+if is_py2:
+    import Queue as queue
+else:
+    import queue as queue
```
selwin: Don't worry about supporting Python 2 since it's been EOL'ed.

tests/test_serializers.py
```diff
@@ -0,0 +1,45 @@
...
+
+import json
+import sys
+from rq.serializers import resolve_serializer, MySerializer
+from rq.exceptions import DeserializationError
+
+try:
+    import unittest
+except ImportError:
+    import unittest2 as unittest  # noqa
```
selwin: `import unittest` is sufficient (Python 3).

solababs:
@selwin Looking at failing tests, seems like we still need to support python 2.7, correct?

selwin:
No need, feel free to remove Python 2.7 from the test matrix in this PR: https://github.com/rq/rq/blob/master/.travis.yml#L6

rq/serializers.py
```diff
@@ -0,0 +1,42 @@
...
+    if not serializer:
+        return pickle
+
+    default_serializer_methods = ('dumps', 'loads')
+
+    for instance_method in default_serializer_methods:
+        if not hasattr(serializer, instance_method):
+            raise NotImplementedError('Serializer should have (dumps, loads) methods.')
+
+    return MySerializer(serializer)
```
selwin: I don't think we need another layer of abstraction in `MySerializer` class.

If you think about it, it's also a bit weird in that it's a serializer class that accepts another `serializer` as an argument, while the serializer being passed into `resolve_serializer` already has `dumps` and `loads` methods.

So this can be further simplified:

```python
def resolve_serializer(serializer):
     """This function checks the user defined serializer for ('dumps', 'loads') methods
     It returns a default pickle serializer if not found else it returns a MySerializer
     The returned serializer objects implement ('dumps', 'loads') methods
     """
     if not serializer:
         return pickle

    if not hasattr(serializer, 'dumps') or not hasattr(serializer, 'loads'):
        raise NotImplementedError('Serializer should have (dumps, loads) methods.')

    return serializer
```

rq/serializers.py
```diff
@@ -0,0 +1,42 @@
...
+    def loads(self, serialized_string):
+        try:
+            obj = self.serializer.loads(serialized_string)
+        except Exception as e:
+            raise DeserializationError('Could not deserialize', serialized_string, e)
+
+        return obj
+
+
+def resolve_serializer(serializer):
```
selwin: `resolve_serializer` must also be able to import a serializer from a string, for example `path.to.serializer`, similar to how you can specify a string to [import job class](https://github.com/rq/rq/blob/master/rq/queue.py#L73).

This is needed when we run `rq worker --serializer=path.to.serializer`

solababs:
@selwin build errors are based on pypy now

selwin:
You'll need to change the test matrix from `pypy` to `pypy3`.

Do you mind adding some documentation for this feature too? The docs lives under `/docs/`

solababs:
@selwin see updated PR with documentation

solababs:
@selwin any updates on this?

selwin:
So far so good, I think. Please give me a few more days to look at this. :)

selwin:
Do you mind merging master to this branch so we can see whether tests coverage has decreased?

docs/docs/jobs.md
```diff
@@ -82,6 +82,20 @@ job = Job.create(count_words_at_url,
           })
 ```
 
+## Job / Queue Creation with Custom Serializer
+
+When creating a job or queue, you can pass in a custom serializer that will be used for serializing / de-serializing messages.
```
selwin, line 87: Can you please replace `messages` with `job arguments`?

tests/fixtures.py
```diff
@@ -175,3 +175,10 @@ def kill_worker(pid, double_kill, interval=0.5):
         # give the worker time to switch signal handler
         time.sleep(interval)
         os.kill(pid, signal.SIGTERM)
+
+
+class Serializer(object):
+    def loads(self): pass
+
+    def dumps(self): pass
```
selwin, line 183: This class is no longer used. Can you please remove this?

tests/test_serializers.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class TestSerializers(unittest.TestCase):
+    def test_resolve_serializer(self):
+        """Ensure function resolve_serializer works correctly"""
+        serializer = resolve_serializer(None)
+        self.assertIsNotNone(serializer)
+
+        self.assertTrue(hasattr(serializer, 'dumps'))
+        self.assertTrue(hasattr(serializer, 'loads'))
```
selwin, line 19: I think we can simplify this to `self.assertEqual(resolve_serializer(None), pickle)`.

selwin:
@solababs reminder to take another quick look at the last batch of comments. If possible I want to merge this in and make a release with this feature.

solababs:
@selwin sorry for the delayed response, see updated PR

solababs:
@selwin any update on this?

I also noticed that i don't get notifications if i am not mentioned, so if you need to call my attention, please mention me in your comment, thanks.

PR #1238, opened 2020-04-20
Add the queue to the Redis queues set when scheduling a job

mdawar:
Hello,

This pull request adds the queue to the Redis queues set when a job is scheduled, previously if you schedule a job on a queue that doesn't have any jobs then the job will be added to the `ScheduledJobRegistry` but the queue won't be available in the queues set:

```console
127.0.0.1:6379> keys *
1) "rq:scheduled:default"
2) "rq:job:52c77ae6-e85c-43c6-9649-38f293be5bcb"
```

So when getting all the queues using `Queue.all()` we won't get this queue in the list because it wasn't added, so now when scheduling a job the queue will be added to the set in the `Queue.enqueue_at` method which is also used by `Queue.enqueue_in` matching the behavior of the `Queue.enqueue_job` method which adds the queue:

```python
# Add Queue key set
pipe.sadd(self.redis_queues_keys, self.key)
```

This is useful for monitoring purposes and also consistency.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -380,7 +380,7 @@ class Queue(object):
 
         (f, timeout, description, result_ttl, ttl, failure_ttl,
          depends_on, job_id, at_front, meta, args, kwargs) = Queue.parse_args(f, *args, **kwargs)
-        
+
         return self.enqueue_call(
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

Issue #1241, opened 2020-05-03
monitor_work_horse prematurely kills recurring scheduled jobs due to stale started_at


rmartin48:
I use [rq-scheduler](https://github.com/rq/rq-scheduler) (via [django-rq-scheduler](https://github.com/isl-x/django-rq-scheduler)) to do cron style job scheduling. As indicated by @selwin [in this issue](https://github.com/rq/rq/issues/1208#issuecomment-618757142), `rq-scheduler` re-uses the same `job.id` every time a job in re-enqueued which can lead to unexpected behaviour.

One unexpected behaviour that I am experiencing is any recurring job that takes longer than the `DEFAULT_JOB_MONITORING_INTERVAL` of 30 seconds gets killed prematurely by `monitor_work_horse`.

For example, the following job should take 45 seconds to execute and then exit cleanly. The first time it is enqueued, it runs as expected. However, on every subsequent run the work horse is killed.
```
@job(timeout=60)
def long_job():
    print("running...")
    time.sleep(45)
    print("completed")
```
If I schedule the job to run every 5 minutes, it results in this:
```
09:00:00 running...
09:00:45 completed

09:05:00 running...
09:00:30 Killed horse pid 1234
09:00:30 Moving job to FailedJobRegistry (work-horse terminated unexpectedly; waitpid returned None)
```

The following code in `monitor_work_horse` is responsible for killing the work horse:
```
# Kill the job from this side if something is really wrong (interpreter lock/etc).
if job.timeout != -1 and (utcnow() - job.started_at).total_seconds() > (job.timeout + 1):
    self.kill_horse()
    break
```
I was able to determine that `job.started_at` was referencing the time that the job was first run, not the time that the current job was started and therefore was exceeding `job.timeout`.

Further, `prepare_job_execution` and `perform_job` both attempt to set `job.started_at` on the current job so I was confused as to why `monitor_work_horse` was reporting the wrong time. My understanding though is that `prepare_job_execution` and `perform_job` are both executed inside the work horse after it is forked and `monitor_work_horse` executes in the worker, so the updated `started_at` value is not available to the worker.

I am happy to raise a PR to fix this, but I wanted to get the opinion of the package authors on the best way to go about it. I have fixed it in a fork by changing the following line in `monitor_work_horse` so that `job.started_at` always get overwritten:
[...]

PR #1242, opened 2020-05-04
Always set job.started_at in monitor_work_horse

rmartin48:
Solves #1241 where `monitor_work_horse` could have a stale reference to `job.started_at` if the job is being reused by a scheduler (like rq-scheduler).


selwin:
Thanks! I'll make a release sometime in the next few days.

PR #1243, opened 2020-05-04
Parse job_id as keyword argument to delay() (#1236)

### LINKED ISSUE #1236: Could delay() also accept an argument specifying job id?
#415 added an argument to `enqueue()` allowing the caller to specify the id of the job being created. Is it intentional that the `delay()` method of functions decorated with `@job` does not offer the same behavior?

No big deal but I did run into this when I tried switching some uses from `enqueue` to `delay` for the purpose of specifying some common job metadata on the decorated function itself, rather than having to make sure to do so in the several places that function might be enqueued.

grayshirt:
Closes #1236 by popping `job_id` from keyword arguments given to a decorated job function's `delay()` method and passing it to `enqueue_call()`.

### DIFF
--- a/rq/decorators.py
+++ b/rq/decorators.py
@@ -51,6 +51,7 @@ class job(object):  # noqa
                 queue = self.queue
 
             depends_on = kwargs.pop('depends_on', None)
+            job_id = kwargs.pop('job_id', None)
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

grayshirt:
You're welcome!

Issue #1246, opened 2020-05-10
`raise e` vs `raise` causing error to become untraceable


orokusaki:
This line results in the underlying error being untraceable.

https://github.com/rq/rq/blob/9d6f38df0e7629962004f82021e0ed45dd488e2c/rq/worker.py#L757

That also results in a failed task that has no exception value when viewed later (e.g., using `django-rq`). I can see the untraceable exception printed in my logs, which happened to be:

    UnicodeEncodeError: 'ascii' codec can't encode character u'\u201c' in position 2166: ordinal not in range(128)

Also, is that line even intentional or a remnant from local debugging? I see the following unreachable line just below:

https://github.com/rq/rq/blob/9d6f38df0e7629962004f82021e0ed45dd488e2c/rq/worker.py#L758

selwin:
Fixed in https://github.com/rq/rq/pull/1247

PR #1247, opened 2020-05-12
Remove extraneous try/except

orokusaki:
This addresses issue #1246.

The exception handling block was raising the caught exception in-place, which caused the original traceback info to be lost. Rather than replace `raise e` with `raise`, I simply removed the whole try / except, since no action was being taken in the except block.


selwin:
Thanks!

PR #1248, opened 2020-05-13
Use unicode interpolation to avoid str(...)

orokusaki:
Hey @selwin, thanks for merging my other #1247 patch so quickly; I wrote another small patch.

In Python 2, if a `UnicodeDecodeError` occurs in the Job, the Unicode value in `Job.exc_info` results in a new `UnicodeDecodeError` when trying to cast the `exc_info` value to a `str`.

This pull replaces that `str(...)` call with `u'%s' % (...,)`, which essentially functions as `unicode(value)` in Python 2 and `str(value)` in Python 3.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -516,7 +516,7 @@ class Job(object):
             except Exception:
                 obj['result'] = "Unserializable return value"
         if self.exc_info is not None:
-            _exec_info = u'%s' % (self.exc_info,)
+            _exec_info = '%s' % (self.exc_info,)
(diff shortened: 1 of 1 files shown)


selwin:
@orokusaki can you please provide a test case for this? Also, since Python 2 is EOL, we no longer support Python 2.

orokusaki:
@selwin no sweat; this pull is strictly for Python 2 support, so I'm closing it out (we're about to update to 3 in a few weeks anyhow, so this will be moot)

Issue #1244, opened 2020-05-07
Fix Redis DeprecationWarning: Pipeline.hmset() is deprecated. Use Pipeline.hset() instead.


adamchainz:
rq.job and other places are triggering this warning from redis:

```
Pipeline.hmset() is deprecated. Use Pipeline.hset() instead.
```

This was deprecated in redis 3.5.0: https://github.com/andymccurdy/redis-py/blob/252c840ea2ade01637b4ed7c834dd83b130041b4/CHANGES#L17

selwin:
Mind opening a PR for this?

adamchainz:
I don't have the time myself.

selwin:
@bbayles I had to revert your PR in https://github.com/rq/rq/pull/1260 because it's causing crashes in deployments with [Redis server < 4.0](https://github.com/rq/rq/issues/1256).

bbayles:
Alas! The docs aren't quite clear on what Redis server and `redis-py` versions are required/supported. Perhaps they could be updated to be more specific? Currently the front page says:
> RQ requires Redis >= 3.0.0.

And `requirements.txt` says:
> redis>=3.0

Is it a coincidence that the versions are the same?

tomhamiltonstubber:
For clarity, is this an issue or not?

On 1.4.2 I'm still getting a deprecation warning with redis v5.

Happy to create a PR.

dmitrypol:
+1

selwin:
@tomhamiltonstubber yes it's still an issue since the previous PR is crashing deployments on Redis server < 4. @aparcar plans to work on a PR that adds [Redis server version awareness](https://github.com/rq/rq/pull/1271#discussion_r446564849) to RQ. Hopefully we can get this patch in after this change.

aparcar:
What about this case here? https://github.com/rq/rq/blob/master/rq/job.py#L147 It uses `hset` directly instead of the compat function `hmset`? Just trying to understand the current setup

aparcar:
Related https://github.com/rq/rq/pull/1286

selwin:
Fixed in https://github.com/rq/rq/pull/1286/files

PR #1253, opened 2020-05-15
Avoid deprecation warnings on redis-py 3.5.0 hmset

bbayles:
Re: https://github.com/rq/rq/issues/1244, this PR adds an `hmset` function to the compatibility module such that `hset` is preferred over `hmset` when running `redis-py` 3.5.0+.

### DIFF
--- a/rq/compat/__init__.py
+++ b/rq/compat/__init__.py
@@ -104,3 +104,12 @@ except ImportError:
             return timedelta(0)
 
     utc = UTC()
+
+
+def hmset(pipe_or_connection, name, mapping):
+    # redis-py versions 3.5.0 and above accept a mapping parameter for hset
+    try:
+        return pipe_or_connection.hset(name, mapping=mapping)
+    # earlier versions require hmset to be used
+    except TypeError:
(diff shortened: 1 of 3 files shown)


rq/compat/__init__.py
```diff
@@ -103,4 +103,13 @@ def tzname(self, dt):
...
 
-    utc = UTC()
\ No newline at end of file
+    utc = UTC()
+
+
+def hmset(pipe_or_connection, name, mapping):
+    # redis-py versions 3.5.0 and above accept a mapping parameter for hset
+    try:
+        return pipe_or_connection.hset(name, mapping=mapping)
```
bbayles, line 112: This [signature](https://github.com/andymccurdy/redis-py/pull/1271/files#diff-085cbe85ef9bc2ad832d79163eb39172R2999) for `hset` changed in `redis-py` 3.5.0, so this will immediately raise a `TypeError` if it's wrong.

bbayles: An alternative would be to check `redis.__version__`.

selwin:
Thanks!

Issue #1252, opened 2020-05-14
Pickle protocol incompatibilities in 1.4.0


bbayles:
b91b89a183db70f0ca32b1e49eb05a02f9dce0c9 changed the serializer to use `pickle.HIGHEST_PROTOCOL`.

[PR 1219](https://github.com/rq/rq/pull/1219/files#diff-cda0bf8e8e5a8904118d64f3c4c979e0R13-R14) (shipped in 1.4.0) now just `pickle.dumps` as the default serializer. Since it's not called with a `protocol` keyword, it will use [`pickle.DEFAULT_PROTOCOL`](https://docs.python.org/3/library/pickle.html#pickle.DEFAULT_PROTOCOL). This was [commented upon](https://github.com/rq/rq/pull/1219#discussion_r394296664) but not addressed, unfortunately!

This means upgrading `rq` versions (even on the same Python version) changes serialization, which can have negative effects. You can see some of them in the 1.3.0 to 1.4.0 changeset - e.g. [this handler](https://github.com/rq/rq/compare/v1.3.0...v1.4.0#diff-b460bf5f8f26e7bd8f4cb42060d3d5fdL253) for `None` values.

My preference would be for a 1.4.1 to be released, which returns to using `HIGHEST_PROTOCOL` by default. If you'd like, I can make a PR for that?

__Edited to add__: A branch with a proposed fix is [here](https://github.com/rq/rq/compare/master...bbayles:pickle-protocol?expand=1)

selwin:
Yes, please make a PR for this so 1.4.0 to 1.4.1 is backwards compatible. Thanks!

PR #1254, opened 2020-05-16
Use pickle.HIGHEST_PROTOCOL by default

bbayles:
Re: https://github.com/rq/rq/issues/1252, this PR sets the default serializer to use `pickle.HIGHEST_PROTOCOL`, to match with the versions before 1.4.0.

Use can provide a class similar to `DefaultSerializer` to use a different protocol version.

### DIFF
--- a/rq/serializers.py
+++ b/rq/serializers.py
@@ -1,7 +1,13 @@
+from functools import partial
 import pickle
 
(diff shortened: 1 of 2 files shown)


selwin:
Thanks!

PR #1258, opened 2020-05-23
Took into account DST when computing localtime zones.  This wasn't ac…

Evansbee:
This change accounts for local DST offsets when queuing a job with enqueue_at and local time.  As it stands right now, if you were to queue a task for today at 10 PM local time (on the America/EST5EDT time zone) the task would actually get sent in for running at 11 PM due to the fact that time.timezone reports 5 hours of difference whereas because time.daylight. != 0, the time.altzone should be used which correctly makes the offset 4 hours.

### DIFF
--- a/tests/test_scheduler.py
+++ b/tests/test_scheduler.py
@@ -72,7 +72,7 @@ class TestScheduledJobRegistry(RQTestCase):
             # the timestamp maybe different
             registry.schedule(job, datetime(2019, 1, 1))
             self.assertEqual(self.testconn.zscore(registry.key, job.id),
-                             1546300800 + time.timezone)  # 2019-01-01 UTC in Unix timestamp
+                             1546300800 + (time.timezone if time.daylight == 0 else time.altzone))  # 2019-01-01 UTC in Unix timestamp
 
             # Score is always stored in UTC even if datetime is in a different tz
             tz = timezone(timedelta(hours=7))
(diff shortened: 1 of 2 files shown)


tests/test_scheduler.py
```diff
@@ -72,7 +72,7 @@ def test_schedule(self):
             # the timestamp maybe different
             registry.schedule(job, datetime(2019, 1, 1))
             self.assertEqual(self.testconn.zscore(registry.key, job.id),
-                             1546300800 + time.timezone)  # 2019-01-01 UTC in Unix timestamp
+                             1546300800 + (time.timezone if time.daylight == 0 else time.altzone))  # 2019-01-01 UTC in Unix timestamp
```
selwin, line 75: Do you mind improving the test to show that it works properly when daylight saving is both on and off? We should also change the assertion to either use `time.altzone` or `time.timezone` to make sure the result is correct.

Evansbee:
Take a look at this change, should handle both scenarios.  Let me know if you'd like to see additional tests.

selwin:
Thanks!

rstones:
Hi, I have a question about this change, it be great if you could help me clear something up...

I've been having some problems with scheduling jobs, they get run immediately and I've tracked it down to setting this timezone based on whether time.daylight is 0 or 1. It's set to 1 on my laptop so the scheduled time for the job gets shifted to the past. However my time zone (GMT/UTC) isn't currently under daylight savings time.

But from the python time docs https://docs.python.org/3/library/time.html#time.tzset it says:

"It will also set...daylight (to 0 if this timezone does not have any daylight saving time rules, or to nonzero if there is a time, past, present or future when daylight saving time applies)"

This implies to me that time.daylight is a constant for a given timezone, if its 0 we never have to shift the time, if its 1 we will need to shift the time using time.altzone at some point, but we don't necessarily need to shift it right now. So time.daylight = 1 is correct on my system even outside of daylight savings time.

Is my understanding correct here and we should change how the timezone is calculated?

I have just set the correct timezone on my datetime objects before calling enqueue_at() for now so its not urgent but would be interested to hear what you think.

PR #1271, opened 2020-06-11
Add get_job_position and get_position feature

### LINKED ISSUE #1197: Get queue position of job
Hi, is it possible to get the queue position of a job? This would be useful to be available on enqueuing and also on checking the status, allowing the user to see some progress of requested jobs.

aparcar:
Fix #1197

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -153,6 +153,19 @@ class Queue(object):
             if job.origin == self.name:
                 return job
 
+    def get_job_position(self, job_or_id):
+        """Returns the position of a job within the queue
+
+        WARNING: The current implementation has a complexity of worse than O(N)
+        and should not be used for very long job queues. Future implementation
+        may use Redis LPOS command to improve the complexity to O(N) and
+        running natively in Redis C implementation.
+        """
+        job_id = job_or_id.id if isinstance(job_or_id, self.job_class) else job_or_id
+        if job_id in self.job_ids:
+            return self.job_ids.index(job_id)
+        return None
+
     def get_job_ids(self, offset=0, length=-1):
         """Returns a slice of job IDs in the queue."""
         start = offset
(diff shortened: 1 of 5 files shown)


rq/queue.py
```diff
@@ -153,6 +153,19 @@ def fetch_job(self, job_id):
...
+        """Returns the position of a job within the queue
+
+        WARNING: The current implementation has a complexity of worse than O(N)
+        and should not be used for very long job queues. Future implementation
+        may use Redis LPOS command to improve the complexity to O(N) and
+        running natively in Redis C implementation.
+        """
+        job_id = job_or_id.id if isinstance(job_or_id, self.job_class) else job_or_id
+        if job_id in self.job_ids:
+            return self.job_ids.index(job_id)
```
selwin, line 166: https://redis.io/commands/lpos has been added to Redis' docs. Should we first try to use `lpos` before falling back to this?

`lpos` is only available on Redis >= 6.0.6 so most servers won't have it, but the numbers should grow over time.

aparcar: Let's get this into [redis-py first](https://github.com/andymccurdy/redis-py/issues/1353)

aparcar: I created a [PR](https://github.com/andymccurdy/redis-py/pull/1354) let's see how quick this goes upstream. @selwin how do you implement those version checks? Is there a global feature state which determines how functions behave? I'd prefer not to run a try/except each time :)

selwin: We used to have checks for PY2/3. We've never had checks for Redis server version before.

I'm open to however you want to implement this, as long as the implementation is sound. I'm even ok with try/except because this is not a performance sensitive area of the codebase.

aparcar: I'd suggest to merge it now and I'll do a two part PR later, adding *Redis version awareness* and to automatically use LPOS if both py-redis and Redis are the correct version

selwin: Sounds good!

PR #1272, opened 2020-06-11
run_tests: remove special case using rq

aparcar:
The tool seems outdated anyway and the CI uses plain pytest instead of
the script anyway.

### DIFF
--- a/run_tests
+++ b/run_tests
@@ -10,23 +10,10 @@ if ! check_redis_running; then
     exit 2
 fi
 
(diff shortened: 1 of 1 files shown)


aparcar:
Plan B would be to remove the file.

selwin:
I'm ok with removing the file. One less abstraction to worry about. Mind making this change?

aparcar:
Superseded by https://github.com/rq/rq/pull/1281

Issue #1269, opened 2020-06-06
RQ should truncade job args/kwargs to a reasonable length by default


JhonnyBn:
> Apart from that, I think RQ should also truncate job args/kwargs to a reasonable length by default. A PR for this would be much appreciated.

_Originally posted by @selwin in https://github.com/rq/rq/issues/1263#issuecomment-639194904_

Logging job's description with huge arguments makes terminal unreadable and log files too large.
It would be nice if arguments with more than N characters, let's say 100, get truncated.

I guess this could be done [here](https://github.com/rq/rq/blob/29222ea12f/rq/worker.py#L585) (worker.py where it prints job's description) or [here](https://github.com/rq/rq/blob/29222ea12f822e29307ee7ba674166d5031b3ae4/rq/job.py#L668) 
 (job.py, function that creates jobs' descriptions)

selwin:
Fixed in https://github.com/rq/rq/pull/1277

Issue #1263, opened 2020-05-26
Silence/decrease output length


JhonnyBn:
Hello. How can I decrease the output length?
Each time a worker does a job, it prints:
```
15:28:34 default: app.inference.get_prediction(b'\xff\xd8 ..... very very very long .... \xff\xd9') ( job id here )
15:28:47 default: Job OK ( job id here )
```
I want to omit the very very very long part, or silence the output. Is that possible? How?
Thanks

JhonnyBn:
Example code

worker.py
```
import redis
from rq import Worker, Queue, Connection
from config import Config

# Preload AI
import app.inference

listen = ['default']

redis_url = Config.REDIS_URL

conn = redis.from_url(redis_url)

if __name__ == '__main__':
    with Connection(conn):
        worker = Worker(map(Queue, listen))
        worker.work()
```

app.inference.py
```
def get_prediction(image_bytes):
    # process things
    return 0, "worked"
```

app.routes.py
```
from app.inference import get_prediction
from rq import Queue
from worker import conn

queue = Queue(connection=conn)

@app.route('/api/predict', methods=['POST'])
def predict():
	if 'file' not in request.files:
		return
	file = request.files.get('file')
	if not file:
		return
	img_bytes = file.read()
	job = queue.enqueue(get_prediction, img_bytes)
	return jsonify({ "job_id": job.get_id() }), 200
```

selwin:
You can change the [logging level here](https://github.com/rq/rq/blob/master/rq/worker.py#L466).

Apart from that, I think RQ should also truncate job args/kwargs to a reasonable length by default. A PR for this would be much appreciated.

JhonnyBn:
Thank you for the response. I didn't know how logging works so I've read some tutorials online and rq's repository and found multiple ways to do what I wanted.

The best approach I've found is within the worker's parameters, log_job_description=False.

Example code: worker.py
```
if __name__ == '__main__':
    with Connection(conn):
        worker = Worker(map(Queue, listen), log_job_description=False)
        worker.work()
```

The output now is:
```
15:28:34 default: job id here
15:28:47 default: Job OK ( job id here )
```

<hr>

> You can change the [logging level here](https://github.com/rq/rq/blob/master/rq/worker.py#L466).

Thank you again for the response. Changing the logging level to WARNING, I've also reached a solution: no output at all. 

Example code: worker.py
```
if __name__ == '__main__':
    with Connection(conn):
        worker = Worker(map(Queue, listen))
[...]

PR #1277, opened 2020-06-17
Truncate job args/kwargs to a reasonable length

JhonnyBn:
Logging job's description with huge arguments makes terminal unreadable and log files too large.
It would be nice if arguments with more than N characters, let's say 75, get truncated.

See #1263 #1269


selwin:
Thanks!

PR #1286, opened 2020-06-29
Version aware

aparcar:
This is the first thing I came up with, however it's not perfect:

* I used the names `redis_version` and `connection_version`, however `library_version` and `server_version` or something entirely different might be easier to grasp?
* There is no certainty that `redis-py` 3.5.4 will actually support LPOS, I just came up with that version.
* Should the legacy and undocumented StrictVersion module be used (it's part of stdlib)
* Should `queue` contain the version information?
* There is a mixup of using the compat function `hmset` and directly `hset` in the `worker` class. Shouldn't we just raise the minimal version to `3.5.0`?
* Redis server supports `HSET` mappings starting from version 4.0, however there is no trivial way to check the version from within compat function `hmset` as a pipeline can be passed over.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -7,11 +7,11 @@ import pickle
 import warnings
 import zlib
 
+from distutils.version import StrictVersion
 from functools import partial
 from uuid import uuid4
 
-from rq.compat import (as_text, decode_redis_hash, hmset, string_types,
-                       text_type)
+from rq.compat import as_text, decode_redis_hash, string_types, text_type
 from .connections import resolve_connection
 from .exceptions import NoSuchJobError
 from .local import LocalStack
@@ -345,6 +345,7 @@ class Job(object):
         self._dependency_ids = []
         self.meta = {}
         self.serializer = resolve_serializer(serializer)
+        self.redis_server_version = None
 
     def __repr__(self):  # noqa  # pragma: no cover
         return '{0}({1!r}, enqueued_at={2!r})'.format(self.__class__.__name__,
@@ -565,7 +566,21 @@ class Job(object):
         key = self.key
         connection = pipeline if pipeline is not None else self.connection
 
-        hmset(connection, key, self.to_dict(include_meta=include_meta))
+        mapping = self.to_dict(include_meta=include_meta)
+
+        if self.get_redis_server_version() >= StrictVersion("4.0.0"):
+            connection.hset(key, mapping=mapping)
+        else:
+            connection.hmset(key, mapping)
+
+    def get_redis_server_version(self):
+        """Return Redis server version of connection"""
+        if not self.redis_server_version:
+            self.redis_server_version = StrictVersion(
+                self.connection.info("server")["redis_version"]
+            )
+
+        return self.redis_server_version
 
     def save_meta(self):
         """Stores job meta from the job instance to the corresponding Redis key."""
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -13,6 +13,7 @@ import time
 import traceback
 import warnings
 from datetime import timedelta
+from distutils.version import StrictVersion
 from uuid import uuid4
 
 try:
@@ -23,7 +24,7 @@ except ImportError:
 from redis import WatchError
 
 from . import worker_registration
-from .compat import PY2, as_text, hmset, string_types, text_type
+from .compat import PY2, as_text, string_types, text_type
 from .connections import get_current_connection, push_connection, pop_connection
 
 from .defaults import (DEFAULT_RESULT_TTL,
@@ -167,6 +168,8 @@ class Worker(object):
             connection = get_current_connection()
         self.connection = connection
 
+        self.redis_server_version = None
+
         if prepare_for_work:
             self.hostname = socket.gethostname()
             self.pid = os.getpid()
@@ -215,6 +218,15 @@ class Worker(object):
                 self.push_exc_handler(handler)
         elif exception_handlers is not None:
             self.push_exc_handler(exception_handlers)
+
+    def get_redis_server_version(self):
+        """Return Redis server version of connection"""
+        if not self.redis_server_version:
+            self.redis_server_version = StrictVersion(
+                self.connection.info("server")["redis_version"]
+            )
+
+        return self.redis_server_version
 
     def validate_queues(self):
         """Sanity check for the given queues."""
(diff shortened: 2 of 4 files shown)


selwin:
> I used the names redis_version and connection_version, however library_version and server_version or something entirely different might be easier to grasp?

How about being explicit and renaming them to `redis_server_version` and `redis_py_version`?

> There is no certainty that redis-py 3.5.4 will actually support LPOS, I just came up with that version.

Should we use `try/except` on this one? Since the exception happens in Redis library and not at the Redis server level, this shouldn't cause any performance hit.

> Should queue contain the version information?

Only if needed

> Should the legacy and undocumented StrictVersion module be used (it's part of stdlib)

If there's a better alternative, I'm open to it (I have none :))

> There is a mixup of using the compat function hmset and directly hset in the worker class. Shouldn't we just raise the minimal version to 3.5.0?

I'm ok with requiring redis-py >= 3.5.0

> Redis server supports HSET mappings starting from version 4.0, however there is no trivial way to check the version from within compat function hmset as a pipeline can be passed over.

We can add a `redis_server_version` to `hmset`. What do you think?

aparcar:
> How about being explicit and renaming them to redis_server_version and redis_py_version?

Will update the PR

> Should we use try/except on this one? Since the exception happens in Redis library and not at the Redis server level, this shouldn't cause any performance hit.

Okay I'll leave the 6.0.6 check as for the near future nobody will have that version installed anyway and add a try block around the LPOS function.

> Only if needed

I could not come up with a better place as most other modules access the queue at some point it looked like a good fit.

> I'm ok with requiring redis-py >= 3.5.0

Will create a PR

> We can add a redis_server_version to hmset. What do you think?

Will update the PR

aparcar:
I created a PR for `redis-py` 3.5.0 [here](https://github.com/rq/rq/pull/1287). They don't do any compatibility handling nor checking which version of Redis is actually running. Redis 5.x is available via `stretch-backports` which I'd consider as a okayish baseline. Therefore we could just drop support for Redis > 4.0 in an upcoming release and thereby drop `HMSET` entirely. The current mixup of `hmset` and `hset` is *broken* anyway, isn't it?

selwin:
@aparcar I'd actually rather we don't require Redis server >= 4.0 anytime soon because it'll be quite troublesome for many people.

> The current mixup of hmset and hset is broken anyway, isn't it?

Not really, so the current situation is that people on `redis-py` >= 3.5.0 will see a warning saying that `hmset` has been deprecated (but it still works). As far as I know, redis-py maintainer hasn't yet specified when `hmset` is going to be removed from the library.

aparcar:
I reread the `hset` and `hmset` commands had a misconception about it - sorry. I'll adapt the PR accordingly.

aparcar:
I rewrote the whole thing, this should be easier to read.

selwin:
The current approach is cleaner, but `connection.info` is called everytime `redis_min_version` is invoked, meaning one extra trip to Redis everytime a worker or job is saved.

I don't mind `queue.get_job_position()` and `worker.register_birth()` calling `connection.info` when it's needed because this is not an area that's executed often, but `job.save()` would definitely be better off without the extra Redis call because it's called multiple times during a job execution.

selwin:
Just as a follow up, I think it's reasonable for `Worker` to know the server's version info. This means there are a few approaches we can follow from here:
1.  `Worker` can pass on this information to `queue` and `job` (or just job since we're ok with `queue.get_job_position()` to figure out Redis version on it's own). `job` would then decide whether it should use `hset` or `hmset` when `save()` is called.
2. We add an extra argument `use_hset=False` to `job.save()`. This will be the quickest but is not the most elegant solution, as `job` wouldn't know the server version it's connected to.

aparcar:
I'm not yet familiar enough with the `rq` structure, but `Worker` isn't the only class importing `queue` and `job` right? I'd be in favour to apply it for `queue` and `job` at the same time and also not bass any compat args to `job.save()`.

selwin:
Alright.

In that case, let's agree on a few major points on the implementation so you don't waste your time. Are we ok with these major points?

1. We'll have `worker.redis_server_version` and `job.redis_server_version`.

2. [This](https://github.com/rq/rq/blob/master/rq/worker.py#L583) is where `worker` will first see the job it's executing. So this is a place where `worker` can set `job.redis_server_version`.

3. `job.save()` uses `job.redis_server_version()` to determine whether `hset` or `hmset` should be used. If `redis_server_version` is `None`, it will need to look up `redis_server_version` via `connection.info()` before saving.

What do you think?

aparcar:
Sounds good, will look into that

aparcar:
The `get_job_position` function in queue gets either a job or job ID as argument, an ID won't contain the `redis_server_version`. How to deal with that?

selwin:
I'm ok with `queue.get_job_position()` calling `connection.info()` since it's not a performance hotspot. Otherwise, we can use a similar approach and introduce `queue.redis_server_version` too.

rq/job.py
```diff
@@ -565,7 +564,12 @@ def save(self, pipeline=None, include_meta=True):
         key = self.key
         connection = pipeline if pipeline is not None else self.connection
 
-        hmset(connection, key, self.to_dict(include_meta=include_meta))
+        mapping = self.to_dict(include_meta=include_meta)
+
+        if redis_min_version(self.connection, "4.0.0"):
```
selwin, line 569: `redis_min_version` incurs an extra `redis.info` call everytime `job.save()` is called, which is multiple times in every single job lifecycle.

To avoid this, worker should assign `job.redis_server_version` when job is dequeued. `job.save()` should look something like this.

```python
if not self.redis_server_version:
    self.redis_server_version = connection.info("server")["redis_version"]

if self.redis_server_version >= '4.0.0':  # Do a proper version check
   connection.hset(key, mapping=mapping)
else:
    connection.hmset(key, mapping)
```

selwin:
I have a big PR coming in https://github.com/rq/rq/pull/1286/files so it would be good if we can get this in sometime in the next few days so I can take care of potential conflicts.

rq/job.py
aparcar: Please check the latest changes, it applies what you suggest.

rq/worker.py
```diff
@@ -167,6 +168,10 @@ def __init__(self, queues, name=None, default_result_ttl=DEFAULT_RESULT_TTL,
             connection = get_current_connection()
         self.connection = connection
 
+        self.redis_server_version = StrictVersion(
```
selwin, line 171: Let's keep this to `None` and use the `get_redis_server_version()` pattern like in `queue.py` and `job.py`. This way when `Worker()` is instantiated it doesn't implicitly connect to Redis.

aparcar:
Updated

selwin:
Thanks!

PR #1287, opened 2020-06-29
Improve requirements handling

aparcar:
Better handling of requirements by keeping them in sync and raise `redis-py` to 3.5.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ services:
   - redis
 matrix:
   include:
-  - python: "3.4"
   - python: "3.5"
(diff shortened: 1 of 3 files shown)


selwin:
Could you please remove Python 3.4 from the test matrix? `redis-py` >= 3.5.0 no longer supports Python 3.4 so we have to drop it too.

selwin:
Thanks!

tony:
@aparcar @selwin Hi!  Any context behind raising redis to 3.5+?

(asking because we're pinned at a lower version of redis at the moment)

PR #1290, opened 2020-07-05
Replace travis with GitHub action

aparcar:


selwin:
Whoa, this is an awesome pull request. Thanks!

.github/workflows/lint.yml
```diff
@@ -0,0 +1,32 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8
+
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```
selwin, line 32: The complexity is a bit high at 10. I think the recommended complexity ceiling is 5. Is this set to be this way because certain methods in RQ are too complex?

Can we change the max line length to 120?

.github/workflows/test.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  build:
+    name: Python${{ matrix.python-version }}/Redis${{ matrix.redis-version }}/redis-py${{ matrix.redis-py-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8]
+        redis-version: [3, 4, 5, 6]
+        redis-py-version: [3.5.0, 3.5.3]
```
selwin, line 17: Do we need to test for two redis-py versions for now? If not, we can keep the list syntax but only put 3.5.3.

README.md
```diff
@@ -5,7 +5,7 @@ easily.
 
 RQ requires Redis >= 3.0.0.
 
-[![Build status](https://travis-ci.org/rq/rq.svg?branch=master)](https://secure.travis-ci.org/rq/rq)
+![Test rq](https://github.com/aparcar/rq/workflows/Test%20rq/badge.svg)
```
selwin, line 8: Link to the badge should be set to RQ's repo. Can we also change this to `Build status`?

.github/workflows/lint.yml
aparcar: These are the defaults from GitHub, I changed the to your values.

.github/workflows/test.yml
aparcar: Why not? If we don't test 3.5.0 than we should set the minimum required `redis-py` version to 3.5.3.

README.md
aparcar: Oops, no self promotion intended here.

aparcar: And it's also renamed to `Build status` now.

selwin:
@aparcar it seems like we forgot to take codecov into account and this wasn't migrated. Do you mind seting up a code coverage service in Github Actions? It can be codecov or something similar (preferably better).

PR #1291, opened 2020-07-06
setup: Set min Python version to 3.5

aparcar:
As `redis-py` in version 3.5.0 is now required, whoch only support
Python 3.5+, we should remove Python 3.4 from the supported Python
versions as well.


aparcar:
Is [this](https://github.com/rq/rq/blob/master/rq/job.py#L690) still required?

selwin:
Thanks!

Issue #1288, opened 2020-06-30
Code style guide?


aparcar:
Is there a specific style for the code? There is [at least one](https://github.com/rq/rq/blob/master/rq/job.py#L46) flake8 error in the code, shouldn't that be automatically checked via the CI?

Also a auto-linter like [black](https://github.com/psf/black) could be applied, however introducing a massive code change...

selwin:
I think we can start enforcing flake8 as part of the CI build for starters. Black can also be considered, but that's a longer term plan. One step at a time :)

aparcar:
#1290 related

aparcar:
https://github.com/rq/rq/pull/1292 related

PR #1292, opened 2020-07-07
Black style

aparcar:
This PR applied the black code style, adds it to the CI and README. The changes look big, however no functional changed are applied at all.

The line length is set to [120](https://github.com/rq/rq/pull/1290#discussion_r449855514), which is different from black recommendation of [88](https://github.com/psf/black#command-line-options). I would suggest to stick to the 88 recommendation and adapt the `flake8` check.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install flake8
+        pip install flake8 black
 
     - name: Lint with flake8
       run: |
@@ -33,3 +33,7 @@ jobs:
         flake8 . --select=E9,F63,F7,F82 --show-source
         # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
         flake8 . --exit-zero --max-complexity=5
+
+    - name: Lint with black
+      run: |
+        black -S -l 120 rq/
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ RQ requires Redis >= 3.0.0.
 [![Build status](https://github.com/rq/rq/workflows/Test%20rq/badge.svg)](https://github.com/rq/rq/actions?query=workflow%3A%22Test+rq%22)
 [![PyPI](https://img.shields.io/pypi/pyversions/rq.svg)](https://pypi.python.org/pypi/rq)
 [![Coverage](https://codecov.io/gh/rq/rq/branch/master/graph/badge.svg)](https://codecov.io/gh/rq/rq)
+[![Code style: black](https://img.shields.io/badge/code%20style-black-000000.svg)](https://github.com/psf/black)
+
 
 Full documentation can be found [here][d].
 
--- a/rq/__init__.py
+++ b/rq/__init__.py
@@ -1,7 +1,6 @@
 # flake8: noqa
 
-from .connections import (Connection, get_current_connection, pop_connection,
-                          push_connection, use_connection)
+from .connections import Connection, get_current_connection, pop_connection, push_connection, use_connection
 from .job import cancel_job, get_current_job, requeue_job, Retry
 from .queue import Queue
 from .version import VERSION
--- a/rq/cli/cli.py
+++ b/rq/cli/cli.py
@@ -10,22 +10,34 @@ import click
 from redis.exceptions import ConnectionError
 
 from rq import Connection, Retry, __version__ as version
-from rq.cli.helpers import (read_config_file, refresh,
-                            setup_loghandlers_from_args,
-                            show_both, show_queues, show_workers, CliConfig, parse_function_args,
-                            parse_schedule)
+from rq.cli.helpers import (
+    read_config_file,
+    refresh,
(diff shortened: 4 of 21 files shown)


selwin:
> The line length is set to 120, which is different from black recommendation of 88. I would suggest to stick to the 88 recommendation and adapt the flake8 check.

Let's keep it at 120 as most monitors are now wide enough to show 120 characters with no truncation. Forcing some expressions to be split across multiple lines also makes them harder to read in some cases.

I didn't realize black didn't have single/double quote options. This is not a deal breaker, but it's such a shame since single quotes are easier to read. I'm tempted to apply black with `--skip-string-normalization` option.

selwin:
Also, I think we can keep just `redis-py` either one of `3.5.0` and `3.5.3` so the number of tests aren't so large. I think a good compromise for now is to have one minor Redis py version in the matrix and we ignore the patch versions.

aparcar:
> Also, I think we can keep just redis-py either one of 3.5.0 and 3.5.3 so the number of tests aren't so large. I think a good compromise for now is to have one minor Redis py version in the matrix and we ignore the patch versions.

ACK

> I didn't realize black didn't have single/double quote options. This is not a deal breaker, but it's such a shame since single quotes are easier to read. I'm tempted to apply black with --skip-string-normalization option.

I'd recommend the default to stick as close as possible to the *best practises*...

aparcar:
I just rebased this. Is the `'` vs `"` your only concern about this patch?

selwin:
> I just rebased this. Is the ' vs " your only concern about this patch?

Yes. But I'd like to merge https://github.com/rq/rq/pull/1299#issue-447166214 in before getting this PR in so it this doesn't cause merge conflicts.

nzjrs:
FWIW I think these sort of pull requests are more harmful than helpful. They make it harder for everyone that has to maintain out of tree forks for internal or production use cases, and they don't buy that much.

IMO it's more important that a codebase is internally consistent than consistent with some external style.

The amount of work in running a code auto-formatter against a codebase is trivial, so this PR could be recreated at any time

aparcar:
@nzjrs I'm wondering if they can just apply the black style once to their on tree and extract a reformatted patch. The idea of using a code formatter is to make such things easier as well, as you have to adapt to it only once.

> IMO it's more important that a codebase is internally consistent than consistent with some external style.

It is mostly, I guess, but the more people join the less it will be that way. Sure @selwin can focus on that during reviews, but isn't the very idea of auto-formatters to automate that?

> The amount of work in running a code auto-formatter against a codebase is trivial, so this PR could be recreated at any time

That's true, I don't really care when (if ever) this is applied, I just see it as a improvement and therefore created the PR. I can rebase it anytime.

selwin:
I understand both @nzjrs' point as well as having a nice, consistent code base.

If we proceed with this, do I have to run black on my IDE; or does Github have an action that automatically rewrites PRs based on Black rules we defined?

aparcar:
@selwin you'd need to run it each time locally.

codyflynnNC:
@selwin  @aparcar You can configure black to run automatically as a github action(https://github.com/psf/Black/actions/runs/17913292/workflow). You can also configure it run automatically before each commit as a pre-commit hook(https://black.readthedocs.io/en/stable/version_control_integration.html).

homeworkprod:
> > I didn't realize black didn't have single/double quote options. This is not a deal breaker, but it's such a shame since single quotes are easier to read. I'm tempted to apply black with --skip-string-normalization option.
> 
> I'd recommend the default to stick as close as possible to the _best practises_...

Choosing one type of string delimiter over the other is hardly a "best practice".

Being consistent might be. Then again, there are situations when one is more handy than the other, and IMO the idea of Black to just force one kind of style just clashes with that.

After all, they *have* introduced the option to not "normalize" quotes. I use it in every project in which I use Black.

aparcar:
Rebased if that's still of interest.

selwin:
@aparcar if we can keep the single quotes, I'm game :). Thanks for your effort in maintaining this.

aparcar:
@selwin done, to make things pretty you need to run `black -S`

aparcar:
Rebased.

lowercase00:
@aparcar @selwin is this still alive? I do agree that just running `black` would be a great idea a lot easier to make it consistent specially since the codebase does have styling that is not very easy to follow. We do have this and #1288 to close :)

aparcar:
Im happy to rerun the command

lowercase00:
@selwin let us know if there's anything to polish on this (I guess black config is the most relevant thing here).

selwin:
I'm mostly ok with using black. Now the only thing left for me would be to integrate black into my RQ development workflow :)

lowercase00:
If you're on VSCode there's a simple `formatOnSave` that does that automatically.

aparcar:
Updated. I did not format the tests though.

aparcar:
By the way, patches like this seem to be a fine way to sneak in evil code. Please rerun `black` on your local machine and make sure that that you end up with he same diff as I created here :)

selwin:
Thanks!

PR #1293, opened 2020-07-07
Fstrings

aparcar:
Use f-strings instead of other formatting.


aparcar:
Nevermind this PR, I wrongly memorized that f-strings are available in 3.5, however it requires 3.6+...

PR #1295, opened 2020-07-07
Python 2.7 supported dropped in 1.4.0 (CHANGES.md)

mbarkhau:
With revision 02eb983e90aeb8e878e5c0f9989b7b701b8ad21a the `python_requires` was set to `3.4`


aparcar:
There are some files in `rq/compat/` which could also be removed as Python2 is gone...

mbarkhau:
My intention here was only to fill the gap in the documention.

selwin:
Thanks!

Issue #1234, opened 2020-04-17
ChildProcessError: [Errno 10] No child processes


levrik:
After upgrading from RQ 1.1.0 to 1.3.0 I'm seeing the above error with my custom worker class.

That custom class is quite similar to the HerokuWorker:
```python
class JobCancelledException(Exception):
    pass


class CustomWorker(Worker):
    DEFAULT_JOB_CANCEL_TIMEOUT = 60

    def __init__(self, *args, job_cancel_timeout=None, **kwargs):
        super().__init__(*args, **kwargs)
        self.job_cancel_timeout = (
            parse_timeout(job_cancel_timeout) or self.DEFAULT_JOB_CANCEL_TIMEOUT
        )

    def handle_warm_shutdown_request(self):
        if self.horse_pid != 0:
            self.log.info(
                "Warm shut down requested, requesting delayed cancelling of running job"
            )
            self.kill_horse(sig=signal.SIGTERM)
        else:
            self.log.info("Warm shut down requested")

    def setup_work_horse_signals(self):
        """Setup signal handing for the newly spawned work horse."""
        signal.signal(signal.SIGINT, signal.SIG_IGN)
        signal.signal(signal.SIGTERM, self.request_stop_sigterm)

    def request_stop_sigterm(self, signum, frame):
        signal.signal(signal.SIGALRM, self.request_force_stop_sigterm)
        signal.alarm(self.job_cancel_timeout)

    def request_force_stop_sigterm(self, signum, frame):
        raise JobCancelledException(
            "Task cancellation exceeded maximum timeout value "
            "({0} seconds)".format(self.job_cancel_timeout)
        )
```

The exception is raised by the call to `os.waitpid` in the `monitor_work_horse` function.
This seems to have been introduced by #1194. If I revert this change, the process exits without that exception after the timeout.
I think that the `os.waitpid` call in `kill_horse` is blocking the thread and when execution continues, the `os.waitpid` in `monitor_work_horse` doesn't have anything to monitor anymore and raises.
I didn't try the HerokuWorker but since the setup is basically the same, I think we're facing the same issue there.

Actually this issue seems to happen only before the first `os.waitpid` in `monitor_work_horse` timed out.

I'm still investigating more.

levrik:
I think I figured out some more details here.

If I'm killing the work-horse in `handle_warm_shutdown_request()` the call to `kill_horse` blocks until my timeout expires. That basically means that `self._stop_requested` was also never set to `True` because the worker never got there.

[...]

PR #1300, opened 2020-07-11
Split kill_house() fix issues #1234

wevsty:
Details View issues #1234

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -697,7 +697,7 @@ class Worker(object):
         while True:
             try:
                 with UnixSignalDeathPenalty(self.job_monitoring_interval, HorseMonitorTimeoutException):
-                    retpid, ret_val = self.wait_horse()
+                    retpid, ret_val = self.wait_for_horse()
                 break
             except HorseMonitorTimeoutException:
                 # Horse has not exited yet and is still running.
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -385,7 +385,7 @@ class Worker(object):
             else:
                 raise
 
-    def wait_horse(self):
+    def wait_for_horse(self):
         """
         A waiting the end of the horse process and recycling resources.
         """
@@ -408,7 +408,7 @@ class Worker(object):
         if self.horse_pid:
             self.log.debug('Taking down horse %s with me', self.horse_pid)
             self.kill_horse()
-            self.wait_horse()
+            self.wait_for_horse()
         raise SystemExit()
 
     def request_stop(self, signum, frame):
@@ -707,7 +707,7 @@ class Worker(object):
                 # Kill the job from this side if something is really wrong (interpreter lock/etc).
                 if job.timeout != -1 and (utcnow() - job.started_at).total_seconds() > (job.timeout + 60):
                     self.kill_horse()
-                    self.wait_horse()
+                    self.wait_for_horse()
                     break
 
             except OSError as e:
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -390,11 +390,10 @@ class Worker(object):
         """
         A waiting the end of the horse process and recycling resources.
         """
+        pid = None
+        stat = None
         try:
-            pid = None
-            stat = None
-            if self.horse_pid:
-                pid, stat = os.waitpid(self.horse_pid, 0)
+            pid, stat = os.waitpid(self.horse_pid, 0)
         except ChildProcessError as e:
             # ChildProcessError: [Errno 10] No child processes
             pass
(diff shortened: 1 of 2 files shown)


selwin:
Can you try to come up with a test to verify this behavior? Changing this area of the code is always a bit risky so I want to be extra careful to see that exit codes are handled correctly.

Also, we'll need all tests to pass before we merge this in.

wevsty:
> 
> 
> Can you try to come up with a test to verify this behavior? Changing this area of the code is always a bit risky so I want to be extra careful to see that exit codes are handled correctly.
> 
> Also, we'll need all tests to pass before we merge this in.

Only one part of the test failed.
According to the logs, it is in HerokuWorkerShutdownTestCase.test_ handle_shutdown_request in this function.
`
__________ HerokuWorkerShutdownTestCase.test_handle_shutdown_request ___________

self = <tests.test_worker.HerokuWorkerShutdownTestCase testMethod=test_handle_shutdown_request>
mock_logger_info = <MagicMock name='info' id='139955829528224'>

    @mock.patch('rq.worker.logger.info')
    def test_handle_shutdown_request(self, mock_logger_info):
        """Mutate HerokuWorker so _horse_pid refers to an artificial process
        and test handle_warm_shutdown_request"""
        w = HerokuWorker('foo')
    
        path = os.path.join(self.sandbox, 'shouldnt_exist')
        p = Process(target=create_file_after_timeout, args=(path, 2))
        p.start()
        self.assertEqual(p.exitcode, None)
    
        w._horse_pid = p.pid
        w.handle_warm_shutdown_request()
        p.join(2)
        # would expect p.exitcode to be -34 but for some reason os.waitpid is setting it to None, even though
        # the process has ended
>       self.assertEqual(p.exitcode, None)
E       AssertionError: -34 != None

tests/test_worker.py:1217: AssertionError
`
In particular, the comment states that the test was written with the expected value of -34, but that the test returned a value of None.
The change resulted in what the developer expected, so I think the change itself should be fine.
Maybe the failing test issue requires a change in the test script, but I'd like to hear your thoughts before doing so.

selwin:
I think the current approach (splitting the kill and wait) makes a ton of sense. Catching `ChildProcessError` is also 👍 since if the process is already gone, it's safe for the worker to continue.

If possible, I'd like a test case to cover issues raised in https://github.com/rq/rq/issues/1234 .

> In particular, the comment states that the test was written with the expected value of -34, but that the test returned a value of None.
The change resulted in what the developer expected, so I think the change itself should be fine.

You're right. So this is actually a good change.

rq/worker.py
```diff
@@ -378,15 +378,27 @@ def kill_horse(self, sig=SIGKILL):
...
-            os.waitpid(self.horse_pid, 0)
-            self.log.info('Killed horse pid %s', self.horse_pid)
         except OSError as e:
             if e.errno == errno.ESRCH:
                 # "No such process" is fine with us
                 self.log.debug('Horse already dead')
             else:
                 raise
 
+    def wait_horse(self):
```
selwin, line 388: Can we rename this to `wait_for_horse()`?

wevsty:
> 
> 
> I think the current approach (splitting the kill and wait) makes a ton of sense. Catching `ChildProcessError` is also 👍 since if the process is already gone, it's safe for the worker to continue.
> 
> If possible, I'd like a test case to cover issues raised in #1234 .
> 
> > In particular, the comment states that the test was written with the expected value of -34, but that the test returned a value of None.
> > The change resulted in what the developer expected, so I think the change itself should be fine.
> 
> You're right. So this is actually a good change.

I've made the changes visually and the latest commit has passed the test.

But I don't have a good idea about adding new tests to validate this.
I think the original test could have covered this issue, it was just ignored before.

rq/worker.py
```diff
@@ -387,6 +386,20 @@ def kill_horse(self, sig=SIGKILL):
             else:
                 raise
 
+    def wait_for_horse(self):
+        """
+        A waiting the end of the horse process and recycling resources.
+        """
+        try:
+            pid = None
+            stat = None
```
selwin, line 395: `pid` and `stat` should be declared outside of the `try/except` block.

rq/worker.py
```diff
@@ -387,6 +386,20 @@ def kill_horse(self, sig=SIGKILL):
...
+            if self.horse_pid:
```
selwin, line 396: No need to check for the existence of `horse_pid` since this method is only called by the main worker.

wevsty: These codes have been adjusted.

PR #1301, opened 2020-07-11
Fix tox test

lhfelis:
Tox test is broken for some time. Without github workflow, testing locally using tox may be still needed, I think.
Tested OK in ubuntu19.10 with `tox` command.

* Include requirements.txt in MANIFEST.in, to make `python setup.py sdist` good when tox generating project package.
* Follow config files in .github/workflows/ to update the python testing environment and testing commands.


lhfelis:
Before executing `tox`, let redis server run on `127.0.0.1:6379` for test.

aparcar:
I've never used tox, it automatically sets some stuff to appear as being e.g. Python 3.6? 

I'm testing with pyenv and running pytest. I think it's a idea to introduce anything that has to be kept in sync, this already caused some inconsistency between `requirements.txt` and `setup.py`.

So wouldn't it be a cleaner way to run `tox` from the CI?

selwin:
I find it troublesome to support both `tox` and `pytest`. Is there any reason why we should have `tox` configuration in this repo without running this on CI?

I'm not sure why we need to maintain both tox and pytest (although I recall vaguely merging this in).

aparcar:
I guess  using tox comes closer to what GitHub actions do now: Testing multiple things at once. I'm also fine with removing the current matrix setup and just run tox.

lhfelis:
With tox, I use `pyenv local 3.5.9 3.6.10 3.7.6 3.8.3` to set up python environtments and run `tox` good.

Maybe tox can handle most of the logic to make developers able to locally do tests easily and as thoroughly as possible. GitHub Actions can simply trigger tox and do not touch testing details much.

I'm trying to make a patch to see whether it's a good idea.

lhfelis:
Fix tox.ini and let GitHub Actions to call tox now.

Improvement:
* GitHub Actions does not touch unit tests details such as deps now. Just get tox command ready and execute tox script.
* `tox_current_py.sh` gets pytest things done, with current python from `$PATH`. It is simple for both GitHub Actions and developers.
* If developer has full python environments locally, `tox` command could do almost full test but with only one kind of redis server on port 6379.  Developer could also execute `tox_current_py.sh` to test on one python version, to get confidence before pushing to Github. 

Maintenance:
* Change python versions. Modify `python-version` in `.github/workflows/test.yml`, then modify `envlist` in `tox.ini`.
* Change redis-py versions. Modify `deps` and `envlist` in `tox.ini`.
* More static checks. Add steps in `.github/workflows/lint.yml` and `tox.ini`.

lhfelis:
> I find it troublesome to support both `tox` and `pytest`. Is there any reason why we should have `tox` configuration in this repo without running this on CI?
> 
> I'm not sure why we need to maintain both tox and pytest (although I recall vaguely merging this in).

We need one unit test entry point for both CI and developers. I checked tox.ini/setup.py/Actions and found that it's not easy for newcomers to test right and quickly.

lhfelis:
> I guess using tox comes closer to what GitHub actions do now: Testing multiple things at once. I'm also fine with removing the current matrix setup and just run tox.

Agreed. CI layer should better be thin.

Issue #1264, opened 2020-05-26
"cannot pickle '_thread.lock' object" - Python 3.8.2 


orokusaki:
The traceback is as follows:

    Traceback (most recent call last):
      File "worker.py", line 54, in <module>
        worker.work(with_scheduler=True)
      File "/Users/me/Library/Caches/pypoetry/virtualenvs/src-[hash]-py3.8/lib/python3.8/site-packages/rq/worker.py", line 492, in work
        self.scheduler.start()
      File "/Users/me/Library/Caches/pypoetry/virtualenvs/src-[hash]-py3.8/lib/python3.8/site-packages/rq/scheduler.py", line 162, in start
        self._process.start()
      File "/Users/me/.pyenv/versions/3.8.2/lib/python3.8/multiprocessing/process.py", line 121, in start
        self._popen = self._Popen(self)
      File "/Users/me/.pyenv/versions/3.8.2/lib/python3.8/multiprocessing/context.py", line 224, in _Popen
        return _default_context.get_context().Process._Popen(process_obj)
      File "/Users/me/.pyenv/versions/3.8.2/lib/python3.8/multiprocessing/context.py", line 283, in _Popen
        return Popen(process_obj)
      File "/Users/me/.pyenv/versions/3.8.2/lib/python3.8/multiprocessing/popen_spawn_posix.py", line 32, in __init__
        super().__init__(process_obj)
      File "/Users/me/.pyenv/versions/3.8.2/lib/python3.8/multiprocessing/popen_fork.py", line 19, in __init__
        self._launch(process_obj)
      File "/Users/me/.pyenv/versions/3.8.2/lib/python3.8/multiprocessing/popen_spawn_posix.py", line 47, in _launch
        reduction.dump(process_obj, fp)
      File "/Users/me/.pyenv/versions/3.8.2/lib/python3.8/multiprocessing/reduction.py", line 60, in dump
        ForkingPickler(file, protocol).dump(obj)
    TypeError: cannot pickle '_thread.lock' object

I found [this recent post on Stack Overflow](https://stackoverflow.com/questions/60735003/typeerror-cannot-pickle-thread-lock-object-with-rq), corroborating the existence of the issue.

selwin:
Are you able to test this on Python 3.7 or Python 3.8.0/3.8.1? I'm just wondering whether this is a Python bug.

orokusaki:
I haven’t yet tried in 3.8.1, but it functions properly in 3.7.7.

[...]

PR #1302, opened 2020-07-12
Fix scheduler on Python 3.8

selwin:
Fix https://github.com/rq/rq/issues/1264


aparcar:
Should there be an extra test case?

selwin:
[This test](https://github.com/rq/rq/blob/master/tests/test_scheduler.py#L276) fails on Python 3.8.3, but not on Python 3.8. Bumping the Python version in Github Actions should catch this. I'll change the Python version from `3.8` to `3.8.3`.

PR #1303, opened 2020-07-15
handled unhandled exceptions in horse

theambient:
handled unhandled exceptions in horse to prevent a job from being silently dropped without going into FailedRegistry

In my case I got exceptions like this:
```
TimeoutError: [Errno 110] Connection timed out
  File "redis/connection.py", line 539, in connect
    sock = self._connect()
  File "redis/connection.py", line 596, in _connect
    raise err
  File "redis/connection.py", line 584, in _connect
    sock.connect(socket_address)
ConnectionError: Error 110 connecting to <url>:6379. Connection timed out.
  File "rq/worker.py", line 477, in work
    self.execute_job(job, queue)
  File "rq/worker.py", line 670, in execute_job
    self.fork_work_horse(job, queue)
  File "rq/worker.py", line 610, in fork_work_horse
    self.main_work_horse(job, queue)
  File "rq/worker.py", line 687, in main_work_horse
    raise e
  File "rq/worker.py", line 684, in main_work_horse
    self.perform_job(job, queue)
  File "rq/worker.py", line 811, in perform_job
    self.prepare_job_execution(job, heartbeat_ttl)
  File "rq/worker.py", line 725, in prepare_job_execution
    pipeline.execute()
  File "redis/client.py", line 3685, in execute
    self.shard_hint)
  File "redis/connection.py", line 1073, in get_connection
    connection.connect()
  File "redis/connection.py", line 544, in connect
    raise ConnectionError(self._error_message(e))
```

That caused horse to exit silently without setting job status to either finished or failed. The fix solves it by exiting with non=-zero code and those allowing master process to handle this as an error.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -801,7 +801,7 @@ class Worker(object):
             exc_string = self._get_safe_exception_string(
                 traceback.format_exception(*exc_info)
             )
-            self.handle_job_failure(job=job, exc_string=exc_string)
+            self.handle_job_failure(job=job, queue=queue, exc_string=exc_string)
             self.handle_exception(job, *exc_info)
             os._exit(1)
 
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -796,13 +796,10 @@ class Worker(object):
         try:
             self.main_work_horse_int(job, queue)
         except:
-            job.ended_at = utcnow()
             exc_info = sys.exc_info()
             exc_string = self._get_safe_exception_string(
                 traceback.format_exception(*exc_info)
             )
-            self.handle_job_failure(job=job, queue=queue, exc_string=exc_string)
-            self.handle_exception(job, *exc_info)
             os._exit(1)
 
 
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -701,18 +701,8 @@ class Worker(object):
         os.environ['RQ_WORKER_ID'] = self.name
         os.environ['RQ_JOB_ID'] = job.id
         if child_pid == 0:
-            try:
-                self.main_work_horse(job, queue)
-                os._exit(0) # just in case
-            except:
-                job.ended_at = utcnow()
-                exc_info = sys.exc_info()
-                exc_string = self._get_safe_exception_string(
-                    traceback.format_exception(*exc_info)
-                )
-                self.handle_job_failure(job=job, exc_string=exc_string)
-                self.handle_exception(job, *exc_info)
-                os._exit(1)
+            self.main_work_horse(job, queue)
+            os._exit(0) # just in case
         else:
             self._horse_pid = child_pid
             self.procline('Forked {0} at {1}'.format(child_pid, time.time()))
@@ -786,7 +776,7 @@ class Worker(object):
         self.monitor_work_horse(job)
         self.set_state(WorkerStatus.IDLE)
 
-    def main_work_horse(self, job, queue):
+    def main_work_horse_int(self, job, queue):
         """This is the entry point of the newly spawned work horse."""
         # After fork()'ing, always assure we are generating random sequences
         # that are different from the worker.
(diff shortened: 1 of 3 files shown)


selwin:
I think a better approach would be to put [prepare_job_execution](https://github.com/rq/rq/blob/master/rq/worker.py#L908) inside the try/except block [here](https://github.com/rq/rq/blob/master/rq/worker.py#L913). This will allow the horse to handle the error itself.

As an additional fail safe, we can also wrap [perform_job](https://github.com/rq/rq/blob/master/rq/worker.py#L787) in a try/except block and calling `os.exit(1)`, this will let the main worker handle the error in the horse.

Please also provide a test for the latter case.

rq/worker.py
```diff
@@ -572,6 +572,8 @@ def work(self, burst=False, logging_level="INFO", date_format=DEFAULT_LOGGING_DA
                         'Worker %s: found an unhandled exception, quitting...',
                         self.key, exc_info=True
                     )
+                    if self.is_horse:
+                        os._exit(1)
```
selwin, line 576: We should move `os._exit(1)` to a `try/except` block around `perform_job()` in line 789 so we catch this error much earlier.

theambient:
implemented recommended changes after te review and put whole `main_work_horse` under exception guards.

also added a test for horse to fail

tests/fixtures.py
```diff
@@ -37,6 +37,11 @@ def do_nothing():
     """The best job in the world."""
     pass
 
+def raise_exc():
+    raise Exception('raise_exc error')
+
+def raise_exc_mock():
```
selwin, line 43: Why do we need `raise_exc_mock`? Can't we just use `raise_exc()` as mock?

tests/fixtures.py
```diff
@@ -37,6 +37,11 @@ def do_nothing():
...
+    return raise_exc
```
selwin, line 44: Can't we just do this?
```python
def raise_exc():
    
```

rq/worker.py
```diff
@@ -701,7 +701,18 @@ def fork_work_horse(self, job, queue):
         os.environ['RQ_WORKER_ID'] = self.name
         os.environ['RQ_JOB_ID'] = job.id
         if child_pid == 0:
-            self.main_work_horse(job, queue)
+            try:
```
selwin, line 704: I think we should move the `try/except` block into `main_work_horse()`, it looks much cleaner that way.

tests/fixtures.py
theambient: > new_callable allows you to specify a different class, or **callable object, that will be called to create the new object**. By default AsyncMock is used for async functions and MagicMock for the rest.

from https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

otherwise `raise_exc` is called and tests fails always since it is the same as write `raise Exception`.

rq/worker.py
theambient: then I will need to make smth like `main_work_horse_int` to do proper testing and make sure `main_work_horse` always terminates gracefully and not just `perform_job`

like this

```python
    def main_work_horse_int(self, job, queue):
        """This is the entry point of the newly spawned work horse."""
        # After fork()'ing, always assure we are generating random sequences
        # that are different from the worker.
        random.seed()

        self.setup_work_horse_signals()
        self._is_horse = True
        self.log = logger
        self.perform_job(job, queue)

        # os._exit() is the way to exit from childs after a fork(), in
        # contrast to the regular sys.exit()
        os._exit(0)

    def main_work_horse(self, job, queue):
        try:
            self.main_work_horse_int(job, queue)
        except:
            job.ended_at = utcnow()
            exc_info = sys.exc_info()
            exc_string = self._get_safe_exception_string(
                traceback.format_exception(*exc_info)
            )
            self.handle_job_failure(job=job, exc_string=exc_string)
            self.handle_exception(job, *exc_info)
            os._exit(1)
```

is it ok?

theambient:
seems tests have failed because there was no local redis instance, tests are good on my laptop and fails the same way if I drop local redis. 

can I rerun tests somehow?

rq/worker.py
theambient: refactored

theambient:
@selwin I've fixed all tests, can you review PR again?

rq/worker.py
```diff
@@ -791,6 +792,17 @@ def main_work_horse(self, job, queue):
         # contrast to the regular sys.exit()
         os._exit(0)
 
+    def main_work_horse(self, job, queue):
```
selwin, line 795: I prefer not to have another method for the main work horse. Please wrap *only* the `perform_job()` part into a try/except block. If other parts like `random.seed()`, setting up signals and other parts fail, they should fail loudly because those are non job execution errors that need to be handled differently.

theambient: done

selwin:
Thanks!

theambient:
can you give me a heads up when this will be released? or when the new release will come out? have some production issues (tasks are being lost)  that can be solved by this fix.

selwin:
I'll try to release a patch version in a week.

PR #1304, opened 2020-07-17
Fix incorrect worker timeout calculation in SimpleWorker.execute_job

davidmurray:
In our systems, this bug seemed to be the cause of disappearing workers: worker keys would get a very small TTL in Redis and would eventually expire, thus mysteriously "disappearing" from dashboards.


selwin:
Thanks!

PR #1305, opened 2020-07-24
[doc] worker burst mode

zchcai:
Some more clarification for workers' burst mode.
Previously, it might give an impression that the worker will wait until the queue is empty in burst mode.

### DIFF
--- a/docs/docs/workers.md
+++ b/docs/docs/workers.md
@@ -36,9 +36,12 @@ You should use process managers like [Supervisor](/patterns/supervisor/) or
 ### Burst Mode
 
 By default, workers will start working immediately and will block and wait for
-new work when they run out of work.  Workers can also be started in _burst
+new work when they run out of work.  Workers can also be started in _burst     
 mode_ to finish all currently available work and quit as soon as all given
 queues are emptied.
+In _burst mode_, a worker will also quit if there are no available jobs
+prepared to start, even when there exists a non-empty queue. For example, if
+all remaining jobs are in _deferred_ status, block by unfinished depended job.
 
 ```console
 $ rq worker --burst high default low
--- a/docs/docs/workers.md
+++ b/docs/docs/workers.md
@@ -36,7 +36,7 @@ You should use process managers like [Supervisor](/patterns/supervisor/) or
 ### Burst Mode
 
 By default, workers will start working immediately and will block and wait for
-new work when they run out of work.  Workers can also be started in _burst     
+new work when they run out of work.  Workers can also be started in _burst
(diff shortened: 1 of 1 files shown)


zchcai:
Consider label this PR as `docs`.

docs/docs/workers.md
```diff
@@ -39,6 +39,9 @@ By default, workers will start working immediately and will block and wait for
 new work when they run out of work.  Workers can also be started in _burst
 mode_ to finish all currently available work and quit as soon as all given
 queues are emptied.
+In _burst mode_, a worker will also quit if there are no available jobs
+prepared to start, even when there exists a non-empty queue. For example, if
```
selwin, line 43: I'm ok with adding a note to clarify that deferred jobs will not cause workers in burst mode to block. But the statement is a bit misleading because jobs that are deferred are actually not yet `enqueued`, they are actually stored in `DeferredJobRegistry`.

zchcai: > jobs that are deferred are actually not yet enqueued, they are actually stored in DeferredJobRegistry.

Yes, this is the actual behavior.
I think one current misleading part is when users execute/write the code `queue.enqueue(<func_name>, args, ..., depends_on=<another_job_id>`, it is likely to assume this job is already "enqueued" in the queue and the queue should not be empty until all jobs are finished.

One scenario is that all jobs are "enqueued" by the statement `queue.enqueue()`, and then several workers start in burst mode. It is reasonable to assume the workers will finish all jobs, including those deferred ones, before they exit. However, if at first many jobs are deferred due to dependency and some workers take over all available jobs (maybe small amount at first), then the other workers will immediately exit, which is unexpected.

selwin: > I think one current misleading part is when users execute/write the code queue.enqueue(<func_name>, args, ..., depends_on=<another_job_id>, it is likely to assume this job is already "enqueued" in the queue and the queue should not be empty until all jobs are finished.

It's a reasonable assumption. I think this is the part that we need to clarify, not that `a worker will also quit if there are no available jobs prepared to start, even when there exists a non-empty queue` because this is technically not true.

Thanks for pointing this out.

zchcai: Right. If the queue is not empty, the worker will not quit in burst mode.
In detail, all deferred jobs are located in their corresponded registries first before getting into the queue.

zchcai:
I think #1306 talks about the same thing and here we close this one.
In general, `queue.enqueue()` will not immediately let one job into the queue; instead, it may go to one registry first.

PR #1306, opened 2020-07-25
[doc] notes for len(queue)

zchcai:
In general, if we enqueue one job into the queue but this job cannot be started yet, then the queue size will not enlarge.


selwin:
Thanks!

PR #1319, opened 2020-08-12
Allow retries to be set through decorator

### LINKED ISSUE #1318: Retry on @job decorator
I noticed when trying to use the new Retry function that it is not supported with the @job-decorator. Is there any plans to support this, or are there technical limitations to this issue? If neither I'll try to create a PR that adds the retry functionality to the job-decorator.

nerok:
This pull request will close #1318. It allows retrying of jobs to be sat through the @job-decorator.


selwin:
Thanks!

Issue #1310, opened 2020-07-30
Scheduler shouldn't recreate a Redis object


info-prod:
Hello,

Since the changes introduced with MR #1302 , the scheduler will re-create a Redis object from scratch regardless of what it receives as connection. This is undesirable as everything else in RQ will work just fine with a custom class (provided it implements the same interface as py-redis).

My use case for a custom "Redis" object: a proxy which balances the requests between a master and two replicas (depending on whether write access is required). 

Another potential use case: for unit test it would be useful to pass FakeRedis (as suggested in the doc) or another custom "dummy" class instead of a true Redis object.

Thanks for reading and have a nice day.

selwin:
Yes, we can use `connection.__class__` to instantiate the connection. Please open a PR for this.

PR #1327, opened 2020-08-25
fix(scheduler): correctly restore ssl connection after forking

pacahon:
`ssl` attribute is lost after forking since `redis-py` doesn't store `ssl` attribute in connection kwargs https://github.com/andymccurdy/redis-py/blob/5a6a50b50db26e4b1c4b7a6cb9b2a2caa8ba5e8b/redis/client.py#L783
Apparently this PR also fixes https://github.com/rq/rq/issues/1310
Sorry, no Idea how to write unit test for ssl connection, just tested it manually.


selwin:
Thanks!

PR #1348, opened 2020-09-28
feat: avoided "zombie" processes after killing work horse

theambient:
by setting work horse process group and killing the whole group.
might be useful if task runs other subprocesses.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -33,7 +33,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install redis==${{ matrix.redis-py-version }}
-        pip install pytest pytest-cov sentry-sdk codecov mock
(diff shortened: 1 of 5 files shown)


theambient:
Where can I find out failed test (all tests) logs so that I can find out what failed ? in my local env (mac) all tests are passed (and some of them are skipped)

selwin:
You can see the failing checks here, in this case it’s codecov that breaks. I wouldn’t worry about this one since codecov doesn’t properly track codes run by the work horse.

898900D6-D200-4653-BC40-5D88DBDFAD21

Would it be hard to create a test that simulates a job that creates other subprocesses and check whether all of them are killed along with the horse?

theambient:
@selwin sure, i will make such test

selwin:
Thanks!

PR #1349, opened 2020-09-28
feat: added job heartbeat to track whether job is actually executing

theambient:
heartbeat might be needed in cases when worker was hardkilled or the whole VM/docker was forcibly rebooted.

### DIFF
--- a/docs/docs/jobs.md
+++ b/docs/docs/jobs.md
@@ -121,7 +121,7 @@ Some interesting job attributes include:
 * `job.started_at`
 * `job.ended_at`
 * `job.exc_info` stores exception information if job doesn't finish successfully.
-* `job.get_heartbeat()` returns last heartbeat of the job indicating last time the job was executing. Can be used to determine if a worker was killed forcely and to mark the job as `failed`.
+* `job.last_heartbeat` the latest timestamp that's periodically updated when the job is executing. Can be used to determine if the job is still active.
 
 If you want to efficiently fetch a large number of jobs, use `Job.fetch_many()`.
 
--- a/docs/docs/jobs.md
+++ b/docs/docs/jobs.md
@@ -121,6 +121,7 @@ Some interesting job attributes include:
 * `job.started_at`
 * `job.ended_at`
 * `job.exc_info` stores exception information if job doesn't finish successfully.
+* `job.get_heartbeat()` returns last heartbeat of the job indicating last time the job was executing. Can be used to determine if a worker was killed forcely and to mark the job as `failed`.
 
 If you want to efficiently fetch a large number of jobs, use `Job.fetch_many()`.
 
--- a/rq/job.py
+++ b/rq/job.py
@@ -351,7 +351,7 @@ class Job(object):
         # retry_intervals is a list of int e.g [60, 120, 240]
         self.retry_intervals = None
         self.redis_server_version = None
-        self.heartbeat = None
+        self.last_heartbeat = None
 
     def __repr__(self):  # noqa  # pragma: no cover
         return '{0}({1!r}, enqueued_at={2!r})'.format(self.__class__.__name__,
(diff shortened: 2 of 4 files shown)


rq/worker.py
```diff
@@ -729,6 +730,8 @@ def monitor_work_horse(self, job, queue):
                     self.wait_for_horse()
                     break
 
+                job.set_heartbeat(utcnow())
```
selwin, line 733: The call to job and worker heartbeat should be pipelined.

rq/job.py
```diff
@@ -351,6 +351,7 @@ def __init__(self, id=None, connection=None, serializer=None):
         # retry_intervals is a list of int e.g [60, 120, 240]
         self.retry_intervals = None
         self.redis_server_version = None
+        self.heartbeat = None
```
selwin, line 354: This should be renamed to `last_heartbeat`

rq/job.py
```diff
@@ -384,6 +385,21 @@ def set_id(self, value):
             raise TypeError('id must be a string, not {0}'.format(type(value)))
         self._id = value
 
+    def set_heartbeat(self, heartbeat, pipeline=None):
```
selwin, line 388: Can we also rename this to `heartbeat()` so that it's consistent with `worker.heartbeat()`?

rq/job.py
selwin: `job.last_heartbeat` should also be added to the docs here: https://github.com/rq/rq/blob/master/docs/docs/jobs.md

rq/job.py
```diff
@@ -384,6 +385,21 @@ def set_id(self, value):
...
+
+    def get_heartbeat(self, refresh=True):
+        if refresh:
+            raw = self.connection.hget(self.key, 'heartbeat')
+            if raw:
+                self.last_heartbeat = str_to_date(raw)
+            else:
+                self.last_heartbeat = None
+
+        return self.last_heartbeat
```
selwin, line 401: I think we can remove this method, it's unnecessary. When we fetch `job` from Redis, you can already call `job.last_heartbeat`. If you want to refresh job metadata from, call `job.refresh()`.

rq/job.py
```diff
@@ -530,6 +547,7 @@ def to_dict(self, include_meta=True):
             'data': zlib.compress(self.data),
             'started_at': utcformat(self.started_at) if self.started_at else '',
             'ended_at': utcformat(self.ended_at) if self.ended_at else '',
+            'heartbeat': utcformat(self.last_heartbeat) if self.last_heartbeat else '',
```
selwin, line 550: Mind changing this to `last_heartbeat` to keep things consistent?

rq/worker.py
```diff
@@ -713,6 +713,7 @@ def monitor_work_horse(self, job, queue):
 
         ret_val = None
         job.started_at = utcnow()
+        job.heartbeat(job.started_at)
```
selwin, line 716: The first heartbeat should be located in [prepare_job_execution()](https://github.com/rq/rq/blob/master/rq/worker.py#L825) so it can be pipelined with job status changes etc.

theambient:
done

docs/docs/jobs.md
```diff
@@ -121,6 +121,7 @@ Some interesting job attributes include:
 * `job.started_at`
 * `job.ended_at`
 * `job.exc_info` stores exception information if job doesn't finish successfully.
+* `job.get_heartbeat()` returns last heartbeat of the job indicating last time the job was executing. Can be used to determine if a worker was killed forcely and to mark the job as `failed`.
```
selwin, line 124: This needs to be updated to `job.last_heartbeat` - the latest timestamp that's periodically updated when the job is executing. Can be used to determine if the job is still active.

rq/worker.py
selwin: This is no longer needed since the first heartbeat is already set in `prepare_job_execution()`.

docs/docs/jobs.md
theambient: done

rq/worker.py
selwin: Don't forget to remove this line

theambient: thx, done, sorry for being messy

selwin:
Thanks!

PR #1350, opened 2020-09-28
feat: kill horse if task was cancelled/deleted

theambient:
useful for long tasks with big timeout value to prevent worker from being occupied by a task that is of no interest anymore

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -723,6 +723,13 @@ class Worker(object):
                 # Send a heartbeat to keep the worker alive.
                 self.heartbeat(self.job_monitoring_interval + 60)
 
+                job_status = job.get_status()
+                if job_status is None:  # Job was cancelled/deleted
+                    self.log.info('killing running job (the horse) since it was cancelled')
+                    self.kill_horse()
+                    self.wait_for_horse()
+                    break
+
                 # Kill the job from this side if something is really wrong (interpreter lock/etc).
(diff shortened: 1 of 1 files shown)


selwin:
I don’t think this is a good solution. I already have a branch working locally that allows you to cancel an already running job midway.

theambient:
what is this branch ?

selwin:
https://github.com/rq/rq/tree/pubsub

theambient:
Thanks, I believe these two ways of killing running tasks that are no longer required can live alongside. one is active and one is passive.

up to you to decide whether to include or not just to let you know that in my case It works pretty well (and might work well for the others) just because there is no need to know which worker executes the task (and there is no way to find it out).

selwin:
>  just because there is no need to know which worker executes the task (and there is no way to find it out)

This is coming soon, hopefully. I want to introduce a `job.worker_name` variable which will let you know which worker is executing a particular job. From there, we can do `stop_job(job.id)` which will send the kill-horse command to the worker.

theambient:
> This is coming soon, hopefully. I want to introduce a job.worker_name variable which will let you know which worker is executing a particular job. From there, we can do stop_job(job.id) which will send the kill-horse command to the worker.

That would be amazing. As a kind request would be nice to kill the job on `job.delete`/`job.cancel`.

As for `job.worker_name` - I already do it actually, If you want i can prepare a PR.

selwin:
If you already have `job.worker_name` handy, yes please prepare a PR for this.

PR #1351, opened 2020-09-28
Add capability to run jobs on failure, add post-run callbacks, add Job monitoring to monitor status updates, add type hints

FakeEmperor:
Greetings,

I'd like to thank you for your hard work on this project! 
We were frustrated with many queuing frameworks in Python like Celery and Airflow, that didn't quite satisfied our requirements.
We came across this project and were really happy using it so far.

The simplicity is paramount and robustness is great for simple tasks and pipelines! 
After a while, we had a need to add a bit more to what is already there:

## New features

- Callbacks after the Job has finished: we have a case where we might need to start another Job based off what is the result (or an exception type) is.
  
  It can be used like this:

  ```python
  def on_error_callback(job: "Job", error: "BaseException"):
    pass

  def on_success_callback(job: "Job", result: "Any"):
    pass
  
  queue.enqueue(..., on_failure=on_error_callback, on_success=on_success_callback)
  ```

- Cancellation during Job run: a job may be cancelled while it's running. It sets a metadata field "cancelled" to True and introduces new status: `CANCELLED`
  which is only set if the job is cancelled.
  Cancellation during Job run is left outside of this PR and must be implemented in a custom Worker class.
  
  
- Run conditions: a job may run not only when its dependencies are `FINISHED`, but on any *terminal* status, like `CANCELLED` or `FAILED`.
  Run conditions allow defining static pipelines, accounting for any failures and cancellation. 
  For example, where an external cleanup is required, for example.
  
  An example:
  
  ```
  parent = queue.enqueue(...)
  on_success = queue.enqueue(dependency=parent)
  on_failure = queue.enqueue(dependency=parent, run_when=JobStatus.FAILED)
  
  ```

- Job status monitoring:
  You can use job `JobEventQueue` class to monitor JobStatus transitions in future-like interface:
  
  ```python
  with JobEventQueue(job) as events:
    events.receive(only=(JobStatus.CANCELLED,)).result(timeout=1)
  ```
  
  It works pretty well with live cancellation events on Workers.

- Handy aliases and functions to return a tree of related job: `get_children`, `get_child_job_ids`, `get_parents`, `get_parent_job_ids`.
  `get_parents` partially overlaps with `fetch_dependencies` method, but we first added `get_children` and then added the rest for symmetry.

We've also added type stubs, which we can scrub if you do have concerns about compatibility, 
cluttering or added cost of library support.

## Changes

- Added type hints
- Added dev environment `docker-compose.yml` and dockerfile
- Small fixes to tests and customization of Redis endpoint with `$RQ_TESTS_REDIS_HOST`
- Reduced code duplication in `registry.py`


selwin:
Thanks for this PR. Lots of useful ideas here!

I'd like to get them merged in, but ideally they'd have to be done via a few independent PRs, so I can comment on their implementation details (I'm rather picky about how things are implemented). I'll just comment on a few ideas proposed in this PR.

* Success/failure callbacks: if you can make a separate PR for this feature, it would be great. 

* Cancelation: this is something I'm already working on and it will be implemented as `job.status` as opposed to being a separate field.

In addition to simply moving jobs to `CanceledJobRegistry`, canceling currently running jobs are also possible because workers will listen for external commands via Redis pubsub. Hopefully I can finish this up in a week or two. 

* Job dependency run conditions. Another nice idea, but I'd change the API to something like:
 `queue.enqueue(my_func, depends_on=Dependency(job, run_when=[JobStatus.FAILED, JobStatus.CANCELED])`

* Job status monitoring: I plan on using Redis Streams instead of pubsub for this since it provides efficient logging capabilities in addition to realtime monitoring.

* Typing: since Python 3.5 is EOL, I'm happy to include type hints.

* I'm not a fan of using `"` to replace single quotes though. Single quotes are cleaner and easier to read.

* Happy to pull in docker related stuff and other test improvements to this project

FakeEmperor:
Thanks for the comment, @selwin! I'll split the code into separate PRs then. Agreed that it'll be easier to pull them in and keep the conversation local.

Great news about cancellation! I was a bit hesitant to add cancellation to workers since I felt that it might disrupt someone's overrides of a Worker class, but it's great that it'll be in the core soon :)

> Job dependency run conditions. Another nice idea, but I'd change the API to something like:
queue.enqueue(my_func, depends_on=Dependency(job, run_when=[JobStatus.FAILED, JobStatus.CANCELED])

Sure, that was my initial idea, but I thought that a more localized patch would be approved easier :) I'll make the change to allow flexible status-based dependencies.

> I'm not a fan of using " to replace single quotes though. Single quotes are cleaner and easier to read.

I don't have any particular preference, so I can conform with that as well.

Thanks again for the comment! :) With that said, I'm going to close this PR and make a few separate ones in upcoming few days.

freshsun:
@FakeEmperor do we have any progress on success/failure callbacks PR?  that's a quite useful feature, thanks!

FakeEmperor:
> 
> 
> @FakeEmperor do we have any progress on success/failure callbacks PR? that's a quite useful feature, thanks!

Hey @freshsun! Sorry, I haven't been able to split my PR just yet -- lot's of stuff at work, as usual :/ The design of callbacks was a bit upgraded, but hopefully for the better! I've added explicit support for _chainable callbacks_ and _callback jobs_. We've leveraged it to create very fluid pipelines with easy to spot errors on failures, I hope you'll like it once a PR will hit the master.
I'll try to tackle it ASAP, but no promises ;)

PR #1355, opened 2020-10-02
scheduler: now operates with chunks of jobs

### LINKED ISSUE #1354: RQScheduler production running bug
We run RQ in production and have a bunch of scheduled jobs (~10M) in one queue. Also we have ~250 workers that works on this queue. We have starting all these workers with `--with-scheduler` parameter, and have relied on RQ documentation, that says that in one moment only one worker can run with scheduler.
But after working in this mode several days, we noticed, that amount of jobs in our queue continuously grows. And count of jobs in ScheduledJobRegistry didn't reduce. We started to investigate source code of RQ and tried to identify reason of the problem.
And we found this code:
https://github.com/rq/rq/blob/55dce19a95dc043bae1691ac3bda13c33550a0ec/rq/scheduler.py#L89
So in this line, one instance of RQScheduler acquire locks in Redis. It sets key with pid (also insecure if you running multiple workers with schedulers on different servers and pids can coincide) and most important that it sets expire time to 5 seconds.
After acquiring locks, scheduler calls `enqueue_scheduled_jobs` function and it is blocking call (important).
After that call, scheduler calls `heartbeat` function that renew lock that scheduler acquires.

And cause we have many scheduled jobs, it's enqueuing can take more than 5 seconds, and lock key expired from Redis. One of the other workers with scheduler now can acquire lock and also start to `enqueue_scheduled_jobs`, so jobs duplicates and if this worker dies, it also don't removed from ScheduledJobRegistry and process repeats again and again.

nikkonrom:
Fixes #1354 
It might not be complete fix of this issue but prevent (in most cases) bug, reducing `enqueue_scheduled_jobs` function executing time to less than 5 seconds

### DIFF
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -286,15 +286,11 @@ class ScheduledJobRegistry(BaseRegistry):
         score = timestamp if timestamp is not None else current_timestamp()
         return connection.zremrangebyscore(self.key, 0, score)
 
-    def get_jobs_to_schedule(self, timestamp=None, chunk_size=None):
+    def get_jobs_to_schedule(self, timestamp=None, chunk_size=1000):
         """Remove jobs whose timestamp is in the past from registry."""
         score = timestamp if timestamp is not None else current_timestamp()
-        if not chunk_size:
-            return [as_text(job_id) for job_id in
-                    self.connection.zrangebyscore(self.key, 0, score)]
-        else:
-            return [as_text(job_id) for job_id in
-                    self.connection.zrangebyscore(self.key, 0, score, start=0, num=chunk_size)]
+        return [as_text(job_id) for job_id in
+                self.connection.zrangebyscore(self.key, 0, score, start=0, num=chunk_size)]
 
     def get_scheduled_time(self, job_or_id):
         """Returns datetime (UTC) at which job is scheduled to be enqueued"""
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -38,8 +38,8 @@ class BaseRegistry(object):
 
     def __eq__(self, other):
         return (
-            self.name == other.name and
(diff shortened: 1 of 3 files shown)


rq/registry.py
```diff
@@ -286,11 +286,15 @@ def remove_jobs(self, timestamp=None, pipeline=None):
         score = timestamp if timestamp is not None else current_timestamp()
         return connection.zremrangebyscore(self.key, 0, score)
 
-    def get_jobs_to_schedule(self, timestamp=None):
+    def get_jobs_to_schedule(self, timestamp=None, chunk_size=None):
```
selwin, line 289: Let’s just set `chunk_size` to default to 1000 so we don’t have to deal with two different code paths.

nikkonrom: LGTM, thanks

nikkonrom:
I still don't know why coverage is still lower, than it was before commit. Regardless of written test for changed function. Please tell me if I need to change something to pass check (or merge if this don't matter and everything else is ok)

tests/test_scheduler.py
```diff
@@ -35,6 +35,21 @@ def test_get_jobs_to_enqueue(self):
         self.assertEqual(registry.get_jobs_to_enqueue(timestamp + 20),
                          ['foo', 'bar'])
 
+    def test_get_jobs_to_schedule_with_chunk_size(self):
+        """Max amount of jobs returns by get_jobs_to_schedule() equal to chunk_size"""
+        queue = Queue(connection=self.testconn)
+        registry = ScheduledJobRegistry(queue=queue)
+        timestamp = current_timestamp()
+        chunk_size = 1000
```
selwin, line 43: To make the test lighter to run, let's just change `chunk_size` to 5 or 10.

nikkonrom: Thanks, done

selwin:
Thanks!

PR #1366, opened 2020-10-22
Propagate logging settings from worker to scheduler

### LINKED ISSUE #1365: Propagate logging level from worker to scheduler
I'm running `rq worker --with-scheduler --quiet`, and it properly quiets the worker logs, but the scheduler remains set to `INFO`, where it prints out `Trying to acquire locks for default` every ~15 minutes. Could we make the `-q/--quiet` propagate to the scheduler? Currently it looks like the scheduler logger is not configurable except by accessing the logger directly.
If I find some time, I'll try to make a PR for this.

christopher-dG:
Closes #1365 

I can confirm that I get the expected (lack of) output when running in quiet mode.

### DIFF
--- a/rq/scheduler.py
+++ b/rq/scheduler.py
@@ -7,24 +7,17 @@ import traceback
 from datetime import datetime
 from multiprocessing import Process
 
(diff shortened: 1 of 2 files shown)


selwin:
Thanks!

Issue #1164, opened 2019-11-29
Jobs are marked as started when docker container is killed


pkajaba:
When I am running rq worker in docker container and container is kill with docker kill, worker seems to be registered on redis as still working. 

Is there any way how to mark started job as failed even without waiting it will reach it's timeout? I tried to call `clean_registries` on worker start, but it does not seem to be doing any trick.

Thanks!

selwin:
Not unless we patch `clean_registries()` to also check for worker PIDs to determine if a job is still being worked on. Would love to have a PR for this.

pkajaba:
would we have to introduce worker_id:job_id mapping or it can be done without it?

selwin:
We should store in Redis as part of `Job`'s data. Should be something like `job.worker_pid`

levkk:
I think this is relevant for jobs that get lost when a worker is abruptly terminated. A worker can be killed by Docker at any time without warning, so if the worker was running a long job during that time, it wouldn't register as failed and wouldn't retry I think.

selwin:
RQ has a process that sweeps started jobs and checks whether they’re still alive after their timeout has ended. Jobs that are harvested this way will be marked as failed and retried (if set).

noahg2:
I wonder if there is any update on when this issue might be resolved (e.g. by #1372), or if some additional contributions might be needed. Getting a fix for this in place would represent a significant quality-of-life improvement for one of my projects, since container eviction can currently result in users' long-running tasks spending a significant amount of time in a "zombie state" before the failure is detected and rectified.

selwin:
If we want to detect long running jobs abnormally terminating early, we'll need to have a way to check whether the worker running the job is still alive.

One way to do this would be to check the worker running the job's last heartbeat. I'd be happy to accept a PR for this.

How I think we should do this:
1. Create a `job.worker_is_still_alive()` that returns whether the worker running the job is still active
2. Introduce a command line option to `rq worker` that tells it to also check for long running job's status when running [maintenance tasks](https://github.com/rq/rq/blob/master/rq/worker.py#L498)
3. If a job is dead, move it to FailedJobRegistry

PR #1372, opened 2020-11-06
Cleanup zombie worker leftovers as part of StartedJobRegistry's cleanup()

rauchy:
When workers die ungracefully, jobs without a timeout will remain in the `StartedJobRegistry` forever.
This PR cleans those up as part of the routine `cleanup()` call in `StartedJobRegistry`. 

This should solve https://github.com/rq/rq/issues/1164

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -753,8 +753,11 @@ class Worker(object):
             self.procline('Forked {0} at {1}'.format(child_pid, time.time()))
 
     def get_heartbeat_ttl(self, job, elapsed_execution_time):
-        remaining_execution_time = job.timeout - elapsed_execution_time
-        return min(remaining_execution_time, self.job_monitoring_interval) + 60
+        if job.timeout and job.timeout > 0:
+            remaining_execution_time = job.timeout - elapsed_execution_time
+            return min(remaining_execution_time, self.job_monitoring_interval) + 60
+        else:
+            return self.job_monitoring_interval + 60
 
     def monitor_work_horse(self, job, queue):
         """The worker will monitor the work horse and make sure that it
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -779,11 +779,8 @@ class Worker(object):
 
                 with self.connection.pipeline() as pipeline:
                     self.heartbeat(timeout, pipeline=pipeline)
+                    queue.started_job_registry.heartbeat(job, timeout, pipeline=pipeline)
                     job.heartbeat(utcnow(), pipeline=pipeline)
-                    pipeline.zadd(
-                        queue.started_job_registry.heartbeats_key,
-                        {job.id: current_timestamp()}
-                    )
                     pipeline.execute()
 
             except OSError as e:
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -40,7 +40,7 @@ from .registry import FailedJobRegistry, StartedJobRegistry, clean_registries
 from .scheduler import RQScheduler
 from .suspension import is_suspended
 from .timeouts import JobTimeoutException, HorseMonitorTimeoutException, UnixSignalDeathPenalty
-from .utils import (backend_class, current_timestamp, ensure_list, enum, get_version,
+from .utils import (backend_class, ensure_list, enum, get_version,
                     make_colorizer, utcformat, utcnow, utcparse)
 from .version import VERSION
 from .worker_registration import clean_worker_registry, get_keys
@@ -760,7 +760,6 @@ class Worker(object):
 
         ret_val = None
         job.started_at = utcnow()
-        timeout = self.job_monitoring_interval + 60
         while True:
             try:
                 with UnixSignalDeathPenalty(self.job_monitoring_interval, HorseMonitorTimeoutException):
@@ -772,14 +771,13 @@ class Worker(object):
 
                 # Kill the job from this side if something is really wrong (interpreter lock/etc).
                 if job.timeout != -1 and (utcnow() - job.started_at).total_seconds() > (job.timeout + 60):
-                    self.heartbeat(timeout)
+                    self.heartbeat(self.job_monitoring_interval + 60)
                     self.kill_horse()
                     self.wait_for_horse()
                     break
 
                 with self.connection.pipeline() as pipeline:
-                    self.heartbeat(timeout, pipeline=pipeline)
-                    queue.started_job_registry.heartbeat(job, timeout, pipeline=pipeline)
+                    self.heartbeat(self.job_monitoring_interval + 60, pipeline=pipeline)
                     job.heartbeat(utcnow(), pipeline=pipeline)
                     pipeline.execute()
 
@@ -881,7 +879,7 @@ class Worker(object):
             self.heartbeat(heartbeat_ttl, pipeline=pipeline)
             registry = StartedJobRegistry(job.origin, self.connection,
                                           job_class=self.job_class)
-            registry.add(job, timeout, heartbeat_ttl=heartbeat_ttl, pipeline=pipeline)
+            registry.add(job, timeout, pipeline=pipeline)
             job.prepare_for_execution(self.name, pipeline=pipeline)
             pipeline.execute()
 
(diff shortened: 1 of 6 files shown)


selwin:
Hey there, this is an interesting addition, but I think this needs a little bit of tweaking. This operation is potentially expensive to run on systems with a large number of workers running at the same time (hundreds or thousands), since it would verify that all workers of jobs in `StartedJobRegistry` are still alive.

There are a few ways we can approach this.

## Approach 1
1. Add another method `registry.get_job_ids_without_expiry()` that returns jobs with infinite timeouts
2. Verify that the workers are still alive

## Approach 2
[This PR](https://github.com/rq/rq/pull/1349) adds `job.last_heartbeat`. As part of the heartbeat process, we can actually track when a job last reported for duty, check the ones who haven't reported as active in the last few minutes and check whether their respective workers are still alive.

Note that I've also been working on adding `job.worker_name` in this PR https://github.com/rq/rq/pull/1375 so we can use this instead.

rauchy:
Thanks @selwin. Approach 2 (+ `job.worker_name`) feels to me like the way to go. I just have a couple of questions:

1. I'm assuming you meant "As part of the **cleanup** process" instead of "As part of the **heartbeat** process"?
2. What would be a good interval to declare a job as "not actively worked on"? I'm thinking it's something like `job_monitoring_interval + 60`, but that one is worker configuration and kinda feels awkward to pass down to the job, don't you think?

selwin:
> I'm assuming you meant "As part of the cleanup process" instead of "As part of the heartbeat process"?

No, "as part of the heartbeat process" is correct, it's [this line here](https://github.com/rq/rq/blob/master/rq/worker.py#L766).  We store the timeout in a sorted set, and the cleanup process periodically check whether any job has "expired".

After we implement this, the cleanup process can rely on this sorted set as part of it's cleaning process (this is better than checking StartedJobRegistry for expired jobs). In most cases we'll be able to detect dead jobs sooner.

> What would be a good interval to declare a job as "not actively worked on"? I'm thinking it's something like job_monitoring_interval + 60, but that one is worker configuration and kinda feels awkward to pass down to the job, don't you think?

I think `job_monitoring_interval + <some interval>` is fine.

rauchy:
Ok, we are talking about the same thing. I meant "look for out of date heartbeats as part of the cleanup process".

This does lead to my second question - how can we determine the right interval when running in the scope of the cleanup process? Passing down the monitoring interval from the worker to the cleanup method seems to be the easiest way, but makes the registry APIs uneven in terms of arguments.

selwin:
> This does lead to my second question - how can we determine the right interval when running in the scope of the cleanup process? Passing down the monitoring interval from the worker to the cleanup method seems to be the easiest way, but makes the registry APIs uneven in terms of arguments.

The expiry is set during the heartbeat process, which is now + 60 seconds. The cleanup process simply looks for heartbeats whose value is lower than the current timestamp and moves those jobs to `FailedJobRegistry`.

rauchy:
Makes sense @selwin. Now using a sorted set for heartbeats per registry.

selwin:
We can actually simplify this PR by quite a bit.

`StartedJobRegistry` already maps a job to it's expiration time [here](https://github.com/rq/rq/blob/master/rq/worker.py#L859). We just need to change:
1. The `ttl` argument from job timeout to heartbeat timeout.
2. On every `job.heartbeat`, the ttl gets extended

I think that's pretty much the changes we need to do.

rauchy:
Oh that makes total sense. I was working under the false assumption that the old non-heartbeats scores had to live side-by-side with heartbeats, but heartbeats do make a better indicator for a job's WIP status. Can you have another look?

rq/worker.py
```diff
@@ -756,13 +757,14 @@ def monitor_work_horse(self, job, queue):
...
-                    self.heartbeat(self.job_monitoring_interval + 60)
+                    self.heartbeat(heartbeat_ttl)
                     self.kill_horse()
                     self.wait_for_horse()
                     break
 
                 with self.connection.pipeline() as pipeline:
-                    self.heartbeat(self.job_monitoring_interval + 60, pipeline=pipeline)
+                    self.heartbeat(heartbeat_ttl, pipeline=pipeline)
+                    queue.started_job_registry.add(job, heartbeat_ttl, pipeline=pipeline)
```
selwin, line 767: Can we move this to `job.heartbeat()`? I think it makes more sense that way.

selwin:
Two more things that we need to take into account:

1. The heartbeat should be no later than when the job is supposed to timeout (plus some buffer time). So for example if `job.timeout` is 65 seconds and `job_monitoring_interval` is 30 seconds (job is checked every 30 seconds), the `heartbeat_ttl`s that are sent should be:
* 90 (30 + 60) at the beginning. `30` is the job monitoring interval, 60 is the buffer time
* 90 (30 + 60) at the 30th second.
* 65 (5 + 60) at the 60th second. `5` is the remaining job execution time.

2. `SimpleWorker` doesn't have a main worker thread that monitors horse's execution so we can't use the variable TTLs sent by the regular `Worker` class.

To solve the two issues, I think we should add a `worker.get_heartbeat_ttl(job, elapsed_execution_time)` that gets called whenever job heartbeat is about to be sent. This way, we can:
1. Easily test the job heartbeat TTL calculation
2. `SimpleWorker` can override this method to always return TTLs as per the original implementation.

rq/worker.py
```diff
@@ -737,6 +737,10 @@ def fork_work_horse(self, job, queue):
             self._horse_pid = child_pid
             self.procline('Forked {0} at {1}'.format(child_pid, time.time()))
 
+    def get_heartbeat_ttl(self, job, elapsed_execution_time):
```
selwin, line 740: Instead of passing `elapsed_execution_time` around, I think it makes sense for the worker to keep track of the start of job execution time. Can we make this change?

selwin:
@rauchy sorry it took so long, I missed this PR. The approach is already good, but I made a comment [here](https://github.com/rq/rq/pull/1372/files#r564136314). It would be good if we can get this addressed.

This is a PR I'd like to pull in.

rauchy:
@selwin whoops, also sorry it took me so long 🙊

Is [this](https://github.com/rq/rq/pull/1372/commits/ae7ee9134ca5d712802d48a37f4f844bfab94622) what you had in mind regarding tracking worker execution time?

rq/worker.py
```diff
@@ -861,25 +878,20 @@ def setup_work_horse_signals(self):
...
-        else:
-            timeout = job.timeout or 180
-
-        if heartbeat_ttl is None:
-            heartbeat_ttl = self.job_monitoring_interval + 60
 
         with self.connection.pipeline() as pipeline:
             self.set_state(WorkerStatus.BUSY, pipeline=pipeline)
             self.set_current_job_id(job.id, pipeline=pipeline)
+            self.set_current_job_working_time(0, pipeline=pipeline)
```
selwin, line 889: I think this needs to be moved to `monitor_work_horse()` after job has finished so this will be set to zero after job execution finishes.

rq/worker.py
```diff
@@ -374,6 +375,11 @@ def _get_state(self):
 
     state = property(_get_state, _set_state)
 
+    def set_current_job_working_time(self, current_job_working_time, pipeline=None):
```
selwin, line 378: I initially thought that this can be a simple variable tracked by the worker itself. Thanks for persisting this in Redis, this is also a useful information to have.

rq/worker.py
rauchy: Good point. I added it to `monitor_work_horse`, but I also kept it here in case a work horse dies (due to an `OSError` for example) without resetting it. WDYT?

rq/worker.py
rauchy: 👍

rauchy:
@selwin &lt;GentlePing /&gt;

selwin:
Sorry I missed this. Thanks!

michaelbrooks:
Turns out this change broke the public interface for Workers when it removed the heartbeat_ttl argument from Worker.prepare_job_execution. https://github.com/michaelbrooks/rq-win/pull/11

Issue #1357, opened 2020-10-08
Job picked from queue with default serializer while worker has json serializer


tvcuyck:
I configured the worker and queue with `serializer=json` but when the job is picked from the queue it tries to de-serialize with pickle and I get pickle errors.

I added some logging: At startup you can see the json serializer, but when the job is processed, the queue and job have a default serializer.

```
Starting Task Worker (pid: 13882)
-- Queue serializer <module 'json' from '/usr/lib/python3.8/json/__init__.py'>
-- Worker serializer <module 'json' from '/usr/lib/python3.8/json/__init__.py'>
16:04:47 Registering birth of worker python-consumer
16:04:47 Worker rq:worker:python-consumer: started, version 1.5.2
16:04:47 *** Listening on queue...
16:04:47 Sent heartbeat to prevent worker timeout. Next one should arrive within 480 seconds.
16:04:47 *** Listening on queue...
16:04:47 Sent heartbeat to prevent worker timeout. Next one should arrive within 480 seconds.
-- Queue serializer <class 'rq.serializers.DefaultSerializer'>
-- Job serializer <class 'rq.serializers.DefaultSerializer'>
16:04:56 queue: Export (52bb0e76-4387-4e46-8fab-f462ce54914c)
16:04:56 Sent heartbeat to prevent worker timeout. Next one should arrive within 480 seconds.
16:04:56 Sent heartbeat to prevent worker timeout. Next one should arrive within 90 seconds.
-- Job serializer <class 'rq.serializers.DefaultSerializer'>
16:04:56 Handling failed execution of job 52bb0e76-4387-4e46-8fab-f462ce54914c
-- Job serializer <class 'rq.serializers.DefaultSerializer'>
16:04:56 Sent heartbeat to prevent worker timeout. Next one should arrive within 480 seconds.
16:04:56 Sent heartbeat to prevent worker timeout. Next one should arrive within 480 seconds.
16:04:56 *** Listening on queue...
16:04:56 Sent heartbeat to prevent worker timeout. Next one should arrive within 480 seconds. 
```

I could fix it by passing the `serializer=self.serializer` form the worker to the queue and job via the `dequeue_any` class method in `dequeue_job_and-maintain_ttl`.

Is this a bug with the json serializer, or was my setup wrong?

selwin:
Are you initializing `Worker` with `serializer=json`? The worker CLI also needs to be updated to accept `serializer` input, I just realized this myself since I usually just run with the default pickle serializer.

[...]

PR #1381, opened 2020-11-18
Fix custom serializer in job fetches

JackBoreczky:
Adds `serializer` as an argument to the `fetch_many` and `dequeue_any` methods. Also adds `self.serializer` as an argument to `fetch` calls when fetching jobs. This ensures that the worker will correctly maintain the custom serializer when pulling jobs off a queue created with a custom serializer. Likely addresses issue in #1357. Encountered and solved with @nathanielweinman

### DIFF
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -46,6 +46,15 @@ class CustomJob(Job):
 class CustomQueue(Queue):
     pass
 
+class CustomSerializer():
+    @staticmethod
+    def dumps(*args, **kwargs):
+        return json.dumps(*args, **kwargs).encode('utf-8')
+
+    @staticmethod
+    def loads(s, *args, **kwargs):
+        return json.loads(s.decode('utf-8'), *args, **kwargs)
+
 
 class TestWorker(RQTestCase):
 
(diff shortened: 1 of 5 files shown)


selwin:
Thanks for the PR! Do you mind creating a test case for this? Simply creating a job using `json` serializer and calling `worker.work(burst=True)` would be sufficient, similar to https://github.com/rq/rq/blob/master/tests/test_worker.py#L110

JackBoreczky:
Hey @selwin! I've added that test but with a custom serializer. Passing `serializer=json` directly causes a `TypeError: a bytes-like object is required, not 'str'` error. The json dumps and load methods differ from the pickle ones in that they use str types where pickle uses bytes. The workaround solution for this I've been using is to create a custom serializer class which makes the correct calls to `encode` and `decode` to convert to and from bytes for use with json.

tests/test_worker.py
```diff
@@ -46,6 +46,15 @@ class CustomJob(Job):
 class CustomQueue(Queue):
     pass
 
+class CustomSerializer():
```
selwin, line 49: Can we rename this to `JSONSerializer` and move this to `serializers.py`? We can then document this as the recommended way to use JSON as serialization method.

JackBoreczky: Perfect, updated with this change. Should the edits to the doc be made in this PR?

selwin:
I think we can document this feature once `--serializer` flag is added to worker cli. Once this feature flag is added, the feature is pretty much complete since you can now both enqueue and execute JSON serialized jobs out of the box.

Do you mind making a PR for that too @JackBoreczky?

TsimafeiTsykunou:
image
Because of that changes worker doesn't start..

Issue #1346, opened 2020-09-25
Regression: SSLConnection `__init__() got an unexpected keyword argument 'ssl'`


Fabma:
Version `==1.5.2`

With rq version `==1.5.2` I am running into an  `__init__() got an unexpected keyword argument 'ssl'`.
(with rq version `==1.5.1` the error does not appear)

```
Traceback (most recent call last):
  File "<>/site-packages/redis/connection.py", line 1185, in get_connection
    connection = self._available_connections.pop()
IndexError: pop from empty list

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<>/site-packages/rq/worker.py", line 931, in perform_job
    self.prepare_job_execution(job, heartbeat_ttl)
  File "<>/site-packages/rq/worker.py", line 826, in prepare_job_execution
    pipeline.execute()
  File "<>/site-packages/redis/client.py", line 4013, in execute
    self.shard_hint)
  File "<>/site-packages/redis/connection.py", line 1187, in get_connection
    connection = self.make_connection()
  File "<>/site-packages/redis/connection.py", line 1227, in make_connection
    return self.connection_class(**self.connection_kwargs)
  File "<>/site-packages/redis/connection.py", line 828, in __init__
    super(SSLConnection, self).__init__(**kwargs)
TypeError: __init__() got an unexpected keyword argument 'ssl'
```

Unfortunately I am not able to provide a concise code fragment to reproduce the error. It looks like a regression from https://github.com/rq/rq/pull/1327
where the ssl keyword and re-use of SSLConnection were introduced:
https://github.com/rq/rq/commit/56e756f512979b0151b539c9473ceadb01afbabf

herb:
I've had the same issue. `redis==3.5.3`. afaict `redis` never supported a `ssl` argument to its `Connection` class. Was #1327 tested against some other redis client implementation/library?

selwin:
Redis does indeed accept `ssl` argument. Check the docs here: https://redis-py.readthedocs.io/en/stable/#redis.Redis

If you can get a failing test case, please post it here.

BobReid:
RQ Worker fails with this argument when used with the --with-scheduler options. I also reproduced the error by running the worker directly.

```py
redis = Redis(..., ssl=True)
worker = Worker('default', connection=redis)
worker.work(with_scheduler=True)
```

Stack Trace
```
Traceback (most recent call last):
[...]

PR #1383, opened 2020-11-25
Fix RQScheduler when run with SSL connection

BobReid:
Resolves Issue: https://github.com/rq/rq/issues/1346

Problem: 
Worker fails with "TypeError: __init__() got an unexpected keyword argument 'ssl'" when run with SSL. 

Cause:
RQScheduler was mutating the connection_kwargs dict held by the redis connection pool and was inserting incompatible keyword args. This caused an error when the pool was empty and a new connection needed to be constructed. 

Fix:
Copy the connection pool kwargs dict in RQScheduler constructor before mutating the arguments

Additional changes:
 - Adds failing test case when running the scheduler under SSL
 - Sets up stunnel tls proxy on port 9736 in the test docker image 
 - Adds a self signed SSL cert and an stunnel config file to test directory for use in the test docker image


BobReid:
Investigation of https://github.com/rq/rq/issues/1346

selwin:
Thanks!

PR #1384, opened 2020-11-27
Add `job.refresh()` to docs

cmangla:
Took me a while to figure out I needed to do `job.refresh()` in my code. Probably worth adding to the main docs.


selwin:
Thanks!

PR #1390, opened 2020-12-11
clean_worker_registry cleans in batches to prevent submitting too muc…

AxeOfMen:
…h data to redis at once when there are a large number of invalid keys

If there are a large number of invalid keys, the call by `clean_worker_registry` to `pipeline.srem()` will fail because redis will reject so much data at once. This pull request breaks the list into smaller manageable batches.

### DIFF
--- a/rq/worker_registration.py
+++ b/rq/worker_registration.py
@@ -1,8 +1,10 @@
 from .compat import as_text
 
+from rq.utils import split_list
 
 WORKERS_BY_QUEUE_KEY = 'rq:workers:%s'
 REDIS_WORKER_KEYS = 'rq:workers'
+MAX_REMOVABLE_KEYS = 1000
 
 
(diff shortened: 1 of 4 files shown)


rq/worker_registration.py
```diff
@@ -1,8 +1,10 @@
 from .compat import as_text
 
+from rq.utils import split_list
 
 WORKERS_BY_QUEUE_KEY = 'rq:workers:%s'
 REDIS_WORKER_KEYS = 'rq:workers'
+MAX_REMOVABLE_KEYS = 1000
```
selwin, line 7: Mind renaming this to `MAX_KEYS`? We may reuse this in other places.

AxeOfMen: Done

selwin:
Thanks!

PR #1394, opened 2020-12-20
Add a "stopped" JobStatus and the machinery to properly handle it

### LINKED ISSUE #1389: stop_stop_job_command() results in misleading error message and task state
After sending a stop job command, the following warning is logged:

>WARNING: Moving job to FailedJobRegistry (work-horse terminated unexpectedly; waitpid returned 9)

The first thing about this is that since the job was deliberately stopped, it's inappropriate to say that the job failed due to the work-horse terminating unexpectedly.

The second thing is that a cancelled job is considered to be "failed", and triggers the ```handle_job_failure()``` method, but there is no special [job status code](https://github.com/rq/rq/blob/016da14723d21f6801b19cbff97f2fb9bee5c5db/rq/job.py#L29-L37) that allows distinguishing between failed/crashed jobs and jobs that were deliberately stopped.  This is something that a dev (or perhaps RQ itself, see comment below) might want to know, so that they can treat these jobs differently.  It would be good to add a new job status code named "canceled" or "stopped" for this purpose.

We have a concrete use case for needing this because we record job statuses in our own database, and we do have a separate state for cancelled jobs vs. jobs that failed "normally".

```
Dec 10 19:29:29 pulp2-nightly-pulp3-source-centos7.localhost.example.com gunicorn[9720]: pulp [c80542dbc4934f75aa83420c8b23df3e]: pulpcore.tasking.util:INFO: Canceling task: 351d9bb0-fcaa-4953-a96a-6a713ad32919
Dec 10 19:29:29 pulp2-nightly-pulp3-source-centos7.localhost.example.com rq[9717]: pulp [None]: rq.worker:INFO: Killed horse pid 9789
Dec 10 19:29:29 pulp2-nightly-pulp3-source-centos7.localhost.example.com rq[9717]: pulp [None]: rq.worker:WARNING: Moving job to FailedJobRegistry (work-horse terminated unexpectedly; waitpid returned 9)
Dec 10 19:29:29 pulp2-nightly-pulp3-source-centos7.localhost.example.com rq[9717]: pulp [None]: rq.worker:INFO: 9717@pulp2-nightly-pulp3-source-centos7.localhost.example.com: 0cbb7b3b-5a38-4e2c-9dde-b579ce36fb08
Dec 10 19:29:29 pulp2-nightly-pulp3-source-centos7.localhost.example.com rq[9717]: pulp [None]: rq.worker:INFO: 9717@pulp2-nightly-pulp3-source-centos7.localhost.example.com: Job OK (0cbb7b3b-5a38-4e2c-9dde-b579ce36fb08)
```

dralley:
fixes #1389

### DIFF
--- a/rq/command.py
+++ b/rq/command.py
@@ -3,7 +3,7 @@ import os
 import signal
 
 from rq.exceptions import InvalidJobOperation
-from rq.job import Job, JobStatus
+from rq.job import Job
 
 
 PUBSUB_CHANNEL_TEMPLATE = 'rq:pubsub:%s'
@@ -72,7 +72,9 @@ def handle_stop_job_command(worker, payload):
     job_id = payload.get('job_id')
     worker.log.debug('Received command to stop job %s', job_id)
     if job_id and worker.get_current_job_id() == job_id:
-        worker.get_current_job().set_status(JobStatus.STOPPED)
+        # Sets the '_stopped_job_id' so that the job failure handler knows it
+        # was intentional.
+        worker._stopped_job_id = job_id
         worker.kill_horse()
     else:
         worker.log.info('Not working on job %s, command ignored.', job_id)
--- a/rq/command.py
+++ b/rq/command.py
@@ -3,7 +3,7 @@ import os
 import signal
 
 from rq.exceptions import InvalidJobOperation
-from rq.job import Job
+from rq.job import Job, JobStatus
 
 
 PUBSUB_CHANNEL_TEMPLATE = 'rq:pubsub:%s'
@@ -72,6 +72,7 @@ def handle_stop_job_command(worker, payload):
     job_id = payload.get('job_id')
     worker.log.debug('Received command to stop job %s', job_id)
     if job_id and worker.get_current_job_id() == job_id:
+        worker.get_current_job().set_status(JobStatus.STOPPED)
         worker.kill_horse()
     else:
         worker.log.info('Not working on job %s, command ignored.', job_id)
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -802,17 +802,25 @@ class Worker(object):
             if not job.ended_at:
                 job.ended_at = utcnow()
 
-            # Unhandled failure: move the job to the failed queue
-            self.log.warning((
-                'Moving job to FailedJobRegistry '
-                '(work-horse terminated unexpectedly; waitpid returned {})'
-            ).format(ret_val))
+            if job.is_stopped:
+                # Work-horse killed deliberately
+                self.log.warning('Job stopped by user, moving job to FailedJobRegistry')
+                self.handle_job_failure(
+                    job, queue=queue,
+                    exc_string="Job stopped by user, work-horse terminated."
+                )
+            else:
+                # Unhandled failure: move the job to the failed queue
+                self.log.warning((
+                    'Moving job to FailedJobRegistry '
+                    '(work-horse terminated unexpectedly; waitpid returned {})'
+                ).format(ret_val))
 
-            self.handle_job_failure(
-                job, queue=queue,
-                exc_string="Work-horse was terminated unexpectedly "
-                           "(waitpid returned %s)" % ret_val
-            )
+                self.handle_job_failure(
+                    job, queue=queue,
+                    exc_string="Work-horse was terminated unexpectedly "
+                            "(waitpid returned %s)" % ret_val
+                )
 
     def execute_job(self, job, queue):
         """Spawns a work horse to perform the actual work and passes it a job.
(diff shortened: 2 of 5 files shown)


dralley:
It would probably make sense to also test that stopped jobs are not retried automatically -- but I don't know how to do that.

dralley:
@selwin

selwin:
Sorry for the delayed response. Will take a look soon.

rq/command.py
```diff
@@ -72,6 +72,7 @@ def handle_stop_job_command(worker, payload):
     job_id = payload.get('job_id')
     worker.log.debug('Received command to stop job %s', job_id)
     if job_id and worker.get_current_job_id() == job_id:
+        worker.get_current_job().set_status(JobStatus.STOPPED)
```
selwin, line 75: I don’t like how we’re saving job status to Redis and is retrieved in `handle_job_failure()`. I think we can set a status to indicate that this job is being stopped in `worker` instance. We can probably set something like `worker.stopped_job_id = job_id` here and check this value during `handle_job_failure()` to know whether this job is commanded to stop.

dralley: Is the objection mostly that the job status should always be set in the ```handle_job_*()``` methods and is out of place here?

selwin: > Is the objection mostly that the job status should always be set in the handle_job_*() methods and is out of place here?

My main objection is that we're using Redis to pass status back and forth within the same process and job execution workflow. However, your observation is also valid (making status changes always set in `handle_job_*()` methods is also cleaner).

dralley: @selwin I pushed a new commit, is this what you wanted?

rq/worker.py
```diff
@@ -883,13 +900,18 @@ def handle_job_failure(self, job, queue, started_job_registry=None,
...
-                retry = True
-                retry_interval = job.get_retry_interval()
-                job.retries_left = job.retries_left - 1
-            else:
-                retry = False
+            # Don't attempt to retry if the job was deliberately stopped
+            retry = not job.is_stopped and job.retries_left and job.retries_left > 0
+            # Don't overwrite the JobStatus of stopped jobs
+            if not retry and not job.is_stopped:
                 job.set_status(JobStatus.FAILED, pipeline=pipeline)
```
selwin, line 915: I think we should group the status setting logic together so it's easier to follow. Something like:
```python
if self._stopped_job_id and self._stopped_job_id == job.id:
    job.set_status(JobStatus.STOPPED, pipeline=pipeline)
    self._stopped_job_id = None
else:
    job.set_status(JobStatus.FAILED, pipeline=pipeline)
```

rq/command.py
```diff
@@ -72,6 +72,6 @@ def handle_stop_job_command(worker, payload):
     job_id = payload.get('job_id')
     worker.log.debug('Received command to stop job %s', job_id)
     if job_id and worker.get_current_job_id() == job_id:
-        worker.kill_horse()
+        worker.stop_current_job()
```
selwin, line 75: Let's change this to:
```python
worker._stopped_job_id = job_id
worker.kill_horse()
```

dralley: You're comfortable with accessing a "private" member from outside the class, then?

rq/worker.py
dralley: Adjusted

rq/command.py
selwin: Yes, I'm ok with it. To me, marking something as private means it's not a public API and shouldn't be used by third party code.

This is first party code that's designed for our use and eliminates a network call to Redis, so it's worth it :).

dralley:
@selwin Changes applied

dralley:
^

selwin:
I'll take a look this weekend, thanks!

rq/worker.py
```diff
@@ -790,17 +792,25 @@ def monitor_work_horse(self, job, queue):
...
-                'Moving job to FailedJobRegistry '
-                '(work-horse terminated unexpectedly; waitpid returned {})'
-            ).format(ret_val))
-
-            self.handle_job_failure(
-                job, queue=queue,
-                exc_string="Work-horse was terminated unexpectedly "
-                           "(waitpid returned %s)" % ret_val
-            )
+            if job.is_stopped:
```
selwin, line 795: Instead of putting this check in this `if` block, would be cleaner and easier to understand if we do
```python
if job_status == JobStatus.STOPPED:
    ...
elif job_status not in [JobStatus.FINISHED, JobStatus.FAILED]:  # originally line 790
    ...

```

rq/worker.py
```diff
@@ -883,14 +893,19 @@ def handle_job_failure(self, job, queue, started_job_registry=None,
...
             job.worker_name = None
-            # Requeue/reschedule if retry is configured
-            if job.retries_left and job.retries_left > 0:
-                retry = True
-                retry_interval = job.get_retry_interval()
-                job.retries_left = job.retries_left - 1
+
+            # check whether a job was stopped intentionally and set the job
+            # status appropriately if it was this job.
+            job_is_stopped = self._stopped_job_id and self._stopped_job_id == job.id
```
selwin, line 899: I think we can do `job_is_stopped = self._stopped_job_id == job.id` here, since if `_stopped_job_id` is `None`, it would evaluate to `False` anyway.

rq/worker.py
```diff
@@ -883,14 +893,19 @@ def handle_job_failure(self, job, queue, started_job_registry=None,
...
+            retry = job.retries_left and job.retries_left > 0 and not job_is_stopped
+
+            if job_is_stopped:
+                job.set_status(JobStatus.STOPPED)
```
selwin, line 903: This call should be pipelined whenever possible.

dralley:
@selwin I think I addressed everything

selwin:
Looks good! One last thing, could you please document this job status in https://github.com/rq/rq/blob/master/docs/docs/jobs.md ? We should at least document this in these two sections:

CleanShot 2021-02-02 at 18 32 10@2x

And here: https://github.com/rq/rq/blob/master/docs/docs/jobs.md#stopping-a-currently-executing-job

dralley:
@selwin is any more action needed by me?

selwin:
Thanks!

PR #1396, opened 2020-12-22
rename SIGRTMIN SIGUSR1

virusandy:
didnt work otherwise on pthonnn 3.6

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -1094,7 +1094,7 @@ class HerokuWorker(Worker):
     """
     Modified version of rq worker which:
     * stops work horses getting killed with SIGTERM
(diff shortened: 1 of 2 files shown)


selwin:
I’m not sure what problem this solves. I run RQ on Python 3.6 and it seems to work just fine. Could you please elaborate more?

virusandy:
It doesn't work at least on macOS, because of no SIGRTMIN on this OS. After
changing to SIGUSR1 works well.

selwin:
Any reason why you'd want to run `HerokuWorker` on MacOS though?

selwin:
Closing this PR as I don't think it's relevant. Please reopen if you feel differently.

Issue #260, opened 2013-09-16
Suggestion: multi-job dependency


jchia:
I suggest extending the current dependency feature to allow a job to depend on multiple jobs, not just one, so that it can be run only when all those jobs it depends on have succeeded.

NewbiZ:
This issue was opened 2 years ago. Any ETA for this feature? :)

jchia:
This [pull request](https://github.com/nvie/rq/pull/279) looks dead. I really need multi-job dependency, too, and I've been using my own branch based on the pull request.

jmmills:
I'm just throwing in my 2 cents here. But doesn't adding multi-job dependencies add a lot of potential complexity and perhaps cross of into a different problem domain such as workflows?

I'm not sure what others use cases are, but I usually do simplified workflows by chaining jobs together, I.e one job creates another and returns that next jobs ID


Thanks,
   Jason Mills

- sent from mobile.

olingerc:
I think rediscussing the actual need for this is unnecessary. It has been stated multiple times over the last few years. Many people are waiting for this and have custom branches just for this feature. I sure know that I am. 
Please let us continue working on this and finally finish it.

jchia:
In my use case, my jobs don't have a linear relationship. Some jobs use the
output of multiple jobs as input and I express that as dependence on
multiple jobs. If I express everything as a sequence of jobs with
single-job dependency, just because one job fails everything downstream
cannot start.

RQ with the multiple-job is the only existing Python tool I know that can
express this sort of job scheduling requirement.

olingerc:
I'll have a go at this after the christmas holidays if nobody else is working on it?

selwin:
As far as I know, no one's currently working on it :)

jlopex:
Hey there, I took @Kisioj branch and I made a few changes on top of it then rebased against current master. Will this be a fair start? I haven't tested so much, but I'll go a little bit deeper tomorrow.

https://github.com/jlopex/rq/tree/javi/multi-dependencies

selwin:
I think this PR was almost ready when I stopped working on it so it's a good place to start from.

However, I'm also not against someone else rewriting the implementation. Who knows it could be better? ;). Some parts of the RQ codebase have also changed quite a bit too.

selwin:
TLDR; it's up to you ;)

jlopex:
@selwin it feels that it's based on a subset of your changes. I'll take a deeper look tomorrow and let you know about it. :+1:

[...]

PR #1397, opened 2020-12-28
Multidependencies

skieffer:
Hi, I'm hoping to maybe help get the ball rolling again on Issue [#260](https://github.com/rq/rq/issues/260) with this pull request. I may be completely off-base, but thought I'd give it a shot!

The PR:
* allows lists & tuples of dependencies as the value of `depends_on`, and this is passing a first unit test.
* includes the full list of dependencies (instead of just the first) in the `Job.to_dict()` representation

Note: Before I made any changes, I ran all existing unit tests. All passed _except for the three tests in `test_commands.py`_.
Maybe I'm just missing some fixture in the test setup on my machine? Dunno.

Anyway, _with those three skipped_, all unit tests are passing on my machine after my changes.

My testing was with Python 3.8.3

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -176,7 +176,7 @@ class Job(object):
 
     @property
     def _dependency_id(self):
-        """Returns the first item in self._dependency_ids. Present
+        """Returns the first item in self._dependency_ids. Present to
         preserve compatibility with third party packages..
         """
         if self._dependency_ids:
@@ -184,7 +184,7 @@ class Job(object):
 
     @property
     def dependency(self):
-        """Returns a job's dependency. To avoid repeated Redis fetches, we cache
+        """Returns a job's first dependency. To avoid repeated Redis fetches, we cache
         job.dependency as job._dependency.
         """
         if not self._dependency_ids:
@@ -787,14 +787,14 @@ class Job(object):
         return self.retry_intervals[index]
 
     def register_dependency(self, pipeline=None):
-        """Jobs may have dependencies. Jobs are enqueued only if the job they
-        depend on is successfully performed. We record this relation as
+        """Jobs may have dependencies. Jobs are enqueued only if the jobs they
+        depend on are successfully performed. We record this relation as
         a reverse dependency (a Redis set), with a key that looks something
         like:
 
             rq:job:job_id:dependents = {'job_id_1', 'job_id_2'}
 
-        This method adds the job in its dependency's dependents set
+        This method adds the job in its dependencies' dependents sets,
         and adds the job to DeferredJobRegistry.
         """
         from .registry import DeferredJobRegistry
--- a/rq/job.py
+++ b/rq/job.py
@@ -572,6 +572,7 @@ class Job(object):
         if self._status is not None:
             obj['status'] = self._status
         if self._dependency_ids:
+            obj['dependency_id'] = self._dependency_ids[0]  # for backwards compatibility
             obj['dependency_ids'] = json.dumps(self._dependency_ids)
         if self.meta and include_meta:
             obj['meta'] = self.serializer.dumps(self.meta)
--- a/rq/job.py
+++ b/rq/job.py
@@ -19,7 +19,7 @@ from .exceptions import NoSuchJobError
 from .local import LocalStack
 from .serializers import resolve_serializer
 from .utils import (enum, get_version, import_attribute, parse_timeout, str_to_date,
-                    utcformat, utcnow)
+                    utcformat, utcnow, ensure_list)
 
 # Serialize pickle dumps using the highest pickle protocol (binary, default
 # uses ascii)
@@ -126,10 +126,10 @@ class Job(object):
         job._status = status
         job.meta = meta or {}
 
-        # dependency could be job instance or id, or list or tuple thereof
+        # dependency could be job instance or id, or iterable thereof
         if depends_on is not None:
-            deps = depends_on if isinstance(depends_on, (list, tuple)) else [depends_on]
-            job._dependency_ids = [dep.id if isinstance(dep, Job) else dep for dep in deps]
+            job._dependency_ids = [dep.id if isinstance(dep, Job) else dep
+                                   for dep in ensure_list(depends_on)]
         return job
 
     def get_position(self):
--- a/rq/job.py
+++ b/rq/job.py
@@ -126,9 +126,10 @@ class Job(object):
         job._status = status
         job.meta = meta or {}
 
-        # dependency could be job instance or id
+        # dependency could be job instance or id, or list or tuple thereof
         if depends_on is not None:
-            job._dependency_ids = [depends_on.id if isinstance(depends_on, Job) else depends_on]
+            deps = depends_on if isinstance(depends_on, (list, tuple)) else [depends_on]
+            job._dependency_ids = [dep.id if isinstance(dep, Job) else dep for dep in deps]
         return job
 
     def get_position(self):
--- a/rq/job.py
+++ b/rq/job.py
@@ -499,8 +499,7 @@ class Job(object):
         self.failure_ttl = int(obj.get('failure_ttl')) if obj.get('failure_ttl') else None  # noqa
         self._status = obj.get('status').decode() if obj.get('status') else None
 
-        dependency_ids = json.loads(as_text(obj.get('dependency_ids', '[]')))
-        self._dependency_ids = [as_text(ID) for ID in dependency_ids]
+        self._dependency_ids = json.loads(as_text(obj.get('dependency_ids', '[]')))
 
         self.ttl = int(obj.get('ttl')) if obj.get('ttl') else None
         self.meta = self.serializer.loads(obj.get('meta')) if obj.get('meta') else {}
--- a/rq/job.py
+++ b/rq/job.py
@@ -499,8 +499,8 @@ class Job(object):
         self.failure_ttl = int(obj.get('failure_ttl')) if obj.get('failure_ttl') else None  # noqa
         self._status = obj.get('status').decode() if obj.get('status') else None
 
-        dependency_id = obj.get('dependency_id', None)
-        self._dependency_ids = [as_text(dependency_id)] if dependency_id else []
+        dependency_ids = json.loads(obj.get('dependency_ids', '[]'))
+        self._dependency_ids = [as_text(ID) for ID in dependency_ids]
 
         self.ttl = int(obj.get('ttl')) if obj.get('ttl') else None
         self.meta = self.serializer.loads(obj.get('meta')) if obj.get('meta') else {}
@@ -572,7 +572,7 @@ class Job(object):
         if self._status is not None:
             obj['status'] = self._status
         if self._dependency_ids:
-            obj['dependency_id'] = self._dependency_ids[0]
+            obj['dependency_ids'] = json.dumps(self._dependency_ids)
         if self.meta and include_meta:
             obj['meta'] = self.serializer.dumps(self.meta)
         if self.ttl:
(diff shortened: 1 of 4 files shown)


skieffer:
Hmm, actually, not sure worker `w2` is doing anything in my new unit tests...

Is there an easy way to get two workers working within a unit test?

Tried `os.fork()` but that's not doing it......

rq/job.py
```diff
@@ -498,8 +499,7 @@ def restore(self, raw_data):
         self.failure_ttl = int(obj.get('failure_ttl')) if obj.get('failure_ttl') else None  # noqa
         self._status = obj.get('status').decode() if obj.get('status') else None
 
-        dependency_id = obj.get('dependency_id', None)
-        self._dependency_ids = [as_text(dependency_id)] if dependency_id else []
+        self._dependency_ids = json.loads(as_text(obj.get('dependency_ids', '[]')))
```
selwin, line 502: For backwards compatibility purposes, this should also check for `dependency_id`.

tests/test_job.py
```diff
@@ -432,6 +432,35 @@ def test_description_is_persisted(self):
         Job.fetch(job.id, connection=self.testconn)
         self.assertEqual(job.description, "tests.fixtures.say_hello('Lionel')")
 
+    def test_multiple_dependencies_are_accepted_and_persisted(self):
+        """Ensure job._dependency_ids accepts different input formats, and
+        is set and restored properly"""
+        jobA = Job.create(func=fixtures.some_calculation, args=(3, 1, 4), id="A")
```
selwin, line 438: Let's use snake case instead of camel case to keep things consistent.

tests/test_job.py
```diff
@@ -943,7 +972,82 @@ def test_dependents_are_met_if_dependency_is_deleted(self):
...
+        queue.enqueue_call(fixtures.rpush, args=[key, "A"])
+        queue.enqueue_call(fixtures.rpush, args=[key, "B"])
+
+        self.testconn.delete(key)
+        w1.work(burst=True)
+        w2.work(burst=True)
+        assert queue.count == 0
+
+        order_completed = [as_text(v) for v in self.testconn.lrange(key, 0, 1)]
+        print('\n', order_completed)
```
selwin, line 997: Mind removing the print statements?

tests/test_job.py
```diff
@@ -943,7 +972,82 @@ def test_dependents_are_met_if_dependency_is_deleted(self):
 
         assert dependent_job.dependencies_are_met()
         assert dependent_job.get_status() == JobStatus.QUEUED
-    
+
+    def test_rpush_fixture(self):
+        fixtures.rpush('foo', 'bar')
+        assert as_text(self.testconn.lrange('foo', 0, 0)[0]) == 'bar'
```
selwin, line 978: Can we do `self.testconn.lrange('foo', 0, 0).decode()[0] == 'bar'` here? I want to eventually get rid of `as_text` because this was used for Python 2/3 compatibility purposes.

tests/test_job.py
```diff
@@ -943,7 +972,82 @@ def test_dependents_are_met_if_dependency_is_deleted(self):
...
+        w2 = Worker([queue])
+
+        # When there are no dependencies, the two fast jobs ("A" and "B") run in the order enqueued.
+        queue.enqueue_call(fixtures.long_running_job, kwargs={'timeout': 0.5}, job_id='slow_guy')
+        queue.enqueue_call(fixtures.rpush, args=[key, "A"])
+        queue.enqueue_call(fixtures.rpush, args=[key, "B"])
+
+        self.testconn.delete(key)
+        w1.work(burst=True)
+        w2.work(burst=True)
```
selwin, line 993: Why do we need to use two workers here? It seems like `w1` would just be running all three jobs serially.

skieffer: I think you're right, and I wanted to ask advice here.

The fact that these unit tests actually pass, given that in reality only one worker is
doing all the jobs, is actually quite interesting. I guess it demonstrates that jobs with
dependencies get moved to the deferred job registry at the time they're enqueued,
and that this affects execution order.

But I don't really feel it proves what these unit tests _want_ to prove, i.e., that in
a situation with multiple workers, the free workers really will skip over dependent jobs
until those jobs are ready to go.

If we can get two workers going, we can test that. Is there an easy way to do that,
within a unit test? I tried messing around with `os.fork()`, but it's failing, and I
don't understand why. (Maybe because it's all happening within `pytest`? I can post some
code if you want.)

selwin: I'd rather not mess around with `os.fork()` during testing - it feels somewhat hackish.

> But I don't really feel it proves what these unit tests want to prove, i.e., that in
a situation with multiple workers, the free workers really will skip over dependent jobs
until those jobs are ready to go.
We can do this by comprehensively testing `job.dependencies_are_met()` - this method should only return `True` if all dependencies have been satisfied.

skieffer: Agreed, `os.fork()` is probably best avoided. : ) However I found some existing unit tests in `test_worker.py` and `test_commands.py` where you use `multiprocessing` to achieve something similar, which gave me some ideas. So I've tried again, and I think these tests are in good shape now.

I'm not sure if I understood what you meant about using `job.dependencies_are_met()`, but I've made an attempt in unit test `test_dependencies_are_met_at_execution_time`.

tests/test_job.py
skieffer: Done.

tests/test_job.py
skieffer: Done.

tests/test_job.py
skieffer: Done.

tests/test_job.py
```diff
@@ -943,7 +972,95 @@ def test_dependents_are_met_if_dependency_is_deleted(self):
...
+        jobs_completed = [v.decode() for v in self.testconn.lrange(key, 0, 1)]
+        # Job A completed first, and was done by worker 2.
+        assert jobs_completed[0] == "A:w2"
+
+        # When job "A" depends on the slow job, then job "B" finishes before "A".
+        queue.enqueue_call(fixtures.long_running_job, kwargs={'timeout': 0.5}, job_id='slow_job')
+        queue.enqueue_call(fixtures.rpush, args=[key, "A", True], depends_on='slow_job')
+        queue.enqueue_call(fixtures.rpush, args=[key, "B", True])
+
+        self.testconn.delete(key)
```
selwin, line 1022: Nitpick: mind moving all `self.testconn.delete(key)` lines to after assertions are completed? (line 1015) This provides a better separation between separate test cases.

tests/test_job.py
```diff
@@ -943,7 +972,95 @@ def test_dependents_are_met_if_dependency_is_deleted(self):
...
+        fixtures.burst_two_workers(queue)
+        assert queue.count == 0
+
+        jobs_completed = [v.decode() for v in self.testconn.lrange(key, 0, 1)]
+        # Job B completed first, again by worker 2.
+        assert jobs_completed[0] == "B:w2"
```
selwin, line 1028: Should we wait for 0.75 seconds and assert that all three jobs finish?

tests/test_job.py
```diff
@@ -943,7 +972,95 @@ def test_dependents_are_met_if_dependency_is_deleted(self):
...
+        queue.enqueue_call(fixtures.rpush, args=[key, "A", True])
+        queue.enqueue_call(fixtures.rpush, args=[key, "B", True])
+
+        self.testconn.delete(key)
+        fixtures.burst_two_workers(queue)
+        assert queue.count == 0
+
+        jobs_completed = [v.decode() for v in self.testconn.lrange(key, 0, 1)]
+        # Job A completed first, and was done by worker 2.
+        assert jobs_completed[0] == "A:w2"
```
selwin, line 1015: Can we rewrite this as `self.assertEqual(jobs_completed[0], "A:w2)`? Want to keep this consistent across our codebase.

tests/test_job.py
```diff
@@ -943,7 +972,95 @@ def test_dependents_are_met_if_dependency_is_deleted(self):
...
+        queue.enqueue_call(fixtures.rpush, args=[key, "A", True], depends_on=['one_slow_job', 'another_slow_job'])
+        queue.enqueue_call(fixtures.rpush, args=[key, "B", True], depends_on='one_slow_job')
+
+        self.testconn.delete(key)
+        fixtures.burst_two_workers(queue)
+        assert queue.count == 0
+
+        jobs_completed = [v.decode() for v in self.testconn.lrange(key, 0, 1)]
+        # Job B completed first, again by worker 1.
+        assert jobs_completed[0] == "B:w1"
```
selwin, line 1062: Can we assert that all four jobs complete in the expected order?

skieffer: Done, plus the reformatting for the above comments.

tests/test_job.py
skieffer: Done.

tests/test_job.py
```diff
@@ -943,7 +972,90 @@ def test_dependents_are_met_if_dependency_is_deleted(self):
...
         assert dependent_job.dependencies_are_met()
         assert dependent_job.get_status() == JobStatus.QUEUED
-    
+
+    def test_dependencies_are_met_at_execution_time(self):
+        queue = Queue(connection=self.testconn)
+
+        queue.enqueue_call(fixtures.say_hello, job_id="A")
+        queue.enqueue_call(fixtures.say_hello, job_id="B")
+        job_C = queue.enqueue_call(fixtures.check_dependencies_are_met, job_id="C", depends_on=["A", "B"])
```
selwin, line 981: Nitpick: can we use the regular `queue.enqueue()` API here? Tests will be cleaner this way, I think `enqueue_call` should be used only when we're specifically testing it.

tests/test_job.py
```diff
@@ -943,7 +972,90 @@ def test_dependents_are_met_if_dependency_is_deleted(self):
...
+    def test_rpush_fixture(self):
+        fixtures.rpush('foo', 'bar')
+        assert self.testconn.lrange('foo', 0, 0)[0].decode() == 'bar'
+
+    def test_start_worker_fixture(self):
+        queue = Queue(name='testing', connection=self.testconn)
+        queue.enqueue(fixtures.say_hello)
+        conn_kwargs = self.testconn.connection_pool.connection_kwargs
+        fixtures.start_worker(queue.name, conn_kwargs, 'w1', True)
+        assert not queue.jobs
```
selwin, line 997: Mind splitting these two tests to a separate `FixtureTestCase` class?

tests/test_job.py
```diff
@@ -943,7 +972,90 @@ def test_dependents_are_met_if_dependency_is_deleted(self):
...
+
+    def test_execution_order_with_sole_dependency(self):
+        queue = Queue(connection=self.testconn)
+        key = 'test_job:job_order'
+
+        # When there are no dependencies, the two fast jobs ("A" and "B") run in the order enqueued.
+        # Worker 1 will be busy with the slow job, so worker 2 will complete both fast jobs.
+        job_slow = queue.enqueue_call(fixtures.rpush, args=[key, "slow", True, 0.5], job_id='slow_job')
+        job_A = queue.enqueue_call(fixtures.rpush, args=[key, "A", True])
+        job_B = queue.enqueue_call(fixtures.rpush, args=[key, "B", True])
```
selwin, line 1007: Please use `queue.enqueue()` if you can.

tests/test_job.py
skieffer: I looked around, and most existing subclasses of `RQTestCase` seem to live in their own module called `test_something.py`, so I followed suit with a new `test_fixtures.py`. Not sure if that's what you had in mind.

Also followed suit with the `from __future__ import.....` line at the top of this module, which most of the `test_somethings.py` seem to include, though I'll admit I don't really know if that's needed or not. Just trying to keep it uniform.

selwin:
Thanks!

skieffer:
My pleasure. Happy to contribute one or two lines of code to a great project. Thanks for taking the time to review, and in general for providing this extremely useful library!

thomasmatecki:
It's awesome to see this coming along.

@wketcha @martin-benson does this work address any of the issues noted in #1170 ?

Issue #1398, opened 2021-01-03
`TypeError: __init__() got an unexpected keyword argument 'path'`


homeworkprod:
Versions:
```
hiredis==1.1.0
redis==3.5.3
rq==1.7.0
```

Launch script is basically:
```py
with connection():
    queues = [get_queue(app)]

    worker = Worker(queues)
    worker.work(with_scheduler=True)
```

Resulting in this exception right on start of the worker:
```pytb
TypeError: __init__() got an unexpected keyword argument 'path'
  File "worker.py", line 32, in <module>
    worker.work(with_scheduler=True)
  File "rq/worker.py", line 546, in work
    self.scheduler.acquire_locks()
  File "rq/scheduler.py", line 95, in acquire_locks
    if self.connection.set(self.get_locking_key(name), pid, nx=True, ex=60):
  File "rq/scheduler.py", line 69, in connection
    self._connection = self._connection_class(**self._connection_kwargs)
```

Observations:
- This works locally with both rq v1.6.1 and v1.7.0.
- This fails on prod with both rq v1.6.1 and v1.7.0.
- It works without the `with_scheduler=True` keyword argument.
- Locally, the Redis URL is a network socket (`redis://127.0.0.1:6379/0`), but on prod it's a file socket (`unix:///var/run/redis/redis.sock?db=0`).

That's all I have for now.

homeworkprod:
Further investigation results:
- Python version is 3.7.3 both locally and on prod.
- rq passes a connection object around which is an instance of `redis.client.Redis` from the `redis` library.
- Class `redis.client.Redis` has a field `unix_socket_path` which is [put into its connection keyword arguments with the key `path`])https://github.com/andymccurdy/redis-py/blob/master/redis/client.py#L776).
- rq's worker [creates a scheduler instance and passes the connection to it](https://github.com/rq/rq/blob/v1.7.0/rq/worker.py#L544).
- When the scheduler [creates its own connection object for the first time](https://github.com/rq/rq/blob/v1.7.0/rq/scheduler.py#L69), the exception is raised as `redis.client.Redis`'s constructor accepts only `unix_socket_path` as a keyword argument but not `path`.

Sooo … yeah. I'm not convinved that changing keyword names is the cleanest strategy here.

homeworkprod:
And it probably makes sense to add a test case that combines a Redis socket path and (a worker with) a scheduler, as it looks like this hasn't been tested before releasing the scheduler mechanism (not blaming anyone!).

[...]

PR #1403, opened 2021-01-21
Fix scheduler when UNIX socket path is configured

homeworkprod:
As I've reported in #1398, an exception was raised when I tried to run a worker with enabled scheduler in combination with a Redis server that the client connects to via UNIX domain socket.

I believe this change fixes the issue, and it works in my setup.

The approach I used is basically the same as the one already used in case an `SSLConnection` is in the pool.

### DIFF
--- a/rq/scheduler.py
+++ b/rq/scheduler.py
@@ -6,7 +6,7 @@ import traceback
 from datetime import datetime
 from multiprocessing import Process
 
-from redis import Redis, SSLConnection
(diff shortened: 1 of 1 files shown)


homeworkprod:
I don't see an easy way to add appropriate tests here.

This would probably involve a second Docker image with an adjusted Redis server configuration to use a UNIX domain socket instead of the network, and then some.

selwin:
Thanks!

PR #1405, opened 2021-01-27
Add runner for asyncio tasks

MyrikLD:
Example:
```
async def run(arg1):
    [some code]

q = Queue(connection=Redis())
job = q.enqueue(run, arg1=arg1)
```

### DIFF
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -722,11 +722,17 @@ class TestJob(RQTestCase):
     def test_create_job_with_async(self):
         """test creating jobs with async function"""
         queue = Queue(connection=self.testconn)
-        job = queue.enqueue(fixtures.say_hello_async, job_id="1234")
-        self.assertEqual(job.id, "1234")
-        job.perform()
 
-        self.assertRaises(TypeError, queue.enqueue, fixtures.say_hello, job_id=1234)
+        async_job = queue.enqueue(fixtures.say_hello_async, job_id="async_job")
+        sync_job = queue.enqueue(fixtures.say_hello, job_id="sync_job")
+
+        self.assertEqual(async_job.id, "async_job")
+        self.assertEqual(sync_job.id, "sync_job")
+
+        async_task_result = async_job.perform()
+        sync_task_result = sync_job.perform()
+
+        self.assertEqual(sync_task_result, async_task_result)
 
     def test_get_call_string_unicode(self):
         """test call string with unicode keyword arguments"""
--- a/tests/test_job.py
+++ b/tests/test_job.py
@@ -719,6 +719,15 @@ class TestJob(RQTestCase):
 
         self.assertRaises(TypeError, queue.enqueue, fixtures.say_hello, job_id=1234)
 
+    def test_create_job_with_async(self):
+        """test creating jobs with async function"""
+        queue = Queue(connection=self.testconn)
+        job = queue.enqueue(fixtures.say_hello_async, job_id="1234")
+        self.assertEqual(job.id, "1234")
+        job.perform()
+
+        self.assertRaises(TypeError, queue.enqueue, fixtures.say_hello, job_id=1234)
+
     def test_get_call_string_unicode(self):
         """test call string with unicode keyword arguments"""
         queue = Queue(connection=self.testconn)
--- a/rq/job.py
+++ b/rq/job.py
@@ -719,7 +719,8 @@ class Job(object):
     def _execute(self):
         result = self.func(*self.args, **self.kwargs)
         if asyncio.iscoroutine(result):
-            return asyncio.run(result)
+            loop = asyncio.get_event_loop()
+            return loop.run_until_complete(asyncio.wait([result]))
         return result
 
     def get_ttl(self, default_ttl=None):
--- a/rq/job.py
+++ b/rq/job.py
@@ -8,6 +8,7 @@ import pickle
 import warnings
 import zlib
 
+import asyncio
 from collections.abc import Iterable
 from distutils.version import StrictVersion
 from functools import partial
@@ -716,7 +717,10 @@ class Job(object):
             pipeline.hmset(self.key, mapping)
 
     def _execute(self):
-        return self.func(*self.args, **self.kwargs)
+        result = self.func(*self.args, **self.kwargs)
+        if asyncio.iscoroutine(result):
+            return asyncio.run(result)
+        return result
 
     def get_ttl(self, default_ttl=None):
         """Returns ttl for a job that determines how long a job will be
--- a/rq/job.py
+++ b/rq/job.py
@@ -720,7 +720,8 @@ class Job(object):
         result = self.func(*self.args, **self.kwargs)
         if asyncio.iscoroutine(result):
             loop = asyncio.get_event_loop()
-            return loop.run_until_complete(asyncio.wait([result]))
+            coro_result = loop.run_until_complete(result)
+            return coro_result
         return result
 
     def get_ttl(self, default_ttl=None):
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -29,6 +29,13 @@ def say_hello(name=None):
     return 'Hi there, %s!' % (name,)
 
 
+async def say_hello_async(name=None):
+    """A async job with a single argument and a return value."""
+    if name is None:
+        name = 'Stranger'
+    return 'Hi there, %s!' % (name,)
+
+
 def say_hello_unicode(name=None):
     """A job with a single argument and a return value."""
     return text_type(say_hello(name))  # noqa
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -31,9 +31,7 @@ def say_hello(name=None):
 
 async def say_hello_async(name=None):
     """A async job with a single argument and a return value."""
-    if name is None:
-        name = 'Stranger'
-    return 'Hi there, %s!' % (name,)
+    return say_hello(name)
 
 
 def say_hello_unicode(name=None):


MyrikLD:
There fixed https://github.com/rq/rq/pull/1405 request

tests/test_job.py
```diff
@@ -719,6 +719,15 @@ def test_create_job_with_id(self):
 
         self.assertRaises(TypeError, queue.enqueue, fixtures.say_hello, job_id=1234)
 
+    def test_create_job_with_async(self):
+        """test creating jobs with async function"""
+        queue = Queue(connection=self.testconn)
+        job = queue.enqueue(fixtures.say_hello_async, job_id="1234")
+        self.assertEqual(job.id, "1234")
+        job.perform()
```
selwin, line 727: Thanks for this PR! Could you please add an assertion here to check that the result is right?

tests/test_job.py
```diff
@@ -719,6 +719,15 @@ def test_create_job_with_id(self):
...
+
+        self.assertRaises(TypeError, queue.enqueue, fixtures.say_hello, job_id=1234)
```
selwin, line 729: Mind writing a comment as to why this line raises `TypeError`?

tests/test_job.py
bpevzner: Ready both comments

selwin:
Thanks!

lowercase00:
@MyrikLD thanks for sharing this. Do you have any experience mixing this with a `gevent` worker in production?
We had been doing this manually for a while (didn't realize it was already merged) with if/else statements, but we use a custom gevent worker, that `spawns` the `perform_job` method.

Looking at [this](https://stackoverflow.com/questions/69888850/how-to-call-async-method-from-greenlet-playwright), it seems that we are indeed creating a new event loop for every coroutine, seems excessive. Any experiences with that?

lowercase00:
Now, looking at the source again, I it indeed creates a new coroutine for each new job.
```
    def _execute(self):
        result = self.func(*self.args, **self.kwargs)
        if asyncio.iscoroutine(result):
            loop = asyncio.new_event_loop()
            coro_result = loop.run_until_complete(result)
            return coro_result
        return result
```

perhaps this could be written like:

```
    def _execute(self):
        result = self.func(*self.args, **self.kwargs)
        if asyncio.iscoroutine(result):
            try:
                loop = asyncio.get_event_loop()
            except RuntimeError as error:
                logger.info("No event loop found, creating one")
                loop = asyncio.new_event_loop()
            coro_result = loop.run_until_complete(result)
            return coro_result
        return result
```

I think this should solve this (to be fair I'm not even sure how much of a problem this is).
Ideas?

MyrikLD:
@lowercase00 No I didnt used gevent in prod, asyncio almost always enough.
Function `get_event_loop` just returning current event loop. I had no problem with it, but we using only latest python, without gevent. Maybe, you can create issue with description of your current `RuntimeError`?

lowercase00:
There is no Runtime error, that was just an except clause in case there was no running loop (this method raises an error if there's no loop).

It seems that as it is now, a new coroutine is being created every time (create new loop is called no matter what - for an async function). Which I guess is not ideal. The idea is: this function is async, let's try to get a running loop. If there is no running loop, than create a new one.

MyrikLD:
Aha! No, I used get_event_loop, not new_event_loop in this pr. @rpkak changed this logic in this pr https://github.com/rq/rq/pull/1563

MyrikLD:
Official documentation advises to use `asyncio.run`, but it available only from python 3.7
https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop

lowercase00:
Yeah! I was looking at the source, didn't realize it was different than the current source.

I think that unless there was a 'loop.run_forever' call, the 'get_event_loop' raises a RuntimeError if there is no loop. And the current logic always creates a new loop.

It seems it would be best to mix those two: try to get the current loop, if no loop exists, create a new one. I could give it a try and share the fork with you guys to you can take a look.

MyrikLD:
`get_event_loop` is deprecated from python 3.10. I think, this code will work better.

```
    def _execute(self):
        result = self.func(*self.args, **self.kwargs)
        if asyncio.iscoroutine(result):
            if sys.version_info >= (3,7):
                coro_result = asyncio.run(result)
            else:
                loop = asyncio.get_running_loop()
                coro_result = loop.run_until_complete(result)
            return coro_result
        return result
 ```

**Comment by lowercase00** [comment]
Yes, should be `running` indeed. But ideally the `run` method should be used as the single entrypoint, so not sure what would be the tradeoffs between these two approaches.

> This function always creates a new event loop and closes it at the end. It should be used as a main entry point for asyncio programs, and should ideally only be called once.

I'm not sure if it would work to use it at `worker.work()` though (I might be needed to wrap it into an async function to be a coroutine, but I'm not sure).

I think we are possibly looking at something like:

```python
def _execute(self):
    is_python37 = sys.version_info >= (3, 7)
    result = self.func(*self.args, **self.kwargs)
    if asyncio.iscoroutine(result):
        try:
            loop = asyncio.get_running_loop() if is_python37 else asyncio.get_event_loop()
        except RuntimeError as error:
            logger.info("No event loop found, creating one")
            loop = asyncio.new_event_loop()
        coro_result = loop.run_until_complete(result)
        return coro_result
    return result
```

- Coroutine only being created if needed
- If there is already a loop running, uses that
- If there is no loop, creates a new one.
- If the system is lower than 3.7 uses the deprecated (`get_event_loop`), else uses the newer (`get_running_loop`).

Haven't tested it though.
The last missing part would be to make this threadsafe, using `run_coroutine_threadsafe`.

rpkak:
I don't know how @selwin handles python version support, but according to [pep 494](https://peps.python.org/pep-0494/) Python 3.6 got deprecated at christmas 2021. If the support of the deprecated python versions is not required we could always just use `asyncio.run`.

selwin:
So I mentioned in https://github.com/rq/rq/blob/master/CHANGES.md#rq-111-2022-07-31 that we'll stop supporting Python 3.5 soon.

As for 3.6, I think we'll have to keep supporting it for a bit longer since a lot of enterprises are still running on older versions of Python.

lowercase00:
I was having the same problem as this issue (https://github.com/rq/rq/issues/1697). I ended up just running a worker and it ended up not really being a problem, but I got curious, and I think there's a way to make it work regardless of how the task is called.

The issue is: when you enqueue a task from a async test, there's already a loop running. So you can't just use the `asyncio.run` and if you `get_running_loop` you can't `run_until_complete` since it will never complete (the `parent` function is still running). This will work with the worker, but won't work with a sync task, or with a SimpleWorker with burst for example.

You can,  however, implement a new loop in a `ThreadExecutor`, and then wait for this second loop to finish, like so:

```python
    def _execute(self):
        # Get's the target
        result = self.func(*self.args, **self.kwargs)
        # If it's not async, just run normally.
        if not asyncio.iscoroutine(result):
            return result

        # The function is a coroutine.
        try:
            # Tries to get the current running loop (if called from an async context)
            # If there is no running loop, it was called from a sync context, so we create a new one (goto RuntimeError block)
            loop = asyncio.get_running_loop()
            import concurrent.futures
            with concurrent.futures.ThreadPoolExecutor() as executor:
                # Submitis the a method that creates a new loop, runs it, waits until finishes and closes.
                future = executor.submit(self._run_task_in_async_context, result)
                # Return the results normally.
                return_value = future.result()
                return return_value

        except RuntimeError:
            # Sync context, there is no active Loop.
            # Creates the new loop, and runs until complete
            loop = asyncio.new_event_loop()
            # Safe to run until complete, since it is a sync context, there's only one task pending.
            coro_result = loop.run_until_complete(result)
            print(coro_result)
            return coro_result

    def _run_task_in_async_context(self, task: t.Coroutine) -> t.Any:
        print("Inside the thread executor")
        import threading
        print(threading.current_thread())
        print("Creating a new loop")
        loop = asyncio.new_event_loop()
        result = loop.run_until_complete(task)
        print("Result is...", result)
        loop.close()
        print("Closing the second loop")
        return result
```

With this implementation, if there's already a loop running, we won't interfere with it, so we submit a new function to the executor that basically spawns a new loop, runs it until it finishes (there will be only the target function in it), and then closes the loop.

If the task is being called from a `sync` context, there won't be a running loop, to the first line inside the `try` block will raise a `RunimeError` and the code will continue normally as it is now basically, creating a new loop, and running it until complete. So you can run a `async` function being called from a `sync` function. You can run a `async` function being called from a `sync` function, and you can run a `sync` function from any context. 

I think this should make RQ fully (maybe?) compatible with `asyncio`, although dependent on the threadpool executor. I couldn't make it work with the `ProcessPoolExecutor`, don't really know why.

TBH this seems like a far from ideal solution, but it seems to work with tests I've done locally.
Setup used:

```python
# app/tasks.py
import asyncio

async def async_job_sample(some_param: str):
    print('Async RQ test: The param is: ' + some_param)
    await asyncio.sleep(0.2)
    print('Async RQ test: Done')
    return some_param
```

```python
# app/app.py
import asyncio
from redis import Redis
from rq import Queue
from rq.worker import SimpleWorker
from app.tasks import async_job_sample

# Test 1, normal
def testing_rq():
    print("Testing...")
    queue = Queue(connection=Redis())
    job = queue.enqueue(async_job_sample, 'Async test string')
    print("Testing... Finished")
    return job

job = testing_rq()

## Test 2: Normal with local worker
def testing_rq2():
    print("Testing...")
    queue = Queue(connection=Redis())
    job = queue.enqueue(async_job_sample, 'Async test string')
    print("Testing... Finished")
    worker = SimpleWorker(queues=[queue], connection=queue.connection)
    worker.work(burst=True)
    return job

job = testing_rq2()


## Test 3: Normal, not async
def testing_rq3():
    print("Testing...")
    queue = Queue(connection=Redis(), is_async=False)
    job = queue.enqueue(async_job_sample, 'Async test string')
    print("Testing... Finished")
    return job

job = testing_rq3()


## Test 4 From async
async def testing_rq3():
    print("Testing...")
    queue = Queue(connection=Redis())
    job = queue.enqueue(async_job_sample, 'Async test string')
    print("Testing... Finished")
    return job

job = asyncio.run(testing_rq3())

etc.

```

PR #1411, opened 2021-02-09
bugfix: Add self.serializer when calling self.fetch_many in fetch_dependencies

jtfidje:
Code fails when I use JSONSerializer and have dependencies. This fix solves it.


selwin:
Thanks!

Issue #1402, opened 2021-01-20
Changing the queues that a worker listens to at runtime


bielcardona:
In order to make certain computations (i.e. I am more focused on "few and long" tasks than "many and short" ones) I plan to create on the fly certain queues of computations and make workers listen to these queues. 

Since the computations have different stages, I consider different kinds of queues, and multiple instances of those. Say I have queues: QA1, QA2, ..., QB1, QB2, ..., QC1, QC2, ... (note that the number of such queues changes at runtime)

I would like to create workers such that each of them only listens to queues of a certain kind. For instance, I would like to create workers that only listen to queues of kind A (QA1, QA2,...). 

Since when I create the workers I do not know which concrete queues they will listen to, I cannot use the `queues` parameter of the CLI.

My first idea was to modify the `queues` attribute of the workers (after they are created). For instance, I create a worker `w` so that it listens to a dummy queue and then, at runtime, I add the created queues to the list `w.queues`, but messages sent to those added queues are not processed by `w`.

My second idea was to subclass Queue, so that each of these kinds of queues belongs to a different subclass, and then pass the parameter `--queue-class` when creating the worker (without passing the `queues` parameter). Unfortunately, this does not work, and the worker so created listens to all queues even if their class is not the specified (is this the expected behaviour?).

Do you know any way I can do what I want to do? Thanks in advance!

selwin:
Can't you just split the jobs into different queues that are executed by different worker classes? (for example CPU heavy jobs and GPU heavy jobs).

If what you want is to change which queues workers listen to at runtime, we actually already have the infrastructure for that. The latest version of RQ lets you [send commands](https://github.com/rq/rq/blob/master/rq/command.py#L12) to workers that are running.

If you want to implement a command that instructs a worker to change which queues it listens to, I'm happy to accept a PR.

PR #1412, opened 2021-02-09
implemented round-robin and random access to queues

bielcardona:
I have subclassed Queue in order to override the dequeue_any methods and implement both a round-robin strategy (if a job is dequeued from a queue, then this queue will be the last to be polled in the next iteration) and a random strategy (the order of the queues is shuffled at each iteration).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,3 @@ Vagrantfile
 Gemfile
 Gemfile.lock
 _site/
-
--- a/.gitignore
+++ b/.gitignore
@@ -20,4 +20,4 @@ Vagrantfile
 Gemfile
 Gemfile.lock
 _site/
-venv
+
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ Vagrantfile
 Gemfile
 Gemfile.lock
 _site/
+venv
--- a/docs/docs/workers.md
+++ b/docs/docs/workers.md
@@ -344,6 +344,111 @@ queue.enqueue(some_func)
 ```
 
 
+### Queues with Round-Robin and Random access
+
+The subclasses of `Queue`, `RoundRobinQueue` and `RandomQueue` 
+implement round-robin and random access to a list of queues. 
+
+Consider the following script:
+
+```python
+from rq.queue import Queue
+from redis import Redis
+from tests.fixtures import say_hello
+
+redis = Redis()
+
+qs = [Queue(f'q{i}', connection=redis) for i in range(5)]
+
+for i in range(5):
+    for j in range(3):
+        qs[i].enqueue(say_hello, f'queue={i}, job={j}')
+
+for k in range(15):
+    job, queue = Queue.dequeue_any(qs, None, connection=redis)
+    print(job.args[0])
+```
+Since the queues are polled in order, the ordering of the dequeued jobs is:
+```
+queue=0, job=0
+queue=0, job=1
+queue=0, job=2
+queue=1, job=0
+queue=1, job=1
+queue=1, job=2
+queue=2, job=0
+queue=2, job=1
+queue=2, job=2
+queue=3, job=0
+queue=3, job=1
+queue=3, job=2
+queue=4, job=0
+queue=4, job=1
+queue=4, job=2
+```
+
+Sometimes you may want to use a Round-Robin strategy, so that all 
+queues are treated equally (at the first iteration the most preferent queue is the first one, but in the second 
+iteration it is the second queue, and so on). Consider the modified script:
+
+```python
+from rq.queue import RoundRobinQueue
+from redis import Redis
+from tests.fixtures import say_hello
+
+redis = Redis()
+
+qs = [RoundRobinQueue(f'q{i}', connection=redis) for i in range(5)]
+
+for i in range(5):
+    for j in range(3):
+        qs[i].enqueue(say_hello, f'queue={i}, job={j}')
+
+for k in range(15):
+    job, queue = RoundRobinQueue.dequeue_any(qs, None, connection=redis)
+    print(job.args[0])
+```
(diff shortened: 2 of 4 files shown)


ffeldhaus:
Thanks a lot for creating the PR! Can you please add test cases in `tests/test_queue.py` in the functions `test_dequeue_any` and `test_dequeue_any_ignores_nonexisting_jobs`? It probably is sufficient to iterate over all existing queue classes there, but maybe all functions in `tests/test_queue.py` should test against the newly added queue classes to detect side effects.

ffeldhaus:
Please also remove the change to .gitignore

bielcardona:
I have added two new tests test_dequeue_any_round_robin and test_dequeue_any_round_robin_ignores_nonexisting_jobs. 
The first one is significant (in my opinion) since it shows that the jobs are dequeued following a round-robin strategy. The second one could be integrated in the original one simply by iterating over the subclasses, as you suggested. 
Please let me know if you want me to make any change.

ffeldhaus:
The tests look good to me. Can you also add tests for RandomQueue?
Please also consider adding an example to the documentation how to use the queues. I think `docs/docs/workers.md` would be the right place.

bielcardona:
Just added tests and docs

selwin:
Sorry for the delayed response. This is fantastic!

I like the idea of being able to dequeue jobs in a round robin fashion. Ideally, I'd like if we can do this using RQ's default worker. Would you be open to modifying the implementation a bit?

I think a good way to do this would be to build this feature into `Worker` instead of `Queue`. It would look somethin llike `worker = Worker(dequeue_strategy=ROUND_ROBIN)`.

After this is done, we can also do this via the command line: `rq worker queue1 queue2 queue3 --dequeue-strategy=round-robin`

bielcardona:
I agree that having it implemented in worker would seem more reasonable, and I am more than willing to adapt the code. However, I see some caveats. 

Imagine you have a given number of workers w_1,...,w_n waiting for jobs in queues q_1,...,q_m, and the dequeue strategy is set to be round-robin. Imagine also that all queues have pending jobs. If the round-robin strategy is implemented at the worker level (and hence each worker knows nothing about the other ones), each of the workers will first look at queue q_1, and hence the n first jobs will be dequeued from q_1, leaving the other queues unattended; the second batch of jobs will be dequeued from q_2, etc. 

With my current implementation, at Queues level, w_1 gets a job from q_1, w_2 a job from q_2, etc. which I think is more fair. 

Maybe the right path to follow (related to another issue I raised https://github.com/rq/rq/issues/1402) would be that workers listening to the same set of queues can tell the other workers to rotate the list of queues every time a job is dequeued. Some other possibility would involve creating a new class "set of workers" or "set of queues" that take care of this.

jpsca:
@bielcardona Thank you for this. Is *exactly* what I *just* needed.

selwin:
> Imagine you have a given number of workers w_1,...,w_n waiting for jobs in queues q_1,...,q_m, and the dequeue strategy is set to be round-robin. Imagine also that all queues have pending jobs. If the round-robin strategy is implemented at the worker level (and hence each worker knows nothing about the other ones), each of the workers will first look at queue q_1, and hence the n first jobs will be dequeued from q_1, leaving the other queues unattended; the second batch of jobs will be dequeued from q_2, etc.

Wouldn't the current `RoundRobinQueue.dequeue_any()` proposal have the same behavior? Each worker instance would have its own `RoundRobinQueue` so they won't know which queue other workers dequeued from last.

The real issue here is that there's no shared resource between workers. Each worker doesn't know what other workers are doing. So fundamentally to solve this we'd need to have either of these two:
1. Shared resources between workers.
2. A worker pool mechanism so that one worker can manage multiple horses.

Solution 2 is something that I think will be most beneficial to have. This solution allows us to do `rq worker --num-horses=4`. This also has the benefit of simplifying deployment of multiple workers.

bielcardona:
>  Wouldn't the current RoundRobinQueue.dequeue_any() proposal have the same behavior? Each worker instance would have its own RoundRobinQueue so they won't know which queue other workers dequeued from last.

Not with the current implementation. There is a class variable inside RoundRobinQueue that takes care of it. 

> The real issue here is that there's no shared resource between workers. Each worker doesn't know what other workers are doing. So fundamentally to solve this we'd need to have either of these two:
> 1. Shared resources between workers.
> 2. A worker pool mechanism so that one worker can manage multiple horses.

IMHO, the best solution would be to have an auxiliary class (say QueueList). An instance of that class would contain a list of queues and a parameter indicating which strategy to implement (round robin, priorities,...). Then, a worker would not listen to the queues but to this intermediate object, which would be responsible for dequeueing from the appropriate queue in each case. Note that it would also work with different workers listening to the same QueueList instance.

selwin:
> Not with the current implementation. There is a class variable inside RoundRobinQueue that takes care of it.

I don’t think a class variable is sufficient. The test case provided works because it runs in a single process. If you run multiple workers, each worker will have it’s own copy of `RoundRobinQueue` which have no shared state. As a result, all workers will dequeue from q1, on the next round all workers will dequeue from q2 and so on.

That’s not to say what we have now is no good, however. Having any kind of round robin or shuffle queue is good to have.

But like I said, it should be implemented on the worker level so it makes sense API wise.

bielcardona:
> I don’t think a class variable is sufficient. The test case provided works because it runs in a single process. If you run multiple workers, each worker will have it’s own copy of RoundRobinQueue which have no shared state. As a result, all workers will dequeue from q1, on the next round all workers will dequeue from q2 and so on.

You are right. 

However, I do not know how to implement it at worker level, since the worker relies on the dequeue_any method, which is handled by Queue (that is why I subclassed Queue). What would be a good strategy?

bielcardona:
it was easier than expected... I'll close this pull request and open a new one

Issue #1423, opened 2021-02-27
Is there a way to have queue with randomly shuffled jobs ?


vitalik:
Currently you can insert job either to the start or to the end of the queue

Is there a workaround to insert job into random position ?
(or in a worker fetch job from random position)

thomasmatecki:
You could probably implement a [`Queue`](https://python-rq.org/docs/workers/#worker-arguments) class that does this. The normal queue is implementing using Redis' `RPUSH` and `LPOP` commands. There is no Redis command to _either insert or remove a random item from the list at key X_. Perhaps one could cook up a queue backed by a Redis set and use [`SRANDMEMBER`](https://redis.io/commands/srandmember).

marians:
I think this issue was closed because someone thought that https://github.com/rq/rq/pull/1425 solved it. I just looked at the PR and my understanding is that the change affected the order in which a worker would select the queue to pick from. It doesn't affect the order of jobs within a queue.

PR #1425, opened 2021-02-28
Workers dequeuing jobs from queues using both Round-Robin and Random strategies

bielcardona:
I have subclassed Worker so that the subclass can dequeue jobs using both a round-robin strategy or a random strategy, using the method reorder_queues().
I have had to add a call to reorder_queues in the base class, which does nothing in this case.


thesamesam:
Nice! This looks really helpful. It'd also fix #1423.

jpsca:
@bielcardona  There is something wrong with the branch. GitHub only sees as changed `rq/worker.py` but not the new tests or documentation.

bielcardona:
I have added doc and tests. The tests run with no error locally, but it seems that the automatic tests run on github fail (not sure if due to my changes)

thesamesam:
Hmm. This is the failure: https://github.com/rq/rq/runs/2080885359#step:7:50. I'm not sure where "In" is coming from.

(Squashing the commits may be helpful too btw)

jpsca:
@thesamesam `float("Inf")` is how you get an "infinite" in Python, probably someone deleted the final "f" by mistake

bielcardona:
oops... sorry. I have corrected it.

aparcar:
I'm a bit confused with this concept, the worker first pics a job and then modifies the queue list according to it's *dequeue style*? So then the next worker pics the job according to the previous worker *reordering* and then reoders itself again?

bielcardona:
@aparcar 
> I'm a bit confused with this concept, the worker first pics a job and then modifies the queue list according to it's dequeue style?

Yes

> So then the next worker pics the job according to the previous worker reordering and then reoders itself again?

No. Each worker has its own queue list.

selwin:
This is a much better attempt, but instead of having two additional worker classes, wouldn't it be better to have a  `dequeue_strategy` attribute on worker?

rq/worker.py
```diff
@@ -1160,3 +1164,22 @@ def request_force_stop_sigrtmin(self, signum, frame):
...
         raise ShutDownImminentException('shut down imminent (signal: %s)' % signal_name(signum), info)
+
+
+class RoundRobinWorker(Worker):
+    """
+    Modified version of Worker that dequeues jobs from the queues using a round-robin strategy.
+    """
+    def reorder_queues(self, reference_queue):
+        pos = self.queues.index(reference_queue)
+        self.queues=self.queues[pos+1:]+self.queues[:pos+1]
```
selwin, line 1175: I don't think we should mutate `self.queues`. Can you create another property `self._ordered_queues` that we use for this purpose?

bielcardona:
> This is a much better attempt, but instead of having two additional worker classes, wouldn't it be better to have a `dequeue_strategy` attribute on worker?

Having custom worker classes makes it straightforward to use these strategies from the command line with no change to the API. You just have to create the worker with the appropriate --worker-class parameter. I think it is a simpler way to do it.

selwin:
Ok, I have no hard opinions on this. Can you implement the change to store the reordered queues in `_ordered_queues`?

rq/worker.py
selwin, line 1175: Can you add the proper spacing around `=` and `+`?

rq/worker.py
```diff
@@ -1160,3 +1164,22 @@ def request_force_stop_sigrtmin(self, signum, frame):
...
+
+
+class RandomWorker(Worker):
+    """
+    Modified version of Worker that dequeues jobs from the queues using a random strategy.
+    """
+
+    def reorder_queues(self, reference_queue):
+        from random import shuffle
```
selwin, line 1184: Can you move this import line to import section at the beginning of file?

tests/test_worker.py
```diff
@@ -1362,3 +1362,56 @@ def setUp(self):
...
+        w.work(burst=True)
+        start_times = []
+        for i in range(5):
+            for j in range(3):
+                job = Job.fetch('q%d_%d' % (i,j))
+                start_times.append(('q%d_%d' % (i,j), job.started_at))
+        sorted_by_time = sorted(start_times, key=lambda tup: tup[1])
+        sorted_ids = [tup[0] for tup in sorted_by_time]
+        expected = ['q%d_%d' % (i,j) for j in range(3) for i in range(5)]
+        self.assertEqual(expected, sorted_ids)
```
selwin, line 1386: To make this clear, can you make this more explicit? Do this:
```python
self.assertEqual(
    sorted_ids,
    ['1_1', '2_1', ...]
)
```
If the list is too long, I don't mind reducing the number of loops to 3 and 3.

tests/test_worker.py
```diff
@@ -1362,3 +1362,56 @@ def setUp(self):
...
+
+
+class TestRandomWorker(RQTestCase):
+    def test_random_worker(self):
+        qs = [Queue('q%d' % i) for i in range(5)]
+
+        for i in range(5):
+            for j in range(3):
+                qs[i].enqueue(say_pid,
+                              job_id='q%d_%d' % (i,j))
```
selwin, line 1396: Can you add space after `,`? Sorry for nitpicking, but I want the code style to be consistent.

rq/worker.py
bielcardona: No problem about adding a new property, but I would like to know where it should be used (apart from the call to `dequeue_any` in line 650). For instance, I see no point in using it in `validate_queues` (line 241), but I am not sure if it should be used in `queue_names` (line 247) or `queue_keys` (line 251). Same for the call to `RQScheduler` in line 553 and the definitions of `refresh`(line 729) and `clean_registries` (line 1088).

bielcardona: done

rq/worker.py
bielcardona: done

tests/test_worker.py
bielcardona: done

tests/test_worker.py
bielcardona: done

rq/worker.py
selwin: `_ordered_queues` is used to shuffle queue orderings. `self.queues` shouldn't be mutated.

bielcardona:
No news on this pull request? I think I fixed all the issues.

selwin:
Sorry for the delay. Will review again soon.

selwin:
Thanks!

PR #1427, opened 2021-03-02
Have hearbeats occur more frequently

### LINKED ISSUE #1426: Increasing `default_worker_ttl` does not cause more heartbeats, just one to occur later
We use a custom worker subclass which specifies `default_worker_ttl`. We also override the heartbeats method to report to a few other systems and then call `super()` so RQ can do it's thing normally. What we've noticed is that:

1. A `default_worker_ttl=30` will heartbeat 15 seconds after the previous heartbeat.
2. A `default_worker_ttl=60` will heartbeat 45 seconds after the previous heartbeat.
3. A `default_worker_ttl=120` will heartbeat 105 seconds after the previous heartbeat.

So even thought he ttl is getting longer, we aren't receiving regular heartbeats the single heartbeat is just occuring much later (15 seconds before its due).

In our case we need heartbeats to occur more regularly, not just when they are due. In reading previous code comments, @dralley pointed to this line here: https://github.com/rq/rq/pull/945/files#r185680378

What would be great if heartbeats "spread" out so that they occur for example every 15 seconds, or a configurable amount of time in between. If auto-spreading isn't acceptable, having a setting to control it would be a reasonable second choice (to me).

bmbouter:
Exactly one heartbeat occurs and it's hardcoded to 15 seconds before the
`default_worker_ttl`. For reliability heartbeats should occur more
frequently, 3-times within each TTL with an interarrival time of
`default_worker_ttl` / 3.

fixes #1426

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -568,7 +568,7 @@ class Worker(object):
                         self.log.info('Worker %s: stopping on request', self.key)
                         break
 
-                    timeout = None if burst else max(1, self.default_worker_ttl - 15)
+                    timeout = None if burst else max(1, int(self.default_worker_ttl / 3))
                     result = self.dequeue_job_and_maintain_ttl(timeout)
                     if result is None:
                         if burst:


bmbouter:
@rq/rq can someone review this change please?

selwin:
I'm not sure what problem this PR is intended to fix. At the moment `DEFAULT_WORKER_TTL` is set to 420 seconds, which means that we wait until 420 seconds before we declare a worker as dead (heartbeat has to arrive before 420 seconds).

Using this default settings, this is what a worker will do before waiting for a job to arrive:
1. [Sends a heartbeat](https://github.com/rq/rq/blob/master/rq/worker.py#L686) with ttl of 480 (420 seconds + 60 seconds buffer time to account for server hiccups etc).
2. Listens to a job on configured queues with a timeout of 405 seconds (420 - 15 seconds). If a job doesn't arrive within 405 seconds, it goes back to step 1 and repeat again.

So if you want RQ to sends heartbeat more often, you should just decrease `DEFAULT_WORKER_TTL`.

bmbouter:
@selwin ty for the feedback. In reading this, I realize I've done a not-great job of explaining the motivation here. Rather than explaining here I've revised the issue body https://github.com/rq/rq/issues/1426

Can you look at that and let me know what you think please? Thanks!

bmbouter:
Hi @selwin did my extensibility description on the issue make any more sense? I'm hoping it does. I'm also hoping this can be accepted because the downside of solving that problem is the cost of one additional heartbeat per TTL which my opinion is that's ok and worth the problems it solves. What do you think?

PR #1428, opened 2021-03-04
Handle deserializing failures gracefully

### LINKED ISSUE #1422: Pickle version incompatibility are not logged
Hi, 

I had a problem with jobs failing silently while running in Docker. I eventually realized that I was running workers in python 3.7, but was enqueuing calls from python 3.9 (an easy enough mistake to make when orchestrating different containers). 

What happens in this case is that the job gets enqueued, the worker starts execution but then the job fails. With `--logging_level DEBUG`, this is the result:

```
09:20:42 default: test_module.add2(1, 2) (586b3d13-cba9-436a-8b6f-45fc2c3bb8f0)
09:20:42 Sent heartbeat to prevent worker timeout. Next one should arrive within 480 seconds.
09:20:42 Sent heartbeat to prevent worker timeout. Next one should arrive within 90 seconds.
09:20:42 Handling failed execution of job 586b3d13-cba9-436a-8b6f-45fc2c3bb8f0
09:20:42 Sent heartbeat to prevent worker timeout. Next one should arrive within 480 seconds.
09:20:42 Sent heartbeat to prevent worker timeout. Next one should arrive within 480 seconds.
09:20:42 *** Listening on default...
```

note that there is no indication here that the job has failed. However, `job.exc_info` does show the traceback,
```
  File "/home/clavigne/miniconda3/envs/funsies3.7/lib/python3.7/site-packages/rq/worker.py", line 970, in perform_job
    self.prepare_job_execution(job, heartbeat_ttl)
  File "/home/clavigne/miniconda3/envs/funsies3.7/lib/python3.7/site-packages/rq/worker.py", line 867, in prepare_job_execution
    self.procline(msg.format(job.func_name, job.origin, time.time()))
  File "/home/clavigne/miniconda3/envs/funsies3.7/lib/python3.7/site-packages/rq/job.py", line 247, in func_name
    self._deserialize_data()
  File "/home/clavigne/miniconda3/envs/funsies3.7/lib/python3.7/site-packages/rq/job.py", line 215, in _deserialize_data
    self._func_name, self._instance, self._args, self._kwargs = self.serializer.loads(self.data)
ValueError: unsupported pickle protocol: 5
```
but this is never logged by `rq worker` itself.

I haven't had the chance to try yet, but I guess this may be fixed by #1303 ? I am also wondering if this could be the cause of errors reported in some of the comment in #758 .

clavigne:
This PR fixes #1422 . 

I've also added a test case for the bug.

The only thing that bothers me about the fix is how it will interact with custom exception handlers. The custom handlers are also passed the bad `Job` objects, and so would also need to guard against deserialization errors when accessing `Job.func_name`, `Job.args`, etc. 

Maybe the `Job` instance should be modified or something to make sure that it's properties are safe to access? Or maybe it's ok to defer to the custom exception handler writer?

Thanks.


selwin:
Thanks, for the PR! I think we should catch deserialization error earlier and explicitly call `handle_exception()` providing the cause of the exception. Could you please catch this error at the initial point where `DeserializationError` is raised?

clavigne:
Good idea, will do.

selwin:
@clavigne friendly reminder to update this if you can :)

selwin:
If I were to choose between code duplication and behavioral consistency, I think I'd choose consistency in this case. I'd rather that calling `job._func_name` raises `DeserializationError` everytime so it's idempotent and behaves predictably.

Do you mind updating the code to reflect this? I think your [initial approach](https://github.com/rq/rq/pull/1428/commits/e0570bac9f84d4da2ba1b114c17042a364be5435) is close enough, but it should catch `DeserializationError` so we can properly log the error differently.

rq/worker.py
```diff
@@ -1040,13 +1040,22 @@ def perform_job(self, job, queue, heartbeat_ttl=None):
...
+
+        # If the job cannot be deserialized, it will raise when func_name or
+        # the other properties are accessed, which will stop exceptions from
+        # being properly logged, so we guard against it here.
+        try:
+            extra = {
+                'func': job.func_name,
+                'arguments': job.args,
+                'kwargs': job.kwargs}
+        except: # noqa
```
selwin, line 1052: We should explicitly catch `DeserializationError` over here, to be explicit.

clavigne:
Hi @selwin , thank you so much for your patience.

I added a new `DeserializationError` exception to deal with the fact that serializers are generic, and made deserialization raise it (using `raise DeserializationError() from e`). This makes the exception catching specific without making it break on user-defined deserializers.

Do you think this is a satisfying solution? thanks again for your feedback.

selwin:
@clavigne this approach looks great! Do you mind fixing CI failures so I can merge this in?

clavigne:
I previously forgot to do a local test run. I fixed the failing test and everything should work now.

thanks for the help!

clavigne:
@selwin This should be ready by the way, all tests pass locally.

Thanks!

PR #1431, opened 2021-03-12
Gracefully handle both str and bytes in str_to_date

twelvelabs:
This PR restores the ability of `str_to_date` to handle timestamps encode as either `str` or `bytes`. This behavior was removed in #1299 (specifically commit b33f3b8a6667d0d6692b502e4b69941c19f85a24) but that breaks backwards compatibility. We have some custom logic that uses `str_to_date` to store extra timestamps in the Job meta. Those timestamps are stored as `str` and upgrading to the latest version of `rq` broke that functionality.

I ran `run_tests_in_docker.sh` and all tests are passing.

### DIFF
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -230,7 +230,7 @@ def str_to_date(date_str):
     if not date_str:
         return
     else:
-        return utcparse(date_str.decode())
+        return utcparse(as_text(date_str))
 
 
 def parse_timeout(timeout):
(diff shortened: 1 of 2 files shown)


rq/utils.py
```diff
@@ -230,7 +230,7 @@ def str_to_date(date_str):
     if not date_str:
         return
     else:
-        return utcparse(date_str.decode())
+        return utcparse(as_text(date_str))
```
selwin, line 233: `as_text` is a Python 2/3 compatibility function that will be removed in the future. Could you build string handling directly in this function instead of using `as_text`?

selwin:
@twelvelabs please update this when you can :)

PR #1439, opened 2021-03-31
Newer pip install command from git

jammie19:
Updated cutting edge pip install command to use HTTPS rather than insecure git+git protocol, as recommended by pip documentation. This allows the command to work with pip >= 21.0.1
This fixes issue #1439


selwin:
Thanks!

Issue #1457, opened 2021-05-04
Can't enqueue staticmethods


pwws:
Worker's can't run class methods decorated with `staticmethod`. Consider the following code and let's assume it's in `staticmethod_test` module:
```python
class TestClass:
    @staticmethod
    def test_job():
        return 'Test static method'
```
Enqueuing `TestClass.test_job` is successful:
```python
queue.enqueue_call(staticmethod_test.TestClass.test_job)
# Job('e8fe13af-b996-401e-a4a9-431e72c5bfa4', enqueued_at=datetime.datetime(2021, 5, 4, 10, 55, 45, 44138))
```
but a `worker` can't handle it:
```python
12:55:45 default: staticmethod_test.test_job() (e8fe13af-b996-401e-a4a9-431e72c5bfa4)
12:55:45 Traceback (most recent call last):
  File "/home/xxx/prog/rq/rq/worker.py", line 1021, in perform_job
    rv = job.perform()
  File "/home/xxx/prog/rq/rq/job.py", line 707, in perform
    self._result = self._execute()
  File "/home/xxx/prog/rq/rq/job.py", line 730, in _execute
    result = self.func(*self.args, **self.kwargs)
  File "/home/xxx/prog/rq/rq/job.py", line 219, in func
    return import_attribute(self.func_name)
  File "/home/xxx/prog/rq/rq/utils.py", line 130, in import_attribute
    return getattr(module, attribute)
AttributeError: module 'staticmethod_test' has no attribute 'test_job'
```

Apparently, I'm not the only one that had this problem: https://github.com/rq/rq/issues/582#issuecomment-368455932, https://github.com/rq/django-rq/issues/248

pwws:
I have prepared a fix for this problem: https://github.com/rq/rq/pull/1458

selwin:
Fixed in #1458

PR #1458, opened 2021-05-04
bugfix: Allow using staticmethods as jobs

pwws:
This PR fixes a problem reported in https://github.com/rq/rq/issues/1457 - it allows to enqueue class functions decorated with `staticmethod`, like `test_job`:

```python
class TestClass:
    @staticmethod
    def test_job():
        return 'Test static method'
```

All the existing tests and a newly created test case pass.

Unfortunately, I see at least one case, in which `staticmethod`s still can't be enqueued - it's when they're members of nested classes, like `nested_test_job`:
```python
class TestClass:
    class NestedClass:
        @staticmethod
        def nested_test_job():
            return 'Test static method'
```

I haven't implemented handling them as I'm not sure, if it's worth the effort - what do you think? :)

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -108,7 +108,7 @@ class Job(object):
             job._instance = func.__self__
             job._func_name = func.__name__
         elif inspect.isfunction(func) or inspect.isbuiltin(func):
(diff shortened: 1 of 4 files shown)


selwin:
Thanks!

PR #1459, opened 2021-05-05
Replace enum function with internal Enum class

MyrikLD:
Removed deprecated (object) inheritance
Add new py38,py39 versions to tox, removed deprecated py27,py34
Replace enum internal function with Enum class


MyrikLD:
Before:
Screenshot_2021-05-05_16-47-02
After:
Screenshot_2021-05-05_16-50-40

selwin:
Thanks!

PR #1462, opened 2021-05-07
Feature/cluster

TimoFrenzel:
Connection type added to distinguish Redis version detection. In addition to using a single Redis server, it is now also possible to use a real Redis cluster.


TimoFrenzel:
wrong pullrequest

PR #1465, opened 2021-05-12
Support for Click 8.0.0

### LINKED ISSUE #1464: rq no longer works with Click 8.0+
When trying to run `rq worker` now throws a `ValueError`. Here is the full traceback:
```
  File "/usr/local/bin/rq", line 5, in <module>
    from rq.cli import main
  File "/usr/local/lib/python3.9/site-packages/rq/cli/__init__.py", line 2, in <module>
    from .cli import main
  File "/usr/local/lib/python3.9/site-packages/rq/cli/cli.py", line 99, in <module>
    def empty(cli_config, all, queues, **options):
  File "/usr/local/lib/python3.9/site-packages/rq/cli/cli.py", line 77, in pass_cli_config
    func = option(func)
  File "/usr/local/lib/python3.9/site-packages/click/decorators.py", line 247, in decorator
    _param_memo(f, OptionClass(param_decls, **option_attrs))
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 2467, in __init__
    super().__init__(param_decls, type=type, multiple=multiple, **attrs)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 2106, in __init__
    raise ValueError(
ValueError: 'default' must be a list when 'multiple' is true.
```

This seems to be an issue with [Click 8.0](https://click.palletsprojects.com/en/8.0.x/changes/#version-8-0-0) which was released yesterday. A temporary downgrade to version 7.1.2 can be used as a temporary workaround.

rpkak:
[click Version 8.0.0](https://github.com/pallets/click/blob/main/CHANGES.rst#version-800) is not compatible with this rq version.

I fixed #1464 in a590bc71bb9d9053b8e10c18d690a0a70439727a, but I found another error caused by pallets/click#1146 (**Edit**: now fixed by 55f4489e76b6c81d2f6905a2455390ea131554f4).

Fix #1464


selwin:
@rpkak can you make the tests pass so I can merge this in? If possible I’d like to make a release this weekend.

rpkak:
The tests on the master branch are failing since 76ac0afbcd07c2a13b17ba92ddb63ca44e3b25aa but I can try to fix them.

rpkak:
Using pytest local does not return any errors, but
```
$ RUN_SLOW_TESTS_TOO=1 pytest --cov=./ --cov-report=xml --durations=5
==================================================================================== test session starts ====================================================================================
platform linux -- Python 3.8.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /home/arne/git/rq
plugins: cov-2.11.1
collected 293 items                                                                                                                                                                         

tests/test_cli.py ..................                                                                                                                                                  [  6%]
tests/test_commands.py ...                                                                                                                                                            [  7%]
tests/test_connection.py .....                                                                                                                                                        [  8%]
tests/test_decorator.py .............                                                                                                                                                 [ 13%]
tests/test_fixtures.py ..                                                                                                                                                             [ 13%]
tests/test_helpers.py .                                                                                                                                                               [ 14%]
tests/test_job.py ............................................................................                                                                                        [ 40%]
tests/test_queue.py ............................................                                                                                                                      [ 55%]
tests/test_registry.py ........................                                                                                                                                       [ 63%]
tests/test_scheduler.py ..............s...                                                                                                                                            [ 69%]
tests/test_sentry.py ...                                                                                                                                                              [ 70%]
tests/test_serializers.py .                                                                                                                                                           [ 70%]
tests/test_utils.py ............                                                                                                                                                      [ 75%]
tests/test_worker.py .....................................................................                                                                                            [ 98%]
tests/test_worker_registration.py ....                                                                                                                                                [100%]

===================================================================================== warnings summary ======================================================================================
tests/test_cli.py::TestRQCli::test_info
tests/test_cli.py::TestRQCli::test_info
tests/test_cli.py::TestRQCli::test_info_only_queues
tests/test_cli.py::TestRQCli::test_info_only_queues
  /home/arne/git/rq/rq/cli/helpers.py:92: DeprecationWarning: 'click.get_terminal_size()' is deprecated and will be removed in Click 8.1. Use 'shutil.get_terminal_size()' instead.
    termwidth, _ = click.get_terminal_size()

-- Docs: https://docs.pytest.org/en/stable/warnings.html

----------- coverage: platform linux, python 3.8.5-final-0 -----------
Coverage XML written to file coverage.xml

==================================================================================== slowest 5 durations ====================================================================================
66.41s call     tests/test_worker.py::WorkerShutdownTestCase::test_work_horse_force_death
21.11s call     tests/test_worker.py::TestWorker::test_heartbeat_busy
6.66s call     tests/test_worker.py::TestWorker::test_custom_exc_handling
5.24s call     tests/test_scheduler.py::TestWorker::test_work
4.46s call     tests/test_worker.py::TestWorker::test_disable_default_exception_handler
================================================================== 292 passed, 1 skipped, 4 warnings in 168.67s (0:02:48) ===================================================================
```

bmbouter:
What about adding version restricting click using `~=` in the requirements file so when click 9 comes out everyone doesn't break?

Also our users are broken until this is merged and released. Thanks for fixing this!

rpkak:
Othervise we could make a test, where we use the build from the development branches of the repositorys of click and redis to get the latest changes before the release of new versions of click and redis.

selwin:
Thanks!

bmbouter:
Thanks for merging! Do you know what release this will come it, and when that release will occur?

selwin:
Hey there, I’ll try to cut a release by Monday.

davidhariri:
Thanks for fixing! Why does this project not pin its dependencies?

selwin:
Because pinning everything may cause conflicts :). If all your third party packages unnecessarily pin their dependencies, pretty soon you’ll run into dependency hell.

davidhariri:
Ah, ok. I see what you mean. If two or more dependencies need the same packages but with different versions? Makes sense.

bmbouter:
I agree with you that pinning is not viable, but when dependencies declare they adhere to semver, I believe the best way to serve users is to prevent the next major version from automatically being brought in. This would be with something like `click~=7.12`. As the click release demonstrated, allowing the next major version of a semver dependency to be brought anytime its released is unsafe.

selwin:
> I agree with you that pinning is not viable, but when dependencies declare they adhere to semver, I believe the best way to serve users is to prevent the next major version from automatically being brought in. This would be with something like click~=7.12.

Well, if I release a new RQ version with click~=8.0`. All of a sudden you won't be able to use RQ if you depend on another package that specifies `click~=7.12` as a dependency (the package owner won't be in a rush to fix it either, because it still works!).

Having said that, I'm not against version pinning (RQ requires redis >= 3.5 and click >= 5), but I mainly use it to drop support for old versions if:
1. Maintenance burden becomes too high.
2. The package is sufficiently old (more than 5 years usually)

> As the click release demonstrated, allowing the next major version of a semver dependency to be brought anytime its released is unsafe.

I agree with you, but we'd be in dependency hell if popular libraries become overly strict with their dependency requirement :). If you want to be more strict, you can pin click to a specific version in your own project's `requirements.txt` and manage the dependencies yourself.

If I'm overly strict with this, it will make life difficult for downstream users. Anyway, there are pros and cons to all approaches, and this is simply one I'm most comfortable with. The previous time this happened was when click 3.0 was released [back in 2015](https://github.com/rq/rq/commit/10f615f4fef8b033e4d866a0f6e486ba782d76e1#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552) so this is really not that bad.

bmbouter:
Thank you for the thoughtful reply. I can tell you're thinking carefully about how to best approach this. My `click~=8.0` proposal has that compatibility drawback, but what about a `click>=5,<9`?

The downside of not doing this is significant. When click released, many production systems broke. It doesn't have to be that way; this is why semver exists.

Also let's consider the two possible outcomes here. My usage falls in the situation where we don't depend on click. Who is in the best position to declare compatibility with it? I believe that is the library who has the dependency on it, in this case RQ. Due to semver, RQ should not declare forward compatibility above its major release.

Let's consider the other case. If we did have a dependency on click, we should declare our min/max also. It's would be unsafe for us to declare major-version forward compatibility on click just like it's unsafe for RQ. If after everyone declaring their actual dependencies conflicts arise, it's due to actual conflict, not arbitrary conflict.

If you still disagree with this thinking, as the project leader, it's your choice. Either way, I'll leave it by saying thank you for your time and for all the effort you've put into making RQ great for us to use.

rpkak:
We also could make a workflow to test rq with the main/master branches of redis-py and click daily.

With this we could have caught these errors on 2020/08/08 when pallets/click#1146 was merged and on 2020/03/05 when pallets/click#1805 was merged and could have fixed them until click 8.0.0 was realesed.

Ping me if I should make a workflow like this.

selwin:
> My click~=8.0 proposal has that compatibility drawback, but what about a click>=5,<9?

This would still create conflicts when future (potentially backward incompatible versions are released). For example, if Flask 3.0 is released and requires `click>=9`, users won't be able to use RQ that depends on `click>=5,<9` without us releasing a new version only to bump the dependency requirement (remember that click's past major releases from 3.X to 7.X require no code changes in RQ).

Like I previously said, no one approach is 100%. I just choose this method to be looser in defining dependencies, at least it gives users the option to be stricter in specifying dependencies in their own projects.

Thanks for the input :)

> We also could make a workflow to test rq with the main/master branches of redis-py and click daily.

With this we could have caught these errors on 2020/08/08 when pallets/click#1146 was merged and on 2020/03/05 when pallets/click#1805 was merged and could have fixed them until click 8.0.0 was realesed.

@rpkak thanks for the idea. How should we proceed with this? Can you create a PR with a Github workflow definition that I can pull in?

PR #1466, opened 2021-05-14
Allows enqueueing by the cli

### LINKED ISSUE #372: Script to enqueue command
[django-rq](https://github.com/ui/django-rq) now has a [rqenqueue](https://github.com/ui/django-rq/blob/master/django_rq/management/commands/rqenqueue.py) command that allows you to easily schedule jobs via command line:

``` python
python manage.py rqenqueue queue_name "path.to.job" "arg1" "arg2"
```

We'll obviously need to make some changes to allow users to specify queue configurations but I think this will be useful to have in RQ.

rpkak:
closes #372


selwin:
This is a good idea :). Feel free to ping me again when this PR is finished.

rpkak:
Ok @selwin, it's finished.

selwin:
@rpkak I'll review this when I have more time in the coming days. Thanks for working on this!

rq/cli/cli.py
```diff
@@ -303,3 +306,51 @@ def resume(cli_config, **options):
...
+@click.option('--failure-ttl', help='Specifies how long failed jobs are kept.')
+@click.option('--description', help='Additional description of the job')
+@click.option('--depends-on', help='Specifies another job id that must complete before this job will be queued.',
+              multiple=True)
+@click.option('--job-id', help='The id of this job')
+@click.option('--at-front', is_flag=True, help='Will place the job at the front of the queue, instead of the back')
+@click.option('--retry-max', help='Maximum amound of retries', default=0, type=int)
+@click.option('--retry-interval', help='Interval between retries in seconds', multiple=True, type=int, default=[0])
+@click.option('--schedule-in', help='Timedelta in what the function is enqueued.')
+@click.option('--schedule-at', help='Time in what the function is enqueued.')
```
selwin, line 326: Can we add an example of what kind of format is expected here?

rq/cli/cli.py
```diff
@@ -303,3 +306,51 @@ def resume(cli_config, **options):
...
+@click.option('--ttl', help='Specifies the maximum queued time of the job before it’s discarded.')
+@click.option('--failure-ttl', help='Specifies how long failed jobs are kept.')
+@click.option('--description', help='Additional description of the job')
+@click.option('--depends-on', help='Specifies another job id that must complete before this job will be queued.',
+              multiple=True)
+@click.option('--job-id', help='The id of this job')
+@click.option('--at-front', is_flag=True, help='Will place the job at the front of the queue, instead of the back')
+@click.option('--retry-max', help='Maximum amound of retries', default=0, type=int)
+@click.option('--retry-interval', help='Interval between retries in seconds', multiple=True, type=int, default=[0])
+@click.option('--schedule-in', help='Timedelta in what the function is enqueued.')
```
selwin, line 325: This should specify that `Schedule job to be enqueued in X seconds`.

rq/cli/cli.py
selwin: Something like `Schedule job to be enqueued at a certain time e.g <format>`

rq/cli/helpers.py
```diff
@@ -198,6 +200,74 @@ def setup_loghandlers_from_args(verbose, quiet, date_format, log_format):
...
 
+def job_func(func, args, kwargs):
+    try:
+        importlib.import_module(func)  # execute file
+    except ImportError:
+        return import_attribute(func)(*args, **kwargs)
+
+
+def cast_element(element, arg_type):
+    if arg_type == bool:
```
selwin, line 211: When do we need to cast booleans from CLI arguments? I think in CLI context we should use something like `--at-front` so we shouldn't need to parse `true` or `false`

rpkak: `cast_element` is not used for something like `--at-front`, but is used [here](https://github.com/rq/rq/blob/809395edbc10c3d46f7477937abe6e80a388479e/rq/cli/helpers.py#L244) for casting the args of the function you want to call.

```
rq enqueue path.to.func --boolean true
```

will call `path.to.func` with the boolean `True` as arg.

selwin: So what I mean is that we shouldn't have a CLI argument with this kind of signature.

* `rq enqueue path.to.func --at-front true`
* `rq enqueue path.to.func --at-front false`

The two commands above can be better expressed as `rq enqueue path.to.func --at-front`

So I don't want to support a use case where the user passed a boolean as CLI argument.

rpkak: I have implemented `--at-front` and `--quiet` as simple click.option with `is_flag=true`.

So if you want to enqueue at front you can call:

```
eq enqueue path.to.func --at-front
```

If you want to enqueue at front with `False` as an argument, you can call:

```
eq enqueue path.to.func --at-front --bool false
```

rpkak:
Any changes for this PR?

rq/cli/helpers.py
selwin: You shouldn't need to call `rq enqueue path.to.func --at-front --bool false` because this is the default behavior.

So if you want to enqueue at front you can call `enqueue path.to.func --at-front` (like you mentioned). Otherwise it should be simply `rq enqueue path.to.func`, no boolean argument necessary.

rpkak: `--boolean` is used for the same as `--int` or `--float`.

`--boolean` is used if you want to add an argument to the function you want to enqueue.

How you can see here `--at-front` is a click option and `--boolean` has no effect to it.

https://github.com/rq/rq/blob/2741c3bf50d4b2a2fae5051690a2c31d552c68b3/rq/cli/cli.py#L322

Here a simple example:

Python:
```python
def func(arg=None):  # path.to.func
    print(arg)
    print(type(arg))
```


Then execute:

```
$ rq enqueue path.to.func --boolean false
```
And it returns:
```
12:14:18 default: path.to.func(False) (97fbe646-4414-427d-9dac-0e3b136d9208)
False
<class 'bool'>
12:14:18 default: Job OK (97fbe646-4414-427d-9dac-0e3b136d9208)
12:14:18 Result is kept for 500 second
```

So you see, that the argument is `False` as a boolean.

But executing:
```
$ rq enqueue path.to.func
```
returns:
```
12:15:58 default: path.to.func() (6be872b6-aaf6-4cf7-b354-84c7229b9827)
None
<class 'NoneType'>
12:15:58 default: Job OK (6be872b6-aaf6-4cf7-b354-84c7229b9827)
12:15:58 Result is kept for 500 seconds
```

So this time there is no argument so the python functions is called without any args and `None`, specifiered in the header of the function, is taken as the default arg of the function.

For all this examples you could add `--at-front` without anything after it to enqueue the job at front.

selwin: Sorry for being dense, now I get it. So the problem is actually passing args/kwargs to the callable itself.

Are you ok if we merge this in without the capability to pass in args/kwargs first to the callable?

If we do support this, I want our CLI API to be similar to https://httpie.io . Take a look at how they handle [non string JSON fields](https://httpie.io/docs#non-string-json-fields).

I'm not sure how easy or difficult it would be to achieve though.

rpkak: Ok, I implement:

<table><thead><tr><th></th><th>keyword</th><th>no keyword</th></tr></thead><tbody><tr><td>string</td><td rowspan="2">Like <a href="https://httpie.io/docs#non-string-json-fields">this</a></td><td>Just the value</td></tr><tr><td>json</td><td><code>:value</code><br>(e.g. <code>:1</code> would be <code>1</code> as an integer.)</td></tr></tbody></table>

rpkak: With `\` as escape char

rpkak: So `:` and `=` are not allowed to be in keywords

selwin: One way we can go about this is to use [ast.literal_eval()](https://docs.python.org/3/library/ast.html#ast.literal_eval) to safely evalute inputs.

`literal_eval()` supports strings, bytes, numbers, tuples, lists, dicts, sets, booleans, and None so this should be good enough. So we basically support whatever `literal_eval()` supports.

selwin: Do you mind documenting this feature too so people know how to use it? The docs are under `/docs` directory.

selwin: Thinking further about this, can we just pass every args/kwargs passed to `enqueue` via `literal_eval()`? So we can avoid the unfamiliar `:=` altogether.

selwin: The goal is to be able to do something like this:

```sh
rq enqueue path.to.func --at-front --result-ttl 60 my_number=1 my_string=abc123
```

`at-front` and `result-ttl` are `enqueue` arguments while `my_number` and `my_string` will be passed to `func`.

rpkak: The problem with `literal_eval()` is that passing `abc123` fails, because it should be `'abc123'`. But passing this in e.g. bash or zsh fails as well because the shell parses `'abc123'` to `abc123`. So you would have to pass `\'abc123\'` or `'"abc123"'`. So I think the current version with `key:=json_value` or `:json_value` is more understandable and simpler to use.

> The goal is to be able to do something like this:
> 
> ```shell
> rq enqueue path.to.func --at-front --result-ttl 60 my_number=1 my_string=abc123
> ```
> 
> `at-front` and `result-ttl` are `enqueue` arguments while `my_number` and `my_string` will be passed to `func`.

After the latest commit that should work except that `my_number` will be parsed as a string so you have to change this to `my_number:=1`.

rpkak: Please write me if I should change `json.loads` to `ast.literal_eval`.

With that you would have the `:=` as well, but you would have access to for example sets or tuples.

rpkak: > Do you mind documenting this feature too so people know how to use it? The docs are under `/docs` directory.

No, but after we discussed how this should actually work.

rpkak: I think the best would be to support both `json.loads` and `ast.literal_eval` because `ast.literal_eval` is more functional but `json.loads` is more standardized. Maybe `ast.literal_eval` could be accessed with changeing `key#=value` and `#value`.

rq/cli/helpers.py
```diff
@@ -198,6 +201,54 @@ def setup_loghandlers_from_args(verbose, quiet, date_format, log_format):
...
+def job_func(func, args, kwargs):
+    try:
+        importlib.import_module(func)  # execute file
+    except ImportError:
+        return import_attribute(func)(*args, **kwargs)
+
+
+def parse_function_arg(argument):
+    if argument.startswith(':'):  # no keyword, json
+        return None, loads(argument[1:])
```
selwin, line 213: Can we use [literal_eval()](https://docs.python.org/3/library/ast.html#ast.literal_eval) instead of JSON to evaluate the arguments?

It's better this way since this is native to Python so we can do:
```
{"key": True}
```
Instead of converting it to JSON like such:
```
{"key": True}
```

I know I suggested we look to httpie for inspiration, but they chose JSON because it's native to how HTTP works.

rpkak: > Can we use [literal_eval()](https://docs.python.org/3/library/ast.html#ast.literal_eval) instead of JSON to evaluate the arguments?

Like I mentioned [here](https://github.com/rq/rq/pull/1466#discussion_r651275396) I would prefer this updated version of [this comment](https://github.com/rq/rq/pull/1466#discussion_r650510889) (Not implemented yet):

This would support both `json.loads` and `ast.literal_eval`.

| | keyword | no keyword |
|-|-|-|
| string | `[key]=[value]` | `[value]`<br>Should not be used if you can't ensure, that the string doesn't contains `:`, `=` or `@`. |
| json | `[key]:=[value]` | `:[value]` |
| ast.literal_eval | `[key]#=[value]` | `#[value]` |

Please write me if I souldn't implement this.

> 
> It's better this way since this is native to Python so we can do:
> 
> [...]
> 
> I know I suggested we look to httpie for inspiration, but they chose JSON because it's native to how HTTP works.

That's true, but if you call this command from another programm, because it is not written in python, I don't think there's any python-eval serializer, but there are json serializers in many programming languages.

rpkak:
@selwin Can you please write me if it's ok now. I implement [this](https://github.com/rq/rq/pull/1466#discussion_r652731913) so both json and literal eval are supported.

rq/cli/helpers.py
```diff
@@ -198,6 +202,82 @@ def setup_loghandlers_from_args(verbose, quiet, date_format, log_format):
...
 
+def job_func(func, args, kwargs):
+    try:
+        importlib.import_module(func)  # execute file
+    except ImportError:
+        return import_attribute(func)(*args, **kwargs)
+
+
+def parse_function_arg(argument):
+    mode = 0  # 0: text, 1: json, 2: literal_eval (python)
```
selwin, line 213: We should just remove modes 1 (plain text) and 2 (json). We shouldn't need to give other options, since we're executing Python code, using `literal_eval()` makes the most sense. This should simplify both the implementation and documentation.

We just need to explain that all args/kwargs will be interpreted as Python.

rpkak: Yes we are executing python, but this cli may be called by other programms (not just shells), which have native json support. Also using strings if you are calling from shell is not very convenient. For example the literal-eval of `'abc'` is `abc` but if your just executing `echo 'test'` in your bash or zsh, you see that your shell parses this to just `test` (not `'test'`) which will throw an error.

selwin: > Yes we are executing python, but this cli may be called by other programms (not just shells), which have native json support.

Yes but the enqueue process will be done from within Python anyway, right? So I don't think supporting JSON makes sense here as they need to be loaded from JSON string to Python objects anyway. So I'm firm on removing JSON support here as it doesn't make sense to have this layer of indirection.

> Also using strings if you are calling from shell is not very convenient. For example the literal-eval of 'abc' is abc but if your just executing echo 'test' in your bash or zsh, you see that your shell parses this to just test (not 'test') which will throw an error.

This is a good point, since this is a CLI, we want this to behave well in the CLI world. Then let's just allow two flavors, text and Python which is evaluated using `literal_eval()`.

Thanks for your input and hard work, sorry this is taking a while.

rpkak: > Yes but the enqueue process will be done from within Python anyway, right?

I don't think so. From python you wouldn't call the cli, because you have the `queue.enqueue` method.

But if you want to enqueue a function for example with node.js (or most other languages), you could make a json from your object and pass it to the cli.

selwin: Alright then, I agree about this in principle, let me review this again to see where we can streamline things.

docs/docs/index.md
```diff
@@ -103,6 +103,64 @@ q.enqueue('my_package.my_module.my_func', 3, 4)
...
+* Execute a function: dot-separated string of package, module and function (Just like
+    passing a string to `queue.enqueue()`).
+* Execute a python file: dot-separated pathname of the file. Because it is technically
+    an import `__name__ == '__main__'` will not work.
+
+#### Arguments:
+
+| | plain text | json | [literal-eval](https://docs.python.org/3/library/ast.html#ast.literal_eval) |
+|-|-|-|-|
+| keyword | `[key]=[value]` | `[key]:=[value]` | `[key]#=[value]` |
```
selwin, line 150: Can we give a few examples of how we can enqueue things with JSON or Python objects?

rq/cli/helpers.py
```diff
@@ -198,6 +202,82 @@ def setup_loghandlers_from_args(verbose, quiet, date_format, log_format):
...
+        if keyword is not None:
+            if keyword in kwargs:
+                raise click.BadParameter('You can\'t specify multiple values for the same keyword.')
+            kwargs[keyword] = value
+        else:
+            args.append(value)
+    return args, kwargs
+
+
+def parse_schedule(schedule_in, schedule_at):
```
selwin, line 272: Can we have a test case for this utility function?

rq/cli/cli.py
```diff
@@ -303,3 +307,53 @@ def resume(cli_config, **options):
...
+    if retry_max > 0:
+        retry = Retry(retry_max, retry_interval)
+
+    schedule = parse_schedule(schedule_in, schedule_at)
+
+    with Connection(cli_config.connection):
+        queue = cli_config.queue_class(queue)
+
+        if schedule is None:
+            job = queue.enqueue_call(job_func, (function, args, kwargs), {}, timeout, result_ttl, ttl, failure_ttl,
```
selwin, line 351: I'm not sure why we're enqueueing `job_func`, we should enqueue `function` that's passed in by the user instead.

rq/cli/helpers.py
selwin, line 213: Let's create a `ParsingMode` enum for this.

rq/cli/helpers.py
```diff
@@ -198,6 +202,82 @@ def setup_loghandlers_from_args(verbose, quiet, date_format, log_format):
...
 
 
+def job_func(func, args, kwargs):
+    try:
+        importlib.import_module(func)  # execute file
+    except ImportError:
+        return import_attribute(func)(*args, **kwargs)
+
+
+def parse_function_arg(argument):
```
selwin, line 212: We'll need a testcase for this function that thoroughly tests all possibilities so we know this is correct.

rq/cli/helpers.py
```diff
@@ -198,6 +202,82 @@ def setup_loghandlers_from_args(verbose, quiet, date_format, log_format):
...
+
+def parse_function_args(arguments):
+    args = []
+    kwargs = {}
+
+    for argument in arguments:
+        try:
+            keyword, value = parse_function_arg(argument)
+        except JSONDecodeError:
+            raise click.BadParameter('Couldn\'t parse json.')
```
selwin, line 259: `Unable to parse <keyword> as JSON`.

rq/cli/helpers.py
```diff
@@ -198,6 +202,82 @@ def setup_loghandlers_from_args(verbose, quiet, date_format, log_format):
...
+        except (ValueError, SyntaxError):
+            raise click.BadParameter(
+                'Could not eval. View: https://docs.python.org/3/library/ast.html#ast.literal_eval')
```
selwin, line 262: `Unable to eval <keyword> as Python object. See https://docs.python.org/3/library/ast.html#ast.literal_eval`

rq/cli/cli.py
rpkak: If you look at `job_func`, you see that it allows you to execute a python file

docs/docs/index.md
rpkak: Where do you mean? In the docs?

selwin: Yes.

rq/cli/cli.py
selwin: Then we should build this into `enqueue_call` proper. I think it's better if we merge this in in a separate PR.

rq/cli/helpers.py
rpkak: I changed `#` to `%` because # will start a comment in .

rq/cli/cli.py
rpkak: I don't think enqueueing an executable python file from python does make much sence so I implemented this only for the cli.

@selwin Please write if I should remove this from this PR.

selwin: Sorry for the late reply. Yes, I think better to remove the ability to enqueue an executable Python file. Thanks.

docs/docs/index.md
```diff
@@ -201,6 +201,84 @@ def report_failure(job, connection, type, value, traceback):
...
+
+Where `[key]` is the keyword and `[value]` is the value which is parsed with the corresponding
+parsing method.
+
+If the first character of `[value]` is `@` the subsequent path will be read.
+
+##### Examples:
+
+* In `rq enqueue path.to.func abc` the argument will be passed as a string without keyword.
+* In `rq enqueue path.to.func abc=def` `def` will be passed as a string with `abc` as keyword.
```
selwin, line 259: I still find this to be confusing to read, maybe we show the resulting call in Python? So:
`rq enqueue path.to.func abc=def` -> `queue.enqueue(path.to.func, abc="def")`

rpkak: Only this line or all lines?

rpkak: There's not an python oneliner option for all the examples.

selwin:
Sorry for the late reply, I've been sick for the past few weeks.

docs/docs/index.md
```diff
@@ -201,6 +201,84 @@ def report_failure(job, connection, type, value, traceback):
...
+* In `rq enqueue path.to.func 'key%=1, 2'` the argument will be parsed using literal-eval to the tuple
+    `(1, 2)` and passed with `key` as keyword.
+* `rq enqueue path.to.func @path/to/file` will read the content of `path/to/file` and pass it as a
+    string without keyword.
+* `rq enqueue path.to.func key=@path/to/file` will read the content of `path/to/file` and pass it as a
+    string with `key` as keyword.
+* `rq enqueue path.to.func :@path/to/file.json` will read the content of `path/to/file.json`, parse it
+    and pass it without keyword.
+* `rq enqueue path.to.func key:=@path/to/file.json` will read the content of `path/to/file.json`, parse
+    it and pass it with `key` as keyword.
```
selwin, line 275: Can we give a Python equivalent for each example? Like what you did in point number 2, the format is great (i.e `rq enqueue path.to.func ... -> queue.enqueue(...)`

docs/docs/index.md
```diff
@@ -201,6 +201,75 @@ def report_failure(job, connection, type, value, traceback):
...
+* `rq enqueue path.to.func abc` -> `queue.enqueue(path.to.func, args=['abc'])`
+* `rq enqueue path.to.func abc=def` -> `queue.enqueue(path.to.func, kwargs={'abc': 'def'})`
+* `rq enqueue path.to.func ':{"json": "abc"}'` -> `queue.enqueue(path.to.func, args=[{'json': 'abc'}])`
+* `rq enqueue path.to.func 'key:={"json": "abc"}'` -> `queue.enqueue(path.to.func, kwargs={'key': {'json': 'abc'}})`
+* `rq enqueue path.to.func '%1, 2'` -> `queue.enqueue(path.to.func, args=[(1, 2)])`
+* `rq enqueue path.to.func 'key%=1, 2'` -> `queue.enqueue(path.to.func, kwargs={'key': (1, 2)})`
+* `rq enqueue path.to.func @path/to/file` -> `queue.enqueue(path.to.func, args=[open('path/to/file', 'r').read()])`
+* `rq enqueue path.to.func key=@path/to/file` -> `queue.enqueue(path.to.func, kwargs={'key': open('path/to/file', 'r').read()})`
+* `rq enqueue path.to.func :@path/to/file.json` -> `queue.enqueue(path.to.func, args=[json.loads(open('path/to/file.json', 'r').read())])`
+* `rq enqueue path.to.func key:=@path/to/file.json` -> `queue.enqueue(path.to.func, kwargs={'key': json.loads(open('path/to/file.json', 'r').read())})`
```
selwin, line 267: Looking at the examples, I think the way CLI arguments are parsed is confusing. We shouldn't implicitly put CLI arguments into `args` or `kwargs`.

For example:

`rq enqueue path.to.func abc` should translate to  `queue.enqueue(path.to.func, 'abc')` (currently `queue.enqueue(path.to.func, args=['abc'])`)

`rq enqueue path.to.func 'key:={"json": "abc"}'` should translate to `queue.enqueue(path.to.func, key={'json': 'abc'})`

So if I wanted ``queue.enqueue(path.to.func, kwargs={'json': 'abc'})`` as a result, I should be doing this on the CLI: 
`rq enqueue path.to.func kwargs:='{"json": "abc"}'`

rpkak: Because of [parse_args()](https://github.com/rq/rq/blob/5cbf16a75922cd64eebbb7f52002c0feaf9bae09/rq/queue.py#L455) all my examples work and `queue.enqueue(path.to.func, kwargs={'json': 'abc'})` has the same result as `queue.enqueue(path.to.func, json='abc')`:

https://github.com/rq/rq/blob/5cbf16a75922cd64eebbb7f52002c0feaf9bae09/rq/queue.py#L486-L489

But I change it

docs/docs/index.md
```diff
@@ -201,6 +201,75 @@ def report_failure(job, connection, type, value, traceback):
...
+parsing method.
+
+If the first character of `[value]` is `@` the subsequent path will be read.
+
+##### Examples:
+
+* `rq enqueue path.to.func abc` -> `queue.enqueue(path.to.func, 'abc')`
+* `rq enqueue path.to.func abc=def` -> `queue.enqueue(path.to.func, abc='def')`
+* `rq enqueue path.to.func ':{"json": "abc"}'` -> `queue.enqueue(path.to.func, {'json': 'abc'})`
+* `rq enqueue path.to.func 'key:={"json": "abc"}'` -> `queue.enqueue(path.to.func, key={'json': 'abc'})`
```
selwin, line 261: The syntax should be `rq enqueue path.to.func key:='{"json": "abc"}'`, similar to how httpie does it.

docs/docs/index.md
```diff
@@ -201,6 +201,75 @@ def report_failure(job, connection, type, value, traceback):
...
+* `rq enqueue path.to.func '%1, 2'` -> `queue.enqueue(path.to.func, (1, 2))`
```
selwin, line 262: Can we give another example using literal_eval by passing Python boolean and `None`? This should make it clearer.

docs/docs/index.md
```diff
@@ -201,6 +201,75 @@ def report_failure(job, connection, type, value, traceback):
...
+* `rq enqueue path.to.func 'key%=1, 2'` -> `queue.enqueue(path.to.func, key=(1, 2))`
+* `rq enqueue path.to.func @path/to/file` -> `queue.enqueue(path.to.func, open('path/to/file', 'r').read())`
+* `rq enqueue path.to.func key=@path/to/file` -> `queue.enqueue(path.to.func, key=open('path/to/file', 'r').read())`
+* `rq enqueue path.to.func :@path/to/file.json` -> `queue.enqueue(path.to.func, json.loads(open('path/to/file.json', 'r').read()))`
+* `rq enqueue path.to.func key:=@path/to/file.json` -> `queue.enqueue(path.to.func, key=json.loads(open('path/to/file.json', 'r').read()))`
```
selwin, line 267: Also, can we make sure each of these examples have a corresponding test case? Just want to be sure that everything works as documented.

docs/docs/index.md
rpkak: Both `key:='{"json": "abc"}'` and `'key:={"json": "abc"}'` are parsed to `key:={"json": "abc"}` by the shell. The `'` avoids that the shell escapes at the `"` and makes the shell parse this as one argument even though there is a space in this arg.

Try

```
echo key:='{"json": "abc"}'
```

and

```
echo 'key:={"json": "abc"}'
```

in the shell.

selwin: Yes, but `key:='{"json": "abc"}'` makes more syntactical sense (`key` being the keyword and what's after `:=` is what's parsed as JSON), so this is what I want to explain.

docs/docs/index.md
```diff
@@ -260,6 +260,8 @@ If the first character of `[value]` is `@` the subsequent path will be read.
 * `rq enqueue path.to.func ':{"json": "abc"}'` -> `queue.enqueue(path.to.func, {'json': 'abc'})`
 * `rq enqueue path.to.func 'key:={"json": "abc"}'` -> `queue.enqueue(path.to.func, key={'json': 'abc'})`
 * `rq enqueue path.to.func '%1, 2'` -> `queue.enqueue(path.to.func, (1, 2))`
+* `rq enqueue path.to.func '%None'` -> `queue.enqueue(path.to.func, None)`
+* `rq enqueue path.to.func '%True'` -> `queue.enqueue(path.to.func, True)`
 * `rq enqueue path.to.func 'key%=1, 2'` -> `queue.enqueue(path.to.func, key=(1, 2))`
```
selwin, line 265: Please change this to `key%='(1, 2)'` along with the tests.

docs/docs/index.md
```diff
@@ -260,6 +260,8 @@ If the first character of `[value]` is `@` the subsequent path will be read.
 * `rq enqueue path.to.func ':{"json": "abc"}'` -> `queue.enqueue(path.to.func, {'json': 'abc'})`
 * `rq enqueue path.to.func 'key:={"json": "abc"}'` -> `queue.enqueue(path.to.func, key={'json': 'abc'})`
 * `rq enqueue path.to.func '%1, 2'` -> `queue.enqueue(path.to.func, (1, 2))`
+* `rq enqueue path.to.func '%None'` -> `queue.enqueue(path.to.func, None)`
+* `rq enqueue path.to.func '%True'` -> `queue.enqueue(path.to.func, True)`
```
selwin, line 264: Could you please add another example, along with tests for passing in a Python dict?
```
rq enqueue path.to.func key%='{"foo": True}'` -> `queue.enqueue(path.to.func, key={"foo": True})
```

docs/docs/index.md
rpkak: Should I change this?

docs/docs/index.md
```diff
@@ -201,6 +201,78 @@ def report_failure(job, connection, type, value, traceback):
 Failure callbacks are limited to 60 seconds of execution time.
 
 
+### CLI Enqueueing
+
```
selwin, line 205: Can you add this `_New in version 1.10.0._` ?

selwin:
Aside from the small doc fix, I think this PR is good to go. Thanks for the effort, this is an incredible PR!

rq/cli/cli.py
```diff
@@ -303,3 +307,53 @@ def resume(cli_config, **options):
...
+@click.option('--retry-max', help='Maximum amound of retries', default=0, type=int)
+@click.option('--retry-interval', help='Interval between retries in seconds', multiple=True, type=int, default=[0])
+@click.option('--schedule-in', help='Delay until the function is enqueued (e.g. 10s, 5m, 2d).')
+@click.option('--schedule-at', help='Schedule job to be enqueued at a certain time formatted in ISO 8601 without '
+              'timezone (e.g. 2021-05-27T21:45:00).')
+@click.option('--quiet', is_flag=True, help='Only logs errors.')
+@click.argument('function')
+@click.argument('arguments', nargs=-1)
+@pass_cli_config
+def enqueue(cli_config, queue, timeout, result_ttl, ttl, failure_ttl, description, depends_on, job_id, at_front,
```
joshcoden, line 333: Should this also allow for passing in a custom serializer?

selwin: Yes, but I think this can be added in the next PR. I'll merge this PR in since this has been WIP for quite a while. Thanks @rpkak !

PR #1472, opened 2021-05-20
Add environment variables for connection

### LINKED ISSUE #1342: Default environment variables
Make worker to read default environment variables such as 
> REDIS_SERVER
> REDIS_PORT
> REDIS_USER
> REDIS_PASSWORD

...

when launching by 
> rq worker

rpkak:
Use same variable names as in config file.

fix #1342

### DIFF
--- a/rq/cli/helpers.py
+++ b/rq/cli/helpers.py
@@ -15,7 +15,7 @@ from redis.sentinel import Sentinel
 from rq.defaults import (DEFAULT_CONNECTION_CLASS, DEFAULT_JOB_CLASS,
                          DEFAULT_QUEUE_CLASS, DEFAULT_WORKER_CLASS)
 from rq.logutils import setup_loghandlers
-from rq.utils import import_attribute, string_types
+from rq.utils import import_attribute
 from rq.worker import WorkerStatus
 
 red = partial(click.style, fg='red')
--- a/rq/cli/helpers.py
+++ b/rq/cli/helpers.py
@@ -50,14 +50,13 @@ def get_redis_from_config(settings, connection_class=Redis):
         return sn.master_for(master_name)
 
     ssl = settings.get('REDIS_SSL', False)
-    if isinstance(ssl, string_types):
+    if isinstance(ssl, str):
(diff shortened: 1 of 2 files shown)


rq/cli/helpers.py
```diff
@@ -48,12 +49,22 @@ def get_redis_from_config(settings, connection_class=Redis):
         sn = Sentinel(instances, socket_timeout=socket_timeout, password=password, db=db)
         return sn.master_for(master_name)
 
+    ssl = settings.get('REDIS_SSL', False)
+    if isinstance(ssl, string_types):
```
selwin, line 53: Can we use `str` instead of `string_types`? We no longer need to support Python 2.

rq/cli/helpers.py
```diff
@@ -48,12 +49,22 @@ def get_redis_from_config(settings, connection_class=Redis):
...
+        if ssl.lower() in ['y', 'yes', 't', 'true']:
+            ssl = True
+        elif ssl.lower() in ['n', 'no', 'f', 'false', '']:
+            ssl = False
+        else:
+            raise ValueError('REDIS_SSL is a boolean and must be \'y\', \'yes\', \'t\', \'true\', \'n\', \'no\', '
```
selwin, line 59: I think the error message could be a bit simpler. We can just say `REDIS_SSL is a boolean and must be "True" or "False"`

selwin:
While we're in the topic of reading environment variables, do you mind adding an optional `RQ_` prefix when reading from environment variables? Currently `REDIS_HOST` may already be used by other libraries too and sometimes this could create conflicts.

rq/cli/helpers.py
```diff
@@ -14,7 +15,7 @@
 from rq.defaults import (DEFAULT_CONNECTION_CLASS, DEFAULT_JOB_CLASS,
                          DEFAULT_QUEUE_CLASS, DEFAULT_WORKER_CLASS)
 from rq.logutils import setup_loghandlers
-from rq.utils import import_attribute
+from rq.utils import import_attribute, string_types
```
selwin, line 18: Can you remove `string_types` from this line?

PR #1484, opened 2021-06-02
Add job.fail

### LINKED ISSUE #1313: Is it possible to fail a job other than by raising an exception?
I heavily use the new _Retrying Failed Jobs_ feature. (Thank you again 🎉)

At them moment I raise an exception, whenever I want a job to fail.
Is it possible to fail a job other than by raising an exception?

This would be interesting because it would sometimes better communicate the intention of the code and also to keep the logs clean.

rpkak:
Added method `job.fail` to fail the job using a method.

`job.fail` takes some args to change the failture behavior. Not specifying anything affects the same as raising an exception, apart from printing the traceback which is disabled by default in `job.fail`.

**For review**: Please tell me if I should change `exc_info` to `orig_exc_info` in some places in 
https://github.com/rq/rq/blob/2ee3416aa1a1653ff5bc3974ed5566ea60c49d93/rq/worker.py#L1045-L1051

Closes #1313 
View: https://github.com/rq/rq/issues/1313#issuecomment-847300810

### DIFF
--- a/rq/exceptions.py
+++ b/rq/exceptions.py
@@ -31,3 +31,14 @@ class ShutDownImminentException(Exception):
 
 class TimeoutFormatError(Exception):
     pass
+
+
+class JobFailture(Exception):
+    def __init__(self, msg, retry=None, seconds_until_next_retry=None, decrease_retries=True, show_traceback=True,
+                 use_exc_handlers=True):
+        super(JobFailture, self).__init__(msg)
+        self.retry = retry
+        self.seconds_until_next_retry = seconds_until_next_retry
+        self.decrease_retries = decrease_retries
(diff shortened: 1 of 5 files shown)


rpkak:
I don't think I can add some more tests increase the coverage.

selwin:
@rpkak please see my [comment here](https://github.com/rq/rq/issues/1313#issuecomment-860000008). Do you think this PR is still relevant?

rpkak:
> @rpkak please see my [comment here](https://github.com/rq/rq/issues/1313#issuecomment-860000008). Do you think this PR is still relevant?

Yes, I think in special cases of a job failture the customizability of the function will usefull.

PR #1485, opened 2021-06-02
Make `Queue.enqueue`, `Queue.enqueue_call`,  `Queue.enqueue_at``Queue.parse_args` accept `pipeline` arg, add `Queue.enqueue_many` method

### LINKED ISSUE #241: Feature request: pipelined enqueues and fetches

### LINKED ISSUE #753: Pipeline enqueue() calls?
Hi,

I do large numbers of my_queue.enqueue() calls at a time, is there a way to do this with a pipeline? I looked at the source but couldn't see anything obvious. I often enqueue thousands of jobs at once, so pipelining seems worthwhile.

Thanks

joshcoden:
Resolves #753 

See also: #241

Allow user to pass in a `pipeline` argument to the enqueue methods.

One main use case is the ability for the caller to use their own pipeline and still be able to utilize the dependency logic defined in `enqueue_call`.

-cc @selwin

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -24,7 +24,7 @@ def compact(lst):
     return [item for item in lst if item is not None]
 
 
-class JobEnqueueData(namedtuple('JobData', ["func", "args", "kwargs", "timeout",
+class JobEnqueueData(namedtuple('JobEnqueueData', ["func", "args", "kwargs", "timeout",
                                             "result_ttl", "ttl", "failure_ttl",
                                             "description", "job_id",
                                             "at_front", "meta", "retry"])):
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -313,7 +313,7 @@ class Queue:
     def enqueue_call(self, func, args=None, kwargs=None, timeout=None,
                      result_ttl=None, ttl=None, failure_ttl=None,
                      description=None, depends_on=None, job_id=None,
-                     at_front=False, meta=None, retry=None):
+                     at_front=False, meta=None, retry=None, pipeline=None):
         """Creates a job to represent the delayed function call and enqueues
         it.
 nd
@@ -337,33 +337,44 @@ nd
         # something else has modified either the set of dependencies or the
         # status of one of them. In this case, we simply retry.
         if depends_on is not None:
-            with self.connection.pipeline() as pipe:
-                while True:
-                    try:
+            pipe = pipeline if pipeline is not None else self.connection.pipeline()
+            while True:
+                try:
+                    # Also calling watch even if caller 
+                    # passed in a pipeline since Queue#create_job
+                    # is called from within this method.
+                    pipe.watch(job.dependencies_key)
 
-                        pipe.watch(job.dependencies_key)
+                    dependencies = job.fetch_dependencies(
+                        watch=True,
+                        pipeline=pipe
+                    )
 
-                        dependencies = job.fetch_dependencies(
-                            watch=True,
-                            pipeline=pipe
-                        )
+                    pipe.multi()
 
-                        pipe.multi()
-
-                        for dependency in dependencies:
-                            if dependency.get_status(refresh=False) != JobStatus.FINISHED:
-                                job.set_status(JobStatus.DEFERRED, pipeline=pipe)
-                                job.register_dependency(pipeline=pipe)
-                                job.save(pipeline=pipe)
-                                job.cleanup(ttl=job.ttl, pipeline=pipe)
+                    for dependency in dependencies:
+                        if dependency.get_status(refresh=False) != JobStatus.FINISHED:
+                            job.set_status(JobStatus.DEFERRED, pipeline=pipe)
+                            job.register_dependency(pipeline=pipe)
+                            job.save(pipeline=pipe)
+                            job.cleanup(ttl=job.ttl, pipeline=pipe)
+                            if pipeline is None:
                                 pipe.execute()
-                                return job
-
-                        break
-                    except WatchError:
+                            return job
+                    job = self.enqueue_job(job, pipeline=pipe, at_front=at_front)
+                    if pipeline is None:
+                        pipe.execute()
+                    return job
+                except WatchError:
+                    if pipeline is None:
                         continue
-
-        job = self.enqueue_job(job, at_front=at_front)
+                    else:
+                        # if pipeline comes from caller, re-raise to them
+                        raise
+        # No need for multi or While True loop because in this case
+        # Anything being watched came from caller, so 
+        # it will simply raise back up to them.    
+        job = self.enqueue_job(job, pipeline=pipeline, at_front=at_front)
         return job
 
     def run_job(self, job):
@@ -401,6 +412,7 @@ nd
         at_front = kwargs.pop('at_front', False)
         meta = kwargs.pop('meta', None)
         retry = kwargs.pop('retry', None)
+        pipeline = kwargs.pop('pipeline', None)
 
         if 'args' in kwargs or 'kwargs' in kwargs:
             assert args == (), 'Extra positional arguments cannot be used when using explicit args and kwargs'  # noqa
@@ -408,32 +420,32 @@ nd
             kwargs = kwargs.pop('kwargs', None)
 
         return (f, timeout, description, result_ttl, ttl, failure_ttl,
-                depends_on, job_id, at_front, meta, retry, args, kwargs)
+                depends_on, job_id, at_front, meta, retry, pipeline, args, kwargs)
 
     def enqueue(self, f, *args, **kwargs):
         """Creates a job to represent the delayed function call and enqueues it."""
 
         (f, timeout, description, result_ttl, ttl, failure_ttl,
-         depends_on, job_id, at_front, meta, retry, args, kwargs) = Queue.parse_args(f, *args, **kwargs)
+         depends_on, job_id, at_front, meta, retry, pipeline, args, kwargs) = Queue.parse_args(f, *args, **kwargs)
 
         return self.enqueue_call(
             func=f, args=args, kwargs=kwargs, timeout=timeout,
             result_ttl=result_ttl, ttl=ttl, failure_ttl=failure_ttl,
             description=description, depends_on=depends_on, job_id=job_id,
-            at_front=at_front, meta=meta, retry=retry
+            at_front=at_front, meta=meta, retry=retry, pipeline=pipeline
         )
 
     def enqueue_at(self, datetime, f, *args, **kwargs):
         """Schedules a job to be enqueued at specified time"""
 
         (f, timeout, description, result_ttl, ttl, failure_ttl,
-         depends_on, job_id, at_front, meta, retry, args, kwargs) = Queue.parse_args(f, *args, **kwargs)
+         depends_on, job_id, at_front, meta, retry, pipeline, args, kwargs) = Queue.parse_args(f, *args, **kwargs)
         job = self.create_job(f, status=JobStatus.SCHEDULED, args=args, kwargs=kwargs,
                               timeout=timeout, result_ttl=result_ttl, ttl=ttl,
                               failure_ttl=failure_ttl, description=description,
                               depends_on=depends_on, job_id=job_id, meta=meta, retry=retry)
 
-        return self.schedule_job(job, datetime)
+        return self.schedule_job(job, datetime, pipeline=pipeline)
 
     def schedule_job(self, job, datetime, pipeline=None):
         """Puts job on ScheduledJobRegistry"""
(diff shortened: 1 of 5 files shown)


selwin:
@joshcoden may I trouble you to implement `queue.enqueue_many()` that takes advantage of this `pipeline` argument? :)

My thinking is that you'd use this as well so might as well build it into the library for everyone to use.

joshcoden:
> @joshcoden may I trouble you to implement `queue.enqueue_many()` that takes advantage of this `pipeline` argument? :)
> 
> My thinking is that you'd use this as well so might as well build it into the library for everyone to use.

I don't mind, but since it wasn't part of my original PR and desire to write tis PR (so thus I don't have an opinionated idea of what the signature and behavior should be), I'm wondering how you would want it implemented.  Some things that come to mind include:

1. Do we need the order of jobs in the input to matter?
2. What does the signature look like?
3.  How about dependency behavior?  Can the jobs in the input depend on each other?  Should I be refactoring it out of `enqueue_call` to be able to be used in multiple places?

selwin:
> 1. Do we need the order of jobs in the input to matter?

Since the API would likely involve accepting an iterable of some sort, I think we might as well enqueue jobs based on the order at which they're passed in.

> 2. What does the signature look like?

I'm open to suggestions, but one thing that came to mind is something like this:
```python
JobData = namedtuple('JobData', ['func', 'args', 'kwargs', 'result_ttl', ...])

queue.enqueue_many(
    JobData(f=foo, kwargs={'bar': 'baz'}, result_ttl=60),
    JobData(f=foo, kwargs={'bar': 'baz'}, result_ttl=60),
)
```

`JobData` would support all arguments accepted by [queue.enqueue()](https://github.com/rq/rq/blob/7a1136644732b6b595019d1aea8e54d769664655/rq/queue.py#L313), except for `depends_on`.

This is not final and I'm happy to explore other suggestions.

3. Yes, dealing with job dependencies will be tricky. We may need to create all jobs without enqueueing them first before setting up the dependencies in a single transaction while we monitor for key changes. I think we can leave out job dependency support for now. I'm happy to accept future enhancements to this if someone is willing to work on it.

joshcoden:
> > 1. Do we need the order of jobs in the input to matter?
> 
> Since the API would likely involve accepting an iterable of some sort, I think we might as well enqueue jobs based on the order at which they're passed in.
>
Makes sense.
> > 2. What does the signature look like?
> 
> I'm open to suggestions, but one thing that came to mind is something like this:
> 
> ```python
> JobData = namedtuple('JobData', ['func', 'args', 'kwargs', 'result_ttl', ...])
> 
> queue.enqueue_many(
>     JobData(f=foo, kwargs={'bar': 'baz'}, result_ttl=60),
>     JobData(f=foo, kwargs={'bar': 'baz'}, result_ttl=60),
> )
> ```
> 
> `JobData` would support all arguments accepted by [queue.enqueue()](https://github.com/rq/rq/blob/7a1136644732b6b595019d1aea8e54d769664655/rq/queue.py#L313), except for `depends_on`.
> 

Got it.  I actually think you do have to make a separate type if you want the closest thing to "passing multiple `args` and `kwargs`".  

> 
> 3. Yes, dealing with job dependencies will be tricky. We may need to create all jobs without enqueueing them first before setting up the dependencies in a single transaction while we monitor for key changes. I think we can leave out job dependency support for now. I'm happy to accept future enhancements to this if someone is willing to work on it.

Yeah, I don't think it will be particularly hard, but would require some significant refactoring.  I agree though, supporting dependencies for now would be scope creep.  I would probably be willing to work on it, but it's not something I'm actually looking to do pipelined enqueueing at the moment and would rather prioritize something more basic for now.

joshcoden:
@selwin Just added my initial implementation and unit tests for `enqueue_many`.  I also refactored the dependency logic and enqueueing out of `enqueue_call` and into its own method, since it would be useful to be able to use that logic elsewhere.  I can revert it if you would rather.  I left this:

https://github.com/rq/rq/blob/6343686d2579bf1d251cb91d08560e2552cef30b/rq/queue.py#L351-L353

Because I didn't think it made sense to pass `depends_on` to the refactored method since it takes a job instance.  But depends_on is not directly stored on the instance, and I wanted to perform the same check as using `depends_on` without having to actual hit the redis connection, and this seemed like the best way available.  Let me know if you have any thoughts on this issue.

rq/queue.py
```diff
@@ -310,10 +335,66 @@ def create_job(self, func, args=None, kwargs=None, timeout=None,
...
+        at_front=False
+    ):
+        # If a _dependent_ job depends on any unfinished job, register all the
+        # _dependent_ job's dependencies instead of enqueueing it.
+        #
+        # `Job#fetch_dependencies` sets WATCH on all dependencies. If
+        # WatchError is raised in the when the pipeline is executed, that means
+        # something else has modified either the set of dependencies or the
+        # status of one of them. In this case, we simply retry.
+        if job._dependency_id is not None:
```
selwin, line 351: Can you check with `job._dependency_ids` instead? `job._dependency_id` will be deprecated soon. I'm ok with this way of checking whether this job has dependencies.

rq/queue.py
```diff
@@ -329,42 +410,46 @@ def enqueue_call(self, func, args=None, kwargs=None, timeout=None,
...
+            pipeline=pipeline,
+            at_front=at_front
+        )
 
-        job = self.enqueue_job(job, at_front=at_front)
-        return job
+    def enqueue_many(
+        self,
+        job_datas,
+        pipeline=None
```
selwin, line 422: Do we need to accept `pipeline` argument here?

rq/queue.py
```diff
@@ -329,42 +410,46 @@ def enqueue_call(self, func, args=None, kwargs=None, timeout=None,
...
+                        ),
+                        pipeline=pipe,
+                        at_front=job_data.at_front
+                    )
+                    for job_data in job_datas
+                ]
+                if pipeline is None:
+                    pipe.execute()
+                return jobs
+            except WatchError:
```
selwin, line 447: Do we need to catch `WatchError` here if `depends_on` is always `None`?

rq/queue.py
```diff
@@ -23,6 +24,30 @@ def compact(lst):
     return [item for item in lst if item is not None]
 
 
+class JobEnqueueData(namedtuple('JobEnqueueData', ["func", "args", "kwargs", "timeout",
```
selwin, line 27: Is there a reason why we're naming this `JobEnqueueData` as opposed to `JobData`? Also open for other ideas.

selwin: Thinking further about this, you're right that `JobEnqueueData` makes more sense since this is not just simply data about job itself. Just wondering if there's a better name for this. Some ideas:
```python
queue.enqueue_many(
    queue.EnqueueData(func=foo, result_ttl=60),
    queue.EnqueueData(func=bar)
)
```

queue.enqueue_many(
    queue.prepare_data(func=foo, result_ttl=60),
    queue.prepare_data(func=bar)
)
```

File: tests/test_queue.py [hunk 5]
```diff
@@ -507,6 +510,114 @@ def test_enqueue_job_with_dependency(self):
...
+                try:
+                    job = q.enqueue_call(say_hello, depends_on=parent_job, pipeline=pipe)
+                    self.assertEqual(q.job_ids, [])
+                    self.assertEqual(job.get_status(refresh=False), JobStatus.DEFERRED)
+                    # Not in registry before execute, since passed in pipeline
+                    self.assertEqual(len(q.deferred_job_registry), 0)
+                    pipe.execute()
+                    break
+                except WatchError:
+                    continue
```
selwin, line 530: We don't need this in the test case.

tests/test_queue.py
```diff
@@ -507,6 +510,114 @@ def test_enqueue_job_with_dependency(self):
...
+                try:
+                    job = q.enqueue_call(say_hello, depends_on=parent_job, pipeline=pipe)
+                    # Pre execute conditions
+                    self.assertEqual(q.job_ids, [])
+                    self.assertEqual(job.timeout, Queue.DEFAULT_TIMEOUT)
+                    self.assertEqual(job.get_status(refresh=False), JobStatus.QUEUED)
+                    pipe.execute()
+                    break
+                except WatchError:
+                    continue
```
selwin, line 548: We don't need this in the test case I think.

selwin:
Also, we'll need to document this feature so it's discoverable. The docs are in the `docs` directory.

rq/queue.py
joshcoden: @selwin I did for completeness sake, especially with the original motivation of this PR was adding the ability to pipeline the various enqueue methods, so it was easy enough to implement this.

rq/queue.py
joshcoden: I guess not, if `pipeline` is passed in without the `except` it would be raised to them, which is what we are doing here anyway.  I can drop this code.

rq/queue.py
joshcoden: I'm fine with either.  You could argue the latter one is better because it does not expose the inner type.

selwin: Ok, then let's go with 
```python
queue.enqueue_many(
    queue.prepare_data(func=foo, result_ttl=60),
    queue.prepare_data(func=bar)
)
```

Also open to suggestion on a better name for `prepare_data`.

joshcoden: @selwin I don't have any strong feelings on the name, `prepare_data` honestly does not seem bad to me.  I'll implement it that way, at least for now.

joshcoden:
> Also, we'll need to document this feature so it's discoverable. The docs are in the `docs` directory.

@selwin I addressed your comments and added some docs, let me know if it's in the right place and if there is anything missing or needing changes.  I assumed it would be in version 1.9.0 when specifying when this new feature was available in the docs.

rq/queue.py
```diff
@@ -310,10 +322,64 @@ def create_job(self, func, args=None, kwargs=None, timeout=None,
 
         return job
 
+    def setup_dependencies_and_enqueue(
```
selwin, line 325: One more small change, can you change this to `setup_dependencies` and leave out the `enqueue_job` line? It's cleaner to have a single method to setup dependencies and we can potentially reuse this in other places.

selwin: Unless the `enqueue_job` is part of the logic and it's hard to untangle, in which case we can leave it as it is.

joshcoden: @selwin I thought about doing that, but that means we could have the case if `depends_on` is True and a job is found we would have that job that may already be in the registry of the queue, or the job got no where at all

joshcoden: @selwin Actually, on second thought, there should be a way to refactor it out.  Maybe we should have a discussion on the best way how?   Because if aren't enqueueing we would need to know if the job was deferred or if none were found and should just be enqueued maybe via a bool return value?

selwin: What if we check the job status? If `job.status` is not `DEFERRED`, then it should be enqueued.

joshcoden: @selwin Done.  Should I be increasing the version number to `1.9.0`?:
https://github.com/rq/rq/blob/5620f2932cffb31df58a164d78e5cd4327b89414/rq/version.py#L5

selwin: Yes, would be great if you could do that. After this is merged in, I'll also pull in https://github.com/rq/rq/pull/1480 and cut a release in a few weeks.

joshcoden: @selwin Awesome, thank you so much!  I have updated the version to 1.9.0.

selwin:
Thanks!

PR #1493, opened 2021-06-14
Docs: Add scheduled to job.get_status()

### LINKED ISSUE #1492: get_status can return "scheduled"
There is an issue in the docs: https://python-rq.org/docs/jobs/ says

> job.get_status() Possible values are queued, started, deferred, finished, stopped, and failed

https://github.com/rq/rq/blob/819d0498c715d693e8a295ec6202c3fe0262e4fb/rq/job.py#L33-L40

rpkak:
Fixes #1492


selwin:
Thanks!

Issue #1488, opened 2021-06-06
Invalid attribute name: fib.slow_fib


54696d21:
```
  File "/home/pi/.local/lib/python3.7/site-packages/rq/worker.py", line 1013, in perform_job
    rv = job.perform()
  File "/home/pi/.local/lib/python3.7/site-packages/rq/job.py", line 709, in perform
    self._result = self._execute()
  File "/home/pi/.local/lib/python3.7/site-packages/rq/job.py", line 732, in _execute
    result = self.func(*self.args, **self.kwargs)
  File "/home/pi/.local/lib/python3.7/site-packages/rq/job.py", line 221, in func
    return import_attribute(self.func_name)
  File "/home/pi/.local/lib/python3.7/site-packages/rq/utils.py", line 144, in import_attribute
    raise ValueError('Invalid attribute name: %s' % name)
ValueError: Invalid attribute name: fib.slow_fib
```

I tried several configurations but I get always the Invalid attribute name error. I tried the example in the readme and the example in the readme. I'm on version 1.8.1

I'm using the command `rq worker --with-scheduler` as worker

rpkak:
Are you executing this command from a place that can import `fib` (e.g. the project folder or a copy of it).

54696d21:
No, I forgot to do that. That fixes it, thank you :)

natbusa:
I would suggest to add some feedback in the form of import trace logging on the `import_attribute` function. 
See pull request #1494

PR #1494, opened 2021-06-15
Added logging trace for failed import for user-level debugging

natbusa:
Sometimes the failure during importing is due to a genuine coding error in the imported module. In this case, the exception is logged as a warning for ease of debugging of the imported user code.  The logic is unchanged and it will be applied anyways regardless of the cause of the import error.

A case where it seems beneficial to have more feedback on the import error has been reported in issue #1488

### DIFF
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -142,8 +142,7 @@ def import_attribute(name):
             module = importlib.import_module(module_name)
             break
         except ImportError:
-            logging.warning("Import error for '%s'" % module_name)
-            logging.warning("Import error trace:", exc_info=True)
+            logging.warning("Import error for '%s'" % module_name, exc_info=True)
             attribute_bits.insert(0, module_name_bits.pop())
 
     if module is None:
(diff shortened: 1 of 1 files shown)


rq/utils.py
```diff
@@ -132,12 +132,18 @@ def import_attribute(name):
...
+    # In this case, the exception is logged as a warning for ease of debugging.
+    # The above logic will apply anyways regardless of the cause of the import error.
     while len(module_name_bits):
         try:
             module_name = '.'.join(module_name_bits)
             module = importlib.import_module(module_name)
             break
         except ImportError:
+            logging.warning("Import error for '%s'" % module_name)
+            logging.warning("Import error trace:", exc_info=True)
```
selwin, line 146: I'm ok with logging this. Just wondering whether this should be logged as warning or error.

natbusa: Many thanks, I chose warning as it's closer to the original logic (no warning at all). But error is surely an option.

natbusa:
small fix: merged to two logging statements into one

selwin:
Thanks!

PR #1495, opened 2021-06-18
add REDIS_SSL_CERT_REQS cli parameter

mgcdanny:
Allow  ssl with password only, without certs by setting  REDIS_SSL=True and REDIS_SSL_CERT_REQS=None

### DIFF
--- a/rq/cli/helpers.py
+++ b/rq/cli/helpers.py
@@ -65,6 +65,7 @@ def get_redis_from_config(settings, connection_class=Redis):
         'password': settings.get('REDIS_PASSWORD', None),
         'ssl': ssl,
         'ssl_ca_certs': settings.get('REDIS_SSL_CA_CERTS', None),
+        'ssl_cert_reqs': settings.get('REDIS_SSL_CERT_REQS', 'required'),
     }
 
     return connection_class(**kwargs)


selwin:
Do you mind documenting this feature? Docs are under `/docs`.

mgcdanny:
@selwin not sure where under /docs would be a good place but I updated the instructions on the Heroku section since that is how I am using it.

edmorley:
@selwin Hi! Do you know roughly when a new release containing this change might be published? :-)

mgcdanny:
Thank you all for the wonderful library.

jelmerdejong:
@mgcdanny with this change now released, did you get Redis on Heroku properly working? I'm still struggling.

richgong:
This doesn't quite work yet. If you pass in "None", this is a string. The logic in Redis connection is:
```
if ssl_cert_reqs is None:
  ssl_cert_reqs = ssl.CERT_NONE
```
So rq needs to convert string "None" to actual None.

richgong:
The other issue is if you configure with `--url` or via `REDIS_URL`, it skips the `REDIS_SSL_CERT_REQS` check entirely and constructs using `Redis.from_url`

PR #1498, opened 2021-06-21
Ensure pipeline in multi mode after dep setup

### LINKED ISSUE #1497: Call pipeline.multi() even if there are no dependencies to ensure pipeline always in buffered mode after call
https://github.com/rq/rq/blob/456743b225c0b5a4339242a5b714c378b86f2fe0/rq/queue.py#L368-L369

We need an else block here to make sure that we always leave the pipeline in the same mode for the caller.

@selwin

I can work on this fix before 1.9.0 is released.

ie.

```python

elif pipeline is not None:
    pipeline.multi()
```

joshcoden:
Resolves #1497

The added unit test shows an example that would have failed prior to this change.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -366,6 +366,8 @@ class Queue:
                     else:
                         # if pipeline comes from caller, re-raise to them
                         raise
+        elif pipeline is not None:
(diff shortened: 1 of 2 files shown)


selwin:
Thanks!

PR #1499, opened 2021-06-21
Extract `Job.get_call_string` logic to `utils.get_call_string`

Necior:
# Motivation / use case

Let's say you want to enqueue invocation of the following function: `f(arg1, arg2, secret_passphrase)`.
Currently, to avoid logging `secret_passphrase`, possible solutions include disabling logging or providing custom `Job.description`.

If you want to mimic rq's default `Job.description`, you have to create a fake `Job` with `func_name`, `args` and `kwargs` attributes. Then you can call `Job.get_call_string(fake_object)`. Such workaround has a potential to break application when the implementation details of `Job` change.

# Proposed solution

Move `Job.get_call_string` logic to `utils` with the following signature `get_call_string(func_name, args, kwargs)`. This way that helper function can be used by applications, for example in the following way:

```python
func_name, args, kwargs = ..., ..., ...
q = get_queue()

kwargs_without_secrets = kwargs.copy()
kwargs_without_secrets['secret_passphrase'] = "**REDACTED**"

my_desc = get_call_string(func_name, args, kwargs_without_secrets)
q.enqueue(f, *args, description=my_desc, **kwargs)
```

# Additional considerations

`job.truncate_long_string` is only used by `get_call_string`. I'm not sure whether it's part of public API so I didn't move it to `utils`.

If you'd like to have it in `utils` instead of keeping it in `job`, I'm ready to help.

### DIFF
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -299,6 +299,13 @@ def split_list(a_list, segment_size):
         yield a_list[i:i + segment_size]
 
 
+def truncate_long_string(data, max_length=None):
+    """Truncate arguments with representation longer than max_length"""
+    if max_length is None:
+        return data
+    return (data[:max_length] + '...') if len(data) > max_length else data
+
+
 def get_call_string(func_name, args, kwargs, max_length=None):
     """Returns a string representation of the call, formatted as a regular
     Python function invocation statement. If max_length is not None, truncate
@@ -307,14 +314,9 @@ def get_call_string(func_name, args, kwargs, max_length=None):
     if func_name is None:
         return None
 
-    def truncate_long_string(data):
-        if max_length is None:
-            return data
-        return (data[:max_length] + '...') if len(data) > max_length else data
+    arg_list = [as_text(truncate_long_string(repr(arg), max_length)) for arg in args]
 
-    arg_list = [as_text(truncate_long_string(repr(arg))) for arg in args]
-
-    kwargs = ['{0}={1}'.format(k, as_text(truncate_long_string(repr(v)))) for k, v in kwargs.items()]
+    kwargs = ['{0}={1}'.format(k, as_text(truncate_long_string(repr(v), max_length))) for k, v in kwargs.items()]
     arg_list += sorted(kwargs)
     args = ', '.join(arg_list)
 
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -310,7 +310,6 @@ def get_call_string(func_name, args, kwargs):
     arg_list = [as_text(truncate_long_string(repr(arg))) for arg in args]
 
     kwargs = ['{0}={1}'.format(k, as_text(truncate_long_string(repr(v)))) for k, v in kwargs.items()]
-    # Sort here because python 3.3 & 3.4 makes different call_string
     arg_list += sorted(kwargs)
     args = ', '.join(arg_list)
 
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -297,3 +297,21 @@ def split_list(a_list, segment_size):
     """
     for i in range(0, len(a_list), segment_size):
         yield a_list[i:i + segment_size]
+
+
+def get_call_string(func_name, args, kwargs):
+    """Returns a string representation of the call, formatted as a regular
+    Python function invocation statement.
+    """
+    if func_name is None:
+        return None
+    from .job import truncate_long_string  # work around circular import issue
+
+    arg_list = [as_text(truncate_long_string(repr(arg))) for arg in args]
+
+    kwargs = ['{0}={1}'.format(k, as_text(truncate_long_string(repr(v)))) for k, v in kwargs.items()]
+    # Sort here because python 3.3 & 3.4 makes different call_string
+    arg_list += sorted(kwargs)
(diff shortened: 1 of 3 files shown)


rpkak:
Already included in #1466.

rq/utils.py
```diff
@@ -297,3 +297,21 @@ def split_list(a_list, segment_size):
...
+    """
+    if func_name is None:
+        return None
+    from .job import truncate_long_string  # work around circular import issue
+
+    arg_list = [as_text(truncate_long_string(repr(arg))) for arg in args]
+
+    kwargs = ['{0}={1}'.format(k, as_text(truncate_long_string(repr(v)))) for k, v in kwargs.items()]
+    # Sort here because python 3.3 & 3.4 makes different call_string
+    arg_list += sorted(kwargs)
```
selwin, line 314: We can drop the sort here since we don't need to worry about Python 3.3 or 3.4 anymore.

selwin:
> job.truncate_long_string is only used by get_call_string. I'm not sure whether it's part of public API so I didn't move it to utils. If you'd like to have it in utils instead of keeping it in job, I'm ready to help.

Yes I think it'd make sense to move this to utils too, preferable as part of `get_call_string()`. Maybe something like `get_call_string(..., truncate=False)`?

I'm also open to ideas on exposing a public API to prevent sensitive information from being logged.

rq/utils.py
Necior: Good catch. The comment was outdated so I removed it.

Anyway, I would keep the sort but for a different reason.

Since rq supports Python ≥ 3.5 and dict ordering is guaranteed starting with Python 3.7, I'd sort to keep function output deterministic. Otherwise, for the same arguments, the result might differ in different Python implementations/versions.

What do you think?

Necior:
> > job.truncate_long_string is only used by get_call_string. I'm not sure whether it's part of public API so I didn't move it to utils. If you'd like to have it in utils instead of keeping it in job, I'm ready to help.
> 
> Yes I think it'd make sense to move this to utils too, preferable as part of `get_call_string()`. Maybe something like `get_call_string(..., truncate=False)`?

I think `maxlen` should remain configurable so I propose `get_call_string(…, truncate=False, maxlen=75)`. To maintain backwards compatibility, pass `truncate=True` in `job.Job.get_call_string`.

I have no preference whether we should default to truncating or not in the utility function so I adhered to your suggestion.

Please take a look at b1720d4207cb5973efd04656422194df8b26b631.

> I'm also open to ideas on exposing a public API to prevent sensitive information from being logged.

I think it's highly application specific. rq is not aware of encoding, used data structures etc. so I think it could be difficult to provide a generic solution. Hence my proposition to make `get_call_string` usable (easily) by applications so they can manage job description (which is what is being logged anyway) themselves.

That said, I see a possibility to mark specific key-values pairs in kwargs and sensitive and avoid logging them on the rq side.

rq/utils.py
```diff
@@ -297,3 +297,25 @@ def split_list(a_list, segment_size):
     """
     for i in range(0, len(a_list), segment_size):
         yield a_list[i:i + segment_size]
+
+
+def get_call_string(func_name, args, kwargs, truncate=False, maxlen=75):
```
selwin, line 302: `maxlen` -> `max_length`.

If we already have `max_length`, we can get rid of `truncate` argument. We can just default `max_length` to `None`.

rq/utils.py
selwin: Ok.

tests/test_utils.py
```diff
@@ -113,3 +113,8 @@ def test_split_list(self):
 
         expected_small_list_count = ceildiv(BIG_LIST_SIZE, SEGMENT_SIZE)
         self.assertEqual(len(small_lists), expected_small_list_count)
+
+    def test_get_call_string(self):
+        """Ensure a case when all arguments are not None works properly"""
+        cs = get_call_string("f", ('some', 'args', 42), {"key1": "value1", "key2": True})
```
selwin, line 119: We should also test the `max_length` argument.

rq/utils.py
Necior: :+1:

rq/utils.py
```diff
@@ -297,3 +297,25 @@ def split_list(a_list, segment_size):
...
+
+def get_call_string(func_name, args, kwargs, max_length=None):
+    """Returns a string representation of the call, formatted as a regular
+    Python function invocation statement. If max_length is not None, truncate
+    arguments with representation longer than max_length.
+    """
+    if func_name is None:
+        return None
+
+    def truncate_long_string(data):
```
selwin, line 310: One more thing, can you please move this to this module's top level so we can re-use this? The function signature should also change to `truncate_long_string(data, max_length=None):`

Necior: Sure, no problem.

I took the liberty of creating an additional test since with that change `truncate_long_string` becomes a "public" function. I hope you're okay with that.

If there's anything else you think I could change in this pull request, I'm ready to help.

selwin:
Thanks!

Necior:
Cool! Thanks for the code review!

Issue #1416, opened 2021-02-18
Synchronous Job Results are getting cleared after default time even after specifying result_ttl = -1


AkashKabra11:
Hi, 
I am trying to make synchronous jobs using rq. I am instantiating the queues like this: 

```
from rq import Queue
my_queue = Queue('sync', is_async=False, connection=Redis())
```

Now, when I create a job like this: 
```
job = my_queue.enqueue(func_name, args=(my_args,), result_ttl=-1)
```
**The Synchronous job results are not persisted after 500 seconds**. On the other hand, job results are persisted for asynchronous jobs. On a quick debug, I found that **the issue lies in this line**: 
https://github.com/rq/rq/blob/fc7940c77b69704ad6c059385fd530c850e5f0be/rq/queue.py#L373

`enqueue` function internally calls `enqueue_call` which internally calls `enqueue_job`. This `enqueue_job` function calls the `run_job` function whenever is_async is false. run_job function looks like this: 
```
    def run_job(self, job):
        job.perform()
        job.set_status(JobStatus.FINISHED)
        job.save(include_meta=False)
        job.cleanup(DEFAULT_RESULT_TTL).  <--------- ISSUE
        return 
```
It performs cleanup after DEFAULT_RESULT_TTL irrespective of what is the result_ttl passed. It should perform cleanup after `job.result_ttl interval`. This small issue is disrupting the entire functionality of Synchronous jobs. I request the contributors to please make this change ASAP.

selwin:
Please open a PR for this :)

lowercase00:
Fixed on #1510

PR #1510, opened 2021-07-08
Issue 1416: Use result_ttl for synchronous queues 

bbayles:
Re: https://github.com/rq/rq/issues/1416, this PR updates queues with `is_async=False` to respect job result TTLs.

The issue diagnoses the problem correctly. The new tests cover the `Queue.run_job` method, which is making the `cleanup` call.


selwin:
Thanks!

Issue #1508, opened 2021-07-05
Google Memory Store Redis doesn't allow CLIENT commands


sbv-csis:
CLIENT command is disabled for Google memory store: https://cloud.google.com/memorystore/docs/redis/product-constraints

Which means version 1.9.0 doesn't work on google memory store (Due to [PR 1463](https://github.com/rq/rq/pull/1463/files#diff-50984f276e8a6fea5f4ba4938afb6f4fbc544fd3751fbd16027dc0a326d6405eR224))
```
Traceback (most recent call last):
  File "/venv/bin/rq", line 8, in <module>
    sys.exit(main())
  File "/venv/lib/python3.9/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/venv/lib/python3.9/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/venv/lib/python3.9/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/venv/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/venv/lib/python3.9/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/venv/lib/python3.9/site-packages/rq/cli/cli.py", line 83, in wrapper
    return ctx.invoke(func, cli_config, *args[1:], **kwargs)
  File "/venv/lib/python3.9/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/venv/lib/python3.9/site-packages/rq/cli/cli.py", line 248, in worker
    worker = cli_config.worker_class(
  File "/venv/lib/python3.9/site-packages/rq/worker.py", line 224, in __init__
    connection.client_setname(self.name)
  File "/venv/lib/python3.9/site-packages/redis/client.py", line 1206, in client_setname
    return self.execute_command('CLIENT SETNAME', name)
  File "/venv/lib/python3.9/site-packages/redis/client.py", line 901, in execute_command
    return self.parse_response(conn, command_name, **options)
  File "/venv/lib/python3.9/site-packages/redis/client.py", line 915, in parse_response
    response = connection.read_response()
  File "/venv/lib/python3.9/site-packages/redis/connection.py", line 756, in read_response
    raise response
[...]

PR #1514, opened 2021-07-14
Handle ResponseError when CLIENT command is not supported

cesarferradas:
The `CLIENT` redis command is not supported in Google Memorystore Redis. This PR catches the `ResponseError` thrown when `rq` tries to use the `CLIENT` command, logs a warning and sets the `ip_address` to `"unknown"`.

See https://github.com/rq/rq/issues/1508 for related issue.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -221,8 +221,13 @@ class Worker:
         if prepare_for_work:
             self.hostname = socket.gethostname()
             self.pid = os.getpid()
-            connection.client_setname(self.name)
-            self.ip_address = [client['addr'] for client in connection.client_list() if client['name'] == self.name][0]
+            try:
+                connection.client_setname(self.name)
+            except redis.exceptions.ResponseError:
+                warnings.warn('CLIENT command not supported, setting ip_address to unknown')
+                self.ip_address = 'unknown'
+            else:
+                self.ip_address = [client['addr'] for client in connection.client_list() if client['name'] == self.name][0]
         else:
             self.hostname = None
             self.pid = None
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -224,7 +224,10 @@ class Worker:
             try:
                 connection.client_setname(self.name)
             except redis.exceptions.ResponseError:
-                warnings.warn('CLIENT command not supported, setting ip_address to unknown')
+                warnings.warn(
+                    'CLIENT command not supported, setting ip_address to unknown',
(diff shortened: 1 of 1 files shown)


rpkak:
Have you tested this?

I don't think that would work.

https://github.com/rq/rq/blob/1868bad9f7bd0f938b8fb301a798d9d97ed1cc0f/rq/worker.py#L224

This line (the one below the one you removed) uses the name feature.

cesarferradas:
Hi @rpkak, testing now but there's a delay as I can only reproduce when I deploy to GCP. You're right though, I need to also change that line. If I do, it will stop populating `ip_address` so it'll break features in `rq info`. Do you have a suggestion for a workaround?

rpkak:
You can catch the error, [print a warning](https://docs.python.org/3/library/warnings.html#warnings.warn) and set the `worker.ip_address` to unknown.

For testing: If it doesn't work on any computer it won't be merged and I think it won't work at GCP as well. Just execute `RUN_SLOW_TESTS_TOO=1 pytest --durations=5`

rpkak:
I would not use `UserWarning` but `Warning`.

cesarferradas:
> You can catch the error, [print a warning](https://docs.python.org/3/library/warnings.html#warnings.warn) and set the `worker.ip_address` to unknown.
> 
> For testing: If it doesn't work on any computer it won't be merged and I think it won't work at GCP as well. Just execute `RUN_SLOW_TESTS_TOO=1 pytest --durations=5`

@rpkak thanks for the suggestion, great idea. Have added this to the PR, and locally all tests now pass except `tests/test_commands.py::TestCommands`, but I think it might be to do with my local not being set up correctly.

Here's the stack trace:
Screenshot 2021-07-14 at 15 19 47

If this is something to do with my env, could we try running the GitHub workflows?

cesarferradas:
> I would not use `UserWarning` but `Warning`.

@rpkak have changed to Warning

rpkak:
I would but I can't. I try to run it at my machine.

rpkak:
Ok works . You can also execute `./run_tests_in_docker.sh` to run use docker for testing.

cesarferradas:
Great! Thanks for your help. What are the next steps to get this merged and released?

rpkak:
Waiting for a collaborator (@selwin).

selwin:
Thanks!

PR #1526, opened 2021-07-29
Docs in Sphinx

### LINKED ISSUE #111: Converting the existing gh-pages docs to Sphinx
I want the existing `gh-pages` docs to be converted to Sphinx-based documentation generation, and move the docs over to readthedocs.org. However, I'm not willing to sacrifice the current layout/design of the RQ documentation.

If someone sees a way to combine the two, I'd be very happy to hear from them.

rpkak:
I moved the documentation from [jekyll](https://jekyllrb.com/) to [sphinx](https://www.sphinx-doc.org/) (@nvie #111).

This PR is not finished yet (you see that this is a draft PR), because I will change some things in the content. So @selwin you can review the layout.

For this PR I created the folder docssphinx. In the end I will move this to docs.

@selwin because there are other PRs that change the doc, I think the best option is, that I implement the new docs after you approved this PR.

Closes #111


selwin:
Hey, so the issue you're referring too is super old. Back then, the docs was still hosted on `gh-pages` branch but I eventually merged it into the master branch.

I much prefer markdown than RST and I don't have any complaints about the docs site being built in Jekyll.

rpkak:
The biggest thing why I made this PR was because readthedocs has version control and so we wouldn't always have to add `New in version 1.9.0.` like [here](https://python-rq.org/docs/#bulk-job-enqueueing).

You also don't have to use the browser but you can generate pdf, epub or man pages as well.

If you prefer md I can port the content pages back to md but I wouldn't port files like `docs/index.rst` or `api.rst`, because they only contain sphinx contents like toctree or autoclass. I don't think that will be a problem because you can use markdown and restructuredtext at the same documentation.

rpkak:
@selwin If you want to look at this documentation, I created a temporary readthedocs link [here](https://testrqtest.rtfd.io).

selwin:
> Also look at the API Reference. This is also impossible for Jekyll.

I think this is the only real win for me.

> I ported every restructuredtext file to MyST (Markedly Structured Text, Markdown but with sphinx options).

I think I still prefer using `.md` or some of the more well supported markdown flavors for portability purposes (maybe CommonMark or Github Flavored Markdown?). This will make it easier in case that we need to switch to a different docs infrastructure.

I looked at the test site you made; some of the styling is a bit messy so I suppose if we end up switching to Sphinx, we should use one of their more established themes.

My worry with this accepting a pull PR like this is that I don't fully understand Sphinx and the build process so it may be a bit troublesome to maintain in the long run.

Let me think about this a bit more. I will get around to reviewing your CLI PR sometime this weekend, I promise.

Thanks for all your efforts, tackling big issues like this. It's much appreciated.

rpkak:
@selwin 
> I think I still prefer using `.md` or some of the more well supported markdown flavors for portability purposes (maybe CommonMark or Github Flavored Markdown?). This will make it easier in case that we need to switch to a different docs infrastructure.

If you look at 0d3f7d8d50e4a26b3d5afbdbb9eb6ba21fbe5f2e, you see that I switched to [MyST](https://myst-parser.readthedocs.io/en/latest/). This is a Markdown flavor which supports rst features:

https://github.com/rq/rq/blob/c4565525e920bec71fb6ffd2832940f20e883915/docssphinx/conf.py


> 
> I looked at the test site you made; some of the styling is a bit messy so I suppose if we end up switching to Sphinx, we should use one of their more established themes.
> 
> My worry with this accepting a pull PR like this is that I don't fully understand Sphinx and the build process so it may be a bit troublesome to maintain in the long run.
> 
> Let me think about this a bit more. I will get around to reviewing your CLI PR sometime this weekend, I promise.
> 
> Thanks for all your efforts, tackling big issues like this. It's much appreciated.

rpkak:
@selwin 
> I think I still prefer using `.md` or some of the more well supported markdown flavors for portability purposes (maybe CommonMark or Github Flavored Markdown?). This will make it easier in case that we need to switch to a different docs infrastructure.

If you look at 0d3f7d8d50e4a26b3d5afbdbb9eb6ba21fbe5f2e, you see that I switched to [MyST](https://myst-parser.readthedocs.io/en/latest/). This is a Markdown flavor which supports rst features:

https://github.com/rq/rq/blob/c4565525e920bec71fb6ffd2832940f20e883915/docssphinx/docs/index.md#L1-L17


> I looked at the test site you made; some of the styling is a bit messy so I suppose if we end up switching to Sphinx, we should use one of their more established themes.

Ok, which [themes](https://sphinx-themes.org/#themes) do you prefer.

selwin:
> If you look at 0d3f7d8, you see that I switched to MyST. This is a Markdown flavor which supports rst features:

I'm aware, but what I'm saying is I'd rather switch to common mark or Github flavored one for portability purposes (easier to move to a different docs infrastructure :).

> Ok, which themes do you prefer.

I think Furo: https://sphinx-themes.org/sample-sites/furo/

PR #1536, opened 2021-08-11
job: add get_meta() function

aparcar:
The newly introduced function returns meta data stored for the job. This
is required since job.meta stays an empty dict until the job is
finished or failed.

With the new function it's possible to store arbiatraty states/stages of
the job and allow the user to track progress. A long running job may
return custom stages like `downloading_data`, `unpacking_data`,
`processing_data`, etc.

This may allow better interfaces since users can track progress.

### DIFF
--- a/docs/docs/jobs.md
+++ b/docs/docs/jobs.md
@@ -100,6 +100,8 @@ print('Status: %s' % job.get_status())
 
 Some interesting job attributes include:
 * `job.get_status()` Possible values are `queued`, `started`, `deferred`, `finished`, `stopped`, `scheduled` and `failed`
+* `job.get_meta(refresh=True)` Returns custom `job.meta` dict containing user
+  stored data. If `refresh` is `True` fresh values are fetched from Redis.
 * `job.origin` queue name of this job
 * `job.func_name`
 * `job.args` arguments passed to the underlying job function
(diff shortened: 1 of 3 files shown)


aparcar:
Might as well extend the get_status function a la `def get_status(include_meta=False)` to do the trick instead of using another function... just realize.

aparcar:
Friendly ping

docs/docs/jobs.md
```diff
@@ -100,6 +100,7 @@ print('Status: %s' % job.get_status())
 
 Some interesting job attributes include:
 * `job.get_status()` Possible values are `queued`, `started`, `deferred`, `finished`, `stopped`, `scheduled` and `failed`
+* `job.get_meta()` Returns custom `job.meta` dict containing user stored data
```
selwin, line 103: Can we add the `refresh` argument here? Also we should clarify that this will read `meta` directly from Redis if `refresh=True`

aparcar: Done

rq/job.py
```diff
@@ -158,6 +158,13 @@ def set_status(self, status, pipeline=None):
         connection = pipeline if pipeline is not None else self.connection
         connection.hset(self.key, 'status', self._status)
 
+    def get_meta(self, refresh=True):
```
selwin, line 161: Can we test this function? You added a test case but I don't think `job.get_meta` is tested.

aparcar: Awkward, you're right. I fixed it.

selwin:
Thanks!

aparcar:
Thanks for the review, please let me know once this ends up in a release.

PR #1538, opened 2021-08-16
Respect serializer

### LINKED ISSUE #1486: Custom Serializer not supported in all Job fetch calls
I'm using `msgpack` as a custom serializer, which works fine when enqueuing and retrieving Jobs. However, when trying to use more advanced features of rq such as sending a stop command to a worker, I get the following error:
```
self.meta = self.serializer.loads(obj.get('meta')) if obj.get('meta') else {}
_pickle.UnpicklingError: unpickling stack underflow
```
This is because there's no option to pass in a serializer to the `send_stop_job_command` function, so `Job.fetch` is missing the right serializer and `Job.restore()` fails to deserialize.
This happens in a couple of other places, such as in the job static methods `cancel_job` and `requeue_job`, and in the registry `remove` function.

joshcoden:
Fixes #1486

See also: https://github.com/rq/rq/pull/1466#discussion_r692705397


selwin:
Whoa, thanks for this PR. Please give me a few days to review this.

tests/test_queue.py
```diff
@@ -776,6 +777,15 @@ def test_getting_registries(self):
...
         self.assertEqual(queue.finished_job_registry, FinishedJobRegistry(queue=queue))
 
+    def test_getting_registries_with_serializer(self):
+        """Getting job registries from queue object (with custom serializer)"""
+        queue = Queue('example', serializer=JSONSerializer)
+        self.assertEqual(queue.scheduled_job_registry, ScheduledJobRegistry(queue=queue))
+        self.assertEqual(queue.started_job_registry, StartedJobRegistry(queue=queue))
+        self.assertEqual(queue.failed_job_registry, FailedJobRegistry(queue=queue))
+        self.assertEqual(queue.deferred_job_registry, DeferredJobRegistry(queue=queue))
+        self.assertEqual(queue.finished_job_registry, FinishedJobRegistry(queue=queue))
```
selwin, line 787: Do you mind providing a more complete test case? The test cases should include cases which would cause RQ to crash prior to this patch.

tests/test_registry.py
```diff
@@ -221,14 +235,33 @@ def test_clean_registries(self):
...
+        queue = Queue(connection=self.testconn, serializer=JSONSerializer)
+
+        finished_job_registry = FinishedJobRegistry(connection=self.testconn, serializer=JSONSerializer)
+        self.testconn.zadd(finished_job_registry.key, {'foo': 1})
+
+        started_job_registry = StartedJobRegistry(connection=self.testconn, serializer=JSONSerializer)
+        self.testconn.zadd(started_job_registry.key, {'foo': 1})
+
+        failed_job_registry = FailedJobRegistry(connection=self.testconn, serializer=JSONSerializer)
+        self.testconn.zadd(failed_job_registry.key, {'foo': 1})
```
selwin, line 250: Same thing here, do you mind adding test cases which would fail prior to this patch?

joshcoden: @selwin This would have failed before, because the cleanup registry method on `BaseRegistry` calls `fetch` on the job class which in the past did not pass a serializer to that call, so this can stay.

selwin:
Thanks!

PR #1546, opened 2021-08-21
job.cancel() puts job into CanceledJobRegistry.

selwin:

PR #1549, opened 2021-08-23
Add option to enqueue a jobs dependents when canceling

joshcoden:
After seeing #1546 merged, I have a use case where if I cancel a job, I would like to be able to optionally enqueue its dependents, without having to mark it as finished.  This PR does not make this the default but lets users do so if they want to.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -16,6 +16,8 @@ from enum import Enum
 from functools import partial
 from uuid import uuid4
 
+from redis import WatchError
+
 from rq.compat import as_text, decode_redis_hash, string_types
 from .connections import resolve_connection
 from .exceptions import DeserializationError, NoSuchJobError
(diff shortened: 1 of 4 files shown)


rq/job.py
```diff
@@ -676,27 +678,47 @@ def save_meta(self):
...
-            registry = CanceledJobRegistry(self.origin, self.connection, job_class=self.__class__)
-            registry.add(self, pipeline=pipeline)
-        pipeline.execute()
+        You can enqueue the jobs dependents optionally, 
+        Same pipelining behavior as Queue.enqueue_dependents on whether or not a pipeline is passed in.
+        """
+        pipe = pipeline or self.connection.pipeline()
+        while True:
+            try:
+                if self.origin:
```
selwin, line 695: We can assume `self.origin` is always there. This check was added a long time ago when the `origin` property was newly added.

joshcoden: @selwin You want me to remove tis check?

PR #1552, opened 2021-08-29
Config Object

### LINKED ISSUE #1417: send_stop_job_command uses built-in Job class and does not consider custom serializers when doing Job.fetch
I have both a custom job class and use the JSON serializer, but if I call send_stop_job_command both neither my CustomJob class or the serializer is used when fetching the state from redis, and therefore causing the application to fail due to a unpickling error.

### LINKED ISSUE #1548: Idea: Some kind of config object
The idea is that because there are some configurable things like job class, worker class, instance of redis, logging format, etc.
Some of them have to be overgiven each time a new method is called and looking at #1538 or cli.py where non of the commands use both `serializer=serializer` and `job_class=job_class`.

The config object would be something like the cli-config, but as a context mananger that can be used by everyone.
This would make passing all the configuration through all functions unnecessary and the next time we a new configuration is added we it hasn't to be implemented to every single function.

I can create a PR for this, so please write me how you think about this idea.

rpkak:
This PR adds a config object as a context manager.

The user can update the config by doing:

```python
from rq.config import Config

with Config(worker_class='rq.SimpleWorker', serializer_class='rq.serializers.JSONSerializer'):
    # Do things where the default worker class is SimpleWorker and the serializer is json
```

In RQ this data can be accessed using:

```python
from rq.config import get_configuration

serializer_class = get_configuration('serializer_class')
```

closes #1548
closes #1417

### DIFF
--- a/rq/config.py
+++ b/rq/config.py
@@ -1,0 +1,108 @@
+from redis import Redis
+
+from .connections import get_current_connection
+from .defaults import (DEFAULT_JOB_CLASS, DEFAULT_QUEUE_CLASS,
+                       DEFAULT_SERIALIZER_CLASS, DEFAULT_WORKER_CLASS)
+from .serializers import resolve_serializer
+from .utils import resolve_class
+
+
+class Config:
+    def __init__(self, template=None, job_class=None, queue_class=None,
+                 worker_class=None, serializer=None, connection=None):
+        self._template = template
+        self._job_class = job_class
+        self._queue_class = queue_class
+        self._worker_class = worker_class
+        self._serializer = serializer
+        self._connection = connection
+
+    @property
+    def job_class(self):
+        if self._job_class is not None:
+            return resolve_class(self._job_class)
+        elif self._template is not None:
+            return self._template.job_class
+        else:
+            return DEFAULT_CONFIG.job_class
+
+    @property
+    def queue_class(self):
+        if self._queue_class is not None:
+            return resolve_class(self._queue_class)
+        elif self._template is not None:
+            return self._template.queue_class
+        else:
+            return DEFAULT_CONFIG.queue_class
+
+    @property
+    def worker_class(self):
+        if self._worker_class is not None:
+            return resolve_class(self._worker_class)
+        elif self._template is not None:
+            return self._template.worker_class
+        else:
+            return DEFAULT_CONFIG.worker_class
+
+    @property
+    def serializer(self):
+        if self._serializer is not None:
+            return resolve_serializer(self._serializer)
+        elif self._template is not None:
+            return self._template.serializer
+        else:
+            return DEFAULT_CONFIG.serializer
+
+    @property
+    def connection(self):
+        if self._connection is not None:
+            return self._connection
+        elif self._template is not None:
+            return self._template.connection
+        else:  # don't break deprecated connections.py
+            current_connection = get_current_connection()
+            return Redis() if current_connection is None else current_connection
+
+    def __repr__(self):
+        if self == DEFAULT_CONFIG:
+            return 'DEFAULT_CONFIG'
+        return '%s(\n    job_class=%r\n    queue_class=%r\n    worker_class=%r\n    serializer=%r\n    connection=%r\n)' % (
+            self.__class__.__name__,
+            self.job_class,
+            self.queue_class,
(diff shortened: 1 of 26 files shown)


selwin:
Implementation wise, this config object shouldn't be accessed as though this is a global or thread local object like 
```python
from rq.config import get_configuration
serializer_class = get_configuration('serializer_class')
```

I have plans to deprecate RQ's usage of thread locals. So this config object should be explicitly passed to `Worker`, `Queue`, `Job` or `Scheduler`. So it should be something like this:
```python
worker = Worker(config=config)
worker.config.get('serializer_class')
```

rpkak:
> I have plans to deprecate RQ's usage of thread locals. So this config object should be explicitly passed to `Worker`, `Queue`, `Job` or `Scheduler`. So it should be something like this:
> 
> ```python
> worker = Worker(config=config)
> worker.config.get('serializer_class')
> ```

The problem of this is that we always need to use the config as an arg and so we could forget to implement it (For Example in functions of this project which are called from multiple other functions that don't need config now but in the future).

That's because I used context manager and a global config stack to make the config accessable from everywhere.

So @selwin can you just write me why you don't like globals?

selwin:
> The problem of this is that we always need to use the config as an arg and so we could forget to implement it (For Example in functions of this project which are called from multiple other functions that don't need config now but in the future).

Correct, but the goal would be to put everything, including connection information into this config object.

> So @selwin can you just write me why you don't like globals?

I like it when everything is passed around so there's no ambiguity. There are many threads discussing why globals shouldn't be used, this is simply one of them https://stackoverflow.com/questions/19158339/why-are-global-variables-evil

To me personally it's easier to maintain a code base where everything is mostly passed from one function to another. No surprises.

rpkak:
Ok @selwin it's used as an arg, that is passed now and I implemented the cli.

It is very ugly yet in the cli because there is no possibility to pass the config yet.
That will get better when I implement worker, queue and job.

rq/config.py
```diff
@@ -0,0 +1,66 @@
...
+    'job_class': resolve_class,
+    'queue_class': resolve_class,
+    'worker_class': resolve_class,
+    'serializer': resolve_serializer,
+    'connection': _as_is
+}
+
+
+class Config:
+    def __init__(self, template=None, **kwargs):
```
selwin, line 22: I'd like to have a more explicitly define config object so we can have an API like this:

```python
config = get_config()
config.serializer_class
config.connection
```

selwin:
The most important thing about this PR is that it breaks backwards compatibility. I'm not willing to make a backwards incompatible change without advanced warning.

This means that `config` should be an additional kwarg on top of currently existing ones. We can raise a `DeprecationWarning` for arguments that will be deprecated in version 2.X, but the actual deprecation will happen after the warning has been in place for a few years.

docs/docs/index.md
```diff
@@ -46,10 +44,20 @@ time.sleep(2)
...
+
+config = Config(connection=Redis(REDIS_IP_OR_DOMAIN))
+q = Queue(config=config)
+```
+
 If you want to put the work on a specific queue, simply specify its name:
 
 ```python
-q = Queue('low', connection=redis_conn)
+q = Queue('low')
```
selwin, line 60: I don't think we can get rid of `connection` for these reasons:
1. Storage must be explicitly defined
2. I want users to be able to easily use RQ with default options without having to worry about config

So the minimum way to define a queue would be `queue = Queue(connection=redis)`.

selwin:
@rpkak I'm not sure. Does the test work on your local machine?

lowercase00:
Hey @rpkak did you end up doing some more work on this?
I got here while thinking about namespaces (eg `myns:job` instead of `rq:job`).

His has appeared in a couple of issues around here. A very trivial implementation for namespaces would be to read an envvar `RQ_NAMESPACE` and then make all the hardcoded namespaces read from a `DEFAULT_NAMESPACE = os.getenv("RQ_NAMESPACE", "rq")`. I run a couple of tests and this works, but would allow for an inconsistent state (the worker is running without the envvar for example).

I was though of having a config being stored within the datastore in `rq:config`, which would allow for a centralized configuration and all resources could read it. So the config object being passed to the queue could add data to `rq:config`.

PR #1560, opened 2021-09-10
Add missing functionality for CanceledJobRegistry

### LINKED ISSUE #1557: `rq.Queue` class missing `canceled_job_registry` property
The `CanceledJobRegistry` class was added recently in #1546, however, no property for it was added to the `Queue` class that exists for all the other registries:

https://github.com/rq/rq/blob/08a4315e8d981b7036f9107ee9a673089eeddd5f/rq/queue.py#L228-L257

### LINKED ISSUE #1558: `CanceledJobRegistry` does not define `cleanup()` causing errors when trying to call `count()` and `get_job_ids()`
Unlike other registries, the `CanceledJobRegistry` does not define a `cleanup` method meaning that it will always raise an error when trying to call any methods defined on `BaseRegistry` that assume an implementation for `cleanup()` exists on implementors.  

See this note on the implementation of the `ScheduledJobRegistry`'s (also on `DeferredJobRegistry`) empty implementation of the `cleanup` method that should probably be copied to `CanceledJobRegistry`:

https://github.com/rq/rq/blob/08a4315e8d981b7036f9107ee9a673089eeddd5f/rq/registry.py#L292-L296

joshcoden:
Fixes #1557, Fixes #1558

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -256,6 +256,12 @@ class Queue:
         from rq.registry import ScheduledJobRegistry
         return ScheduledJobRegistry(queue=self, job_class=self.job_class, serializer=self.serializer)
 
+    @property
+    def canceled_job_registry(self):
+        """Returns this queue's CanceledJobRegistry."""
+        from rq.registry import CanceledJobRegistry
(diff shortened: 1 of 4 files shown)


selwin:
Thanks!

PR #1561, opened 2021-09-11
Fix link Round Robin / Random worker classes

luto:
The link within the [Custom Worker Classes](https://python-rq.org/docs/workers/#custom-worker-classes) section currently does not work. Changing it to be all lowercase fixes that.


selwin:
Thanks!

Issue #1562, opened 2021-09-14
Please create a 1.10.1 release


joshcoden:
The version number was recently updated, but no new release was created:

https://github.com/rq/rq/blob/0d69d081264b8465d93d8447853e777e85573a34/rq/version.py#L5

Could a version 1.10.1 be created from the current master branch?

lowercase00:
@selwin This can probably be closed

PR #1564, opened 2021-09-20
Fix `job.cancel` to remove job from registries if not in queue

joshcoden:
The `job.cancel` function was calling `remove` on the queue, assuming a job is in a queue, not in a registry, this assumption should not be made and this change ensures that it checks not only the queue but also its registries.

Would be nice to get this merged before #1562 is done.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -710,7 +710,6 @@ class Job:
                     pipeline=pipe,
                     remove_from_queue=True
                 )
-                q.remove(self, pipeline=pipe)
 
                 self.set_status(JobStatus.CANCELED, pipeline=pipe)
 
--- a/rq/job.py
+++ b/rq/job.py
@@ -708,7 +708,7 @@ class Job:
                     if pipeline is None:
                         pipe.watch(self.dependents_key)
                     q.enqueue_dependents(self, pipeline=pipeline)
-                self._remove_job_from_container(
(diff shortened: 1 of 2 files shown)


rq/job.py
```diff
@@ -733,13 +738,7 @@ def requeue(self):
...
         return self.failed_job_registry.requeue(self)
 
-    def delete(self, pipeline=None, remove_from_queue=True,
-               delete_dependents=False):
-        """Cancels the job and deletes the job hash from Redis. Jobs depending
-        on this job can optionally be deleted as well."""
-
-        connection = pipeline if pipeline is not None else self.connection
-
+    def _remove_job_from_container(self, pipeline=None, remove_from_queue=True):
```
selwin, line 741: I think we can rename this method to something more explicit like `job._remove_from_registries()`

joshcoden: @selwin I didn't name it that since it can be also be removed from the queue itself as well.

selwin: There's the `remove_from_queue` argument so I think it's still ok. I'm open to a other suggestions, but I think `container` is a bit weird since we don't have this terminology in RQ.

joshcoden: @selwin Fair enough, I'll change it.

selwin:
Thanks!

PR #1568, opened 2021-09-24
Add new `QueuedJobRegistry` to catch jobs dropped by workers before m…

### LINKED ISSUE #1553: rq jobs go into a bad state when redis hits connection limit
Event: message in popped from a non empty queue
State: redis server's client connection count < max_connections

```
- worker.py: work()
  - dequeue_job_and_maintain_ttl
    - job, queue = queue_class.dequeue_any(...)  | message is popped from queue. The job
                                                 | corresponding to the message exists in 
                                                 | JobRegistry, but its state is set to queued.
  - execute_job(job, queue)
    - fork_work_horse(job, queue)
```
Event: New work horse is forked to execute the job
State: redis server's client connection count = max_connections
```
  - worker.py:fork_work_horse(job, queue)
    - main_work_horse(job, queue)
      - perform_job(job, queue)
        - push_connection(self.connection)
        - prepare_job_execution(job)
          - Creates a new connection here     |New connection can't be created here.
          - set job.hearbeat                  |doesnt work
            - Add to StartedJobRegistry       |doesnt work, job is not moved to StartedJobRegistry
```

For the same reason, executing the failure callback is also not possible.

At this point the job persists in redis with status="queued", but the job is not in queue. Workers
try to pop the next item in the queue and the whole cycle continues until the queue is empty. The
jobs still persist in redis and there is no way to process them. They also miss out from any
monitoring systems that monitor the rq registries as they are absent from all of them.

This problem might happen often when running rq on a redis server with low client connection limit.

Traceback that caused me to investigate the issue:

```python
  File "<base-path>/site-packages/rq/worker.py", line 1026, in perform_job
    self.prepare_job_execution(job)
  File "<base-path>/site-packages/rq/worker.py", line 908, in prepare_job_execution
    pipeline.execute()
  File "<base-path>/site-packages/redis/client.py", line 4013, in execute
    self.shard_hint)
  File "<base-path>/site-packages/redis/connection.py", line 1192, in get_connection
    connection.connect()
  File "<base-path>/site-packages/redis/connection.py", line 567, in connect
    self.on_connect()
  File "<base-path>/site-packages/redis/connection.py", line 643, in on_connect
    auth_response = self.read_response()
  File "<base-path>/site-packages/redis/connection.py", line 739, in read_response
    response = self._parser.read_response()
  File "<base-path>/site-packages/redis/connection.py", line 340, in read_response
    raise error
redis.exceptions.ConnectionError: max number of clients reached
During handling of the above exception, another exception occurred:
  File "<base-path>/site-packages/rq/worker.py", line 1052, in perform_job
    self.execute_failure_callback(job)
  File "<base-path>/site-packages/rq/worker.py", line 1013, in execute_failure_callback
    job.heartbeat(utcnow(), CALLBACK_TIMEOUT)
  File "<base-path>/site-packages/rq/job.py", line 434, in heartbeat
    connection.hset(self.key, 'last_heartbeat', utcformat(self.last_heartbeat))
  File "<base-path>/site-packages/redis/client.py", line 3050, in hset
    return self.execute_command('HSET', name, *items)
  File "<base-path>/site-packages/redis/client.py", line 898, in execute_command
    conn = self.connection or pool.get_connection(command_name, **options)
  File "<base-path>/site-packages/redis/connection.py", line 1192, in get_connection
    connection.connect()
  File "<base-path>/site-packages/redis/connection.py", line 567, in connect
    self.on_connect()
  File "<base-path>/site-packages/redis/connection.py", line 643, in on_connect
    auth_response = self.read_response()
  File "<base-path>/site-packages/redis/connection.py", line 739, in read_response
    response = self._parser.read_response()
  File "<base-path>/site-packages/redis/connection.py", line 340, in read_response
    raise error
redis.exceptions.ConnectionError: max number of clients reached
```

joshcoden:
…arked as started

Fixes #1553

### DIFF
--- a/docs/docs/job_registries.md
+++ b/docs/docs/job_registries.md
@@ -12,6 +12,8 @@ executed and removed right after completion (success or failure).
 job to finish).
 * `ScheduledJobRegistry` Holds scheduled jobs.
 * `CanceledJobRegistry` Holds canceled jobs.
+* `QueuedJobRegistry` Holds jobs from the time they are enqueued, until they are moved to the `QueuedJobRegistry`.
+This exists mostly to catch jobs from getting lost between being dequeued and then getting added to the `StartedJobRegistry`, if there are failure in between, this makes sure that jobs get requeued if a worker loses track of the job, notably from redis failures, like hitting connection limits, for example.
 
 You can get the number of jobs in a registry, the ids of the jobs in the registry, and more. 
 Below is an example using a `StartedJobRegistry`.
@@ -62,6 +64,8 @@ queue.deferred_job_registry   # Returns DeferredJobRegistry
 queue.finished_job_registry  # Returns FinishedJobRegistry
 queue.failed_job_registry  # Returns FailedJobRegistry
 queue.scheduled_job_registry  # Returns ScheduledJobRegistry
+queue.canceled_job_registry  # Returns CanceledJobRegistry
+queue.queued_job_registry  # Returns QueuedJobRegistry
 ```
 
 ## Removing Jobs
--- a/docs/docs/job_registries.md
+++ b/docs/docs/job_registries.md
@@ -14,7 +14,7 @@ job to finish).
 * `CanceledJobRegistry` Holds canceled jobs.
 * `QueuedJobRegistry` Holds jobs from the time they are enqueued, until they are moved to the `StartedJobRegistry`.
 This exists mostly to catch jobs from getting lost between being dequeued and then getting added to the `StartedJobRegistry`,
-if there are failure in between, this makes sure that jobs get requeued if a worker loses track of the job, notably from redis failures, like hitting connection limits, for example.
+if there are any failures in between, this makes sure that jobs get requeued if a worker loses track of the job, notably from redis failures, like hitting connection limits, for example.
 
 You can get the number of jobs in a registry, the ids of the jobs in the registry, and more. 
 Below is an example using a `StartedJobRegistry`.
--- a/docs/docs/job_registries.md
+++ b/docs/docs/job_registries.md
@@ -12,8 +12,9 @@ executed and removed right after completion (success or failure).
 job to finish).
 * `ScheduledJobRegistry` Holds scheduled jobs.
 * `CanceledJobRegistry` Holds canceled jobs.
-* `QueuedJobRegistry` Holds jobs from the time they are enqueued, until they are moved to the `QueuedJobRegistry`.
-This exists mostly to catch jobs from getting lost between being dequeued and then getting added to the `StartedJobRegistry`, if there are failure in between, this makes sure that jobs get requeued if a worker loses track of the job, notably from redis failures, like hitting connection limits, for example.
+* `QueuedJobRegistry` Holds jobs from the time they are enqueued, until they are moved to the `StartedJobRegistry`.
+This exists mostly to catch jobs from getting lost between being dequeued and then getting added to the `StartedJobRegistry`,
+if there are failure in between, this makes sure that jobs get requeued if a worker loses track of the job, notably from redis failures, like hitting connection limits, for example.
 
 You can get the number of jobs in a registry, the ids of the jobs in the registry, and more. 
 Below is an example using a `StartedJobRegistry`.
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -566,7 +566,7 @@ nd
 
         job.origin = self.name
         job.enqueued_at = utcnow()
-        # TODO: Specify xx operator to prev'tent jobs from being double added to registry
+        # TODO: Specify xx operator to prevent jobs from being multiple times in the registry
         # (really just updated when it shouldn't already be there?)
         if job.timeout is None:
             job.timeout = self._default_timeout
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -245,6 +245,12 @@ class Queue:
         return FinishedJobRegistry(queue=self, job_class=self.job_class, serializer=self.serializer)
 
     @property
+    def queued_job_registry(self):
+        """Returns this queue's QueuedJobRegistry."""
+        from rq.registry import QueuedJobRegistry
+        return QueuedJobRegistry(queue=self, job_class=self.job_class, serializer=self.serializer)
+
+    @property
     def deferred_job_registry(self):
         """Returns this queue's DeferredJobRegistry."""
         from rq.registry import DeferredJobRegistry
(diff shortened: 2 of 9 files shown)


selwin:
Thanks for the PR.

I think at any one time, a job can only exist in one registry. When a job is moved to `StartedJobRegistry`, it should also be removed from `QueuedJobRegistry` in a single pipeline call. Do you mind making this change?

docs/docs/job_registries.md
```diff
@@ -12,6 +12,8 @@ executed and removed right after completion (success or failure).
 job to finish).
 * `ScheduledJobRegistry` Holds scheduled jobs.
 * `CanceledJobRegistry` Holds canceled jobs.
+* `QueuedJobRegistry` Holds jobs from the time they are enqueued, until they are moved to the `QueuedJobRegistry`.
```
selwin, line 15: I think this should be `until they are moved to the `StartedJobRegistry``

joshcoden:
> > It is when you enqueue, it's just up a level in the call stack:
> 
> I meant when a job is popped off the queue and moved to `StartedJobRegistry` (when it's being worked on), it should also be removed from `QueuedJobRegistry`. As it is now, when a job is being worked on it exists in both `QueuedJobRegistry` and `StartedJobRegistry`

@selwin as mentioned above, I changed it to remove it, please see the code link in the comment above:  https://github.com/rq/rq/pull/1568#issuecomment-928559454: 
https://github.com/rq/rq/blob/8fd8de0742490b5d0fb33a9e9564a0ff609d52fe/rq/worker.py#L913

alella:
| fn_call                      | queue | QueuedJobRegistry | StartedJobRegistry | parent worker | forked worker |
|------------------------------|-------|-------------------|--------------------|---------------|---------------|
| enqueue                      | job   | job               |                    | -             | -             |
| dequeue_job_and_maintain_ttl | -     | job               |                    | active        | doesnt exist  |
| perform_job(job, queue)      | -     | job               |                    | inactive      | active        |
| [prepare_job_execution(job)](https://github.com/rq/rq/blob/8fd8de0742490b5d0fb33a9e9564a0ff609d52fe/rq/worker.py#L913-L914)   | -     | -                 | job                | inactive      | active        |

I think what @joshcoden is trying to say is prepare_job_execution calls `job.queued_job_registry.remove` and `job.heartbeat` in the same pipeline. [job.heartbeat](https://github.com/rq/rq/blob/8fd8de0742490b5d0fb33a9e9564a0ff609d52fe/rq/job.py#L445-L449) in turn calls `started_job_registry.add` using the same pipeline. So the job never exists in both `QueuedJobRegistry` and `StartedJobRegistry` at the same time

rq/queue.py
```diff
@@ -560,11 +566,13 @@ def enqueue_job(self, job, pipeline=None, at_front=False):
 
         job.origin = self.name
         job.enqueued_at = utcnow()
-
+        # TODO: Specify xx operator to prev'tent jobs from being double added to registry
```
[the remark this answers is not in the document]
joshcoden: @alella  @selwin Should `xx` just be set to `ttl`?  I guess it would just update the score if it already exists depending on which other arguments we pass in.

rq/registry.py
```diff
@@ -334,6 +333,72 @@ def cleanup(self):
...
+            try:
+                if front_timestamp:
+                    job = self.job_class.fetch(
+                        job_id,
+                        connection=self.connection,
+                        serializer=self.serializer
+                    )
+                    if job.enqueued_at:
+                        if job.enqueued_at < front_timestamp:
+                            self.requeue(job)
```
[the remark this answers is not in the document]
joshcoden: @alella The user can manually specify the score when calling `add` on the registry.  While no such calls will exist in code, a user code specify a `ttl` that does not align with the `enqueued_at` time:

https://github.com/rq/rq/blob/47110806d1e281f7ba314f837658e247eaaf7a88/rq/registry.py#L64-L72

rq/registry.py
```diff
@@ -334,6 +333,72 @@ def cleanup(self):
...
+                    else:
+                        # TODO: Should we do anything else?
```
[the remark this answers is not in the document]
joshcoden: @alella My only other thought would be to log an error and at least try other jobs, but this should really never happen so it's probably safer to just raise in this case.

selwin:
@alella @joshcoden thanks for explaination.

It is indeed working as expected. I misread this because the the code around this part is not symmetric. Can you create a `job.move_to_started_job_registry()` method that moves the job from `QueuedJobRegistry` into `StartedJobRegistry`? It will be cleaner this way.

With this, we can potentially also cleanup `job.heartbeat(xx)` because right now it does two jobs:
1. Adding job to started job registry
2. Updating the timestamp in started job registry

This already led to race condition bugs fixed in this PR https://github.com/rq/rq/pull/1550

rq/registry.py
```diff
@@ -334,6 +333,66 @@ def cleanup(self):
...
+        queue = self.get_queue()
+        jobs = queue.get_jobs(offset=0, length=1)
+        if jobs: # If the queue is non-empty
+            front_job = jobs[0]
+            if front_job.enqueued_at is not None:
+                return front_job.enqueued_at
+            else:
+                raise InvalidJobOperationError(
+                    "Queued job {} has no enqueue_at value!".format(front_job.id)
+                )
```
selwin, line 391: This could not happen, right? So I think we can skip this check.

rq/registry.py
```diff
@@ -334,6 +333,66 @@ def cleanup(self):
...
+
+    def requeue_stuck_jobs(self):
+        """This method requeues jobs that are popped off the queue but still belong to QueuedJobRegistry
+        It is not defined in cleanup since we don't want this being called everytime count or get_job_ids is called
+        """
+        job_ids = self.get_job_ids()
+        for job_id in job_ids:
+            # If job was enqueued AFTER the front of the queue it must have already been dequeued
+            # This is faster than seeing if the job is in the queue directly.
+            front_timestamp = self._get_front_queue_timestamp()
```
selwin, line 356: We can move this out of the loop so we don't keep getting new timestamp for every single job processed.

rq/registry.py
```diff
@@ -334,6 +333,66 @@ def cleanup(self):
...
+        automatically called by `count()` and `get_job_ids()` methods
+        implemented in BaseRegistry."""
+        pass
+
+    def requeue_stuck_jobs(self):
+        """This method requeues jobs that are popped off the queue but still belong to QueuedJobRegistry
+        It is not defined in cleanup since we don't want this being called everytime count or get_job_ids is called
+        """
+        job_ids = self.get_job_ids()
+        for job_id in job_ids:
```
selwin, line 353: This could be expensive if we have lots of jobs in `QueuedJobRegistry` and sometimes people have millions of jobs enqueued. I wonder if we have a way to only check a subset of the jobs.

rq/registry.py
```diff
@@ -352,3 +411,9 @@ def clean_registries(queue):
                                  job_class=queue.job_class,
                                  serializer=queue.serializer)
     registry.cleanup()
+
+    registry = QueuedJobRegistry(name=queue.name,
+                                 connection=queue.connection,
+                                 job_class=queue.job_class,
+                                 serializer=queue.serializer)
+    registry.requeue_stuck_jobs()
```
selwin, line 419: I'm not sure if we should automatically requeue stuck jobs because:
1. This operation could be heavy depending on the number of jobs you have in the registry
2. If requeued, jobs that are time sensitive could lead to unwanted outcome

lowercase00:
@joshcoden any plans to continue working on this?

PR #1573, opened 2021-10-16
Allow jobs to save multiple execution results

selwin:
Currently each job only stores the result of its last execution. This makes troubleshooting failed executions across multiple executions difficult.

This PR allows RQ to store multiple results and potentially allows jobs to be executed multiple times. This PR is not yet finished, just putting this out there for other people to comment on and gather feedback.


selwin:
Superseded by https://github.com/rq/rq/pull/1725

Issue #758, opened 2016-11-16
Job queued but not run


fossilet:
I have a job not run, in the log of worker: 

2016-11-15 23:18:34,617 11894 {q}default: [job details] (e47b96dc-d0f8-48b4-bbc9-090e6443080a)
2016-11-15 23:18:34,618 11894 Sent heartbeat to prevent worker timeout. Next one should arrive within 420 seconds.
2016-11-15 23:18:34,621 22069 Sent heartbeat to prevent worker timeout. Next one should arrive within 420 seconds.
2016-11-15 23:18:40,741 11894 Sent heartbeat to prevent worker timeout. Next one should arrive within 420 seconds.
2016-11-15 23:18:40,743 11894 
2016-11-15 23:18:40,743 11894 *** Listening on {q}high, {q}default...
2016-11-15 23:18:40,743 11894 Sent heartbeat to prevent worker timeout. Next one should arrive within 420 seconds.

22069 is the forked process to run the job. There should have been lines like below or error:

[timestamp] 22069 Sent heartbeat to prevent worker timeout. Next one should arrive within 420 seconds.
[timestamp] 22069 {q}high: Job OK (e47b96dc-d0f8-48b4-bbc9-090e6443080a)
[timestamp] 22069 Result is kept for 500 seconds

In redis, the job status is queued:

> hgetall rq:job:e47b96dc-d0f8-48b4-bbc9-090e6443080a
 1) "created_at"
 2) "2016-11-15T15:18:34Z"
 3) "status"
 4) "queued"
...

This happens only occasionally.

zealseeker:
I got the same problem. All the jobs I submitted were in a 'queued' status. using the [demo scripts](http://python-rq.org/docs/). If you use `job.perform()` and it will return the result indicating that the jobs is ok. But I don't know the the queue didn't "START TO WORK".

selwin:
Are you sure the queue and worker use the same Redis connection? I'm not aware of any bug in RQ that could lead to a worker not seeing jobs that are queued properly.

nolanbrown:
I've been having the same issue. The redis connection is setup properly, I've inspected everything but sometimes queued jobs stay queued. They don't timeout or ever get processed by the worker.

ssikiki:
+1 the same problem

My env:  rq==0.8.0   redis==2.8.19

I've found that worker sometimes doesn't fetch job .
[...]

PR #1605, opened 2021-12-14
Fixed wrong keys used to WATCH dependencies

th3hamm0r:
**TL;DR:**
This bug has opened a lot of possible race-conditions, since the watch-logic from redis did not fail anymore, if dependencies have been changed in parallel.

This seems related to #758, especially https://github.com/rq/rq/issues/758#issuecomment-791603930, where jobs were hanging in deferred state, but I can't say, that my changes fix all issues in #758.

**Details:**
I found this issue during an upgrade of a project from RQ 0.13.0 to 1.10.1, which heavily uses job dependencies. Fortunately, during first tests I saw hanging jobs in "deferred" state.
Actually, I'm a co-author of the whole watch-logic for dependencies introduced in RQ 0.7.1, which has fixed race-conditions for single dependencies here: https://github.com/rq/rq/issues/739
Unfortunately, with the switch to multiple job dependencies (in RQ 1.2), the job-ids of dependencies have not been prefixed correctly, so probably since 1.2 the whole watch-logic basically does not work correctly, which results in race-conditions.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -485,7 +485,8 @@ class Job:
         connection = pipeline if pipeline is not None else self.connection
 
         if watch and self._dependency_ids:
-            connection.watch(*self._dependency_ids)
+            connection.watch(*[self.key_for(dependency_id)
+                               for dependency_id in self._dependency_ids])
 
         jobs = [job
                 for job in self.fetch_many(self._dependency_ids, connection=self.connection, serializer=self.serializer)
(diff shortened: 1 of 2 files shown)


th3hamm0r:
Fyi: creating tests for these race-conditions is really hard. I've tried to modify the original test from https://github.com/rq/rq/pull/742/files#diff-5c94530959bc0ef10f7a2c0fe76bd4145e10bf6c377627c10042dab0ebcf5c4aR572, but failed... I would still leave that original test, even when it did not "find" the current issue, but it tests some race-issues from the past.

selwin:
Thanks!

Issue #1608, opened 2021-12-30
Child job dependencies?


mattchan-tencent:
Hi,

I've got a job which enqueues other jobs from the worker so I don't know the job ids of all the children from the head node.

Is there a way in rq to track dependencies so that all the children finish before a reduce step takes place? 

Thanks for your help!

selwin:
We don't have such a concept in RQ yet. You can however create a job that depends on multiple other jobs before it gets executed.

mattchan-tencent:
Hi @selwin,

I think I figured out a possible way to work around it. We can store the reduction job ID in the metadata, and as new child jobs are created, we fetch the reduction job and use Job.register_dependency(). We also propagate that ID into those children so they can do the same thing recursively. Does this work? Or am I stepping into some concurrency landmine?

This is a slightly different question, but is there a way to create dependent jobs that run regardless of whether the dependencies failed or not? Quite a few of the jobs in my pipeline fail, but I'm willing to accept them and move on. It seems like in a really large scale reduction job, these failures would be almost unavoidable.

mattchan-tencent:
Answering the second question after a bit of investigation:
The key bit of code is at https://github.com/rq/rq/blob/c5a1ef17345e17269085e7f72858ac9bd6faf1dd/rq/job.py#L990.

In order to accept failed jobs as well, I think we need either a custom error handler to deregister the job from its dependencies, or to modify the rq source and accept finished or failed. Are either of those sane solutions? I will PR if it is.

selwin:
> This is a slightly different question, but is there a way to create dependent jobs that run regardless of whether the dependencies failed or not? Quite a few of the jobs in my pipeline fail, but I'm willing to accept them and move on. It seems like in a really large scale reduction job, these failures would be almost unavoidable.

This is a lower hanging fruit that I also want to solve. 

> In order to accept failed jobs as well, I think we need either a custom error handler to deregister the job from its dependencies, or to modify the rq source and accept finished or failed. Are either of those sane solutions? I will PR if it is.

I'd rather build a proper solution for this. I really like how our [retry API](https://python-rq.org/docs/exceptions/#retrying-failed-jobs) turned out so let's do something similar:
```python
dependency = Dependency(jobs=[job_1, job_2], allow_failure=True)
child_job = queue.enqueue(foo, depends_on=dependency)
```

1. `child_job` would store the `allow_failure` (better name suggestions welcome) in Redis.
[...]

PR #1611, opened 2022-01-15
added Dependency class with allow_failures

mattchan-tencent:
Implementation of proposal in #1608

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -13,6 +13,7 @@ from collections.abc import Iterable
 from datetime import datetime, timedelta, timezone
 from enum import Enum
 from functools import partial
+from typing import List
 from uuid import uuid4
 
 from redis import WatchError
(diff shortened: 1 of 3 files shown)


rq/job.py
```diff
@@ -42,6 +43,18 @@ class JobStatus(str, Enum):
     CANCELED = 'canceled'
 
 
+class Dependency:
+    def __init__(self, jobs: List[Job], allow_failure: bool = False):
+        jobs = ensure_list(jobs)
+        if not all(isinstance(job, Job) for job in jobs):
```
selwin, line 49: This should accept job ID's in addition to job instances.

rq/job.py
```diff
@@ -78,7 +91,7 @@ class Job:
     @classmethod
     def create(cls, func, args=None, kwargs=None, connection=None,
                result_ttl=None, ttl=None, status=None, description=None,
-               depends_on=None, timeout=None, id=None, origin=None, meta=None,
+               depends_on: Dependency = None, timeout=None, id=None, origin=None, meta=None,
```
selwin, line 94: `depends_on` needs to be backwards compatible so we don't break existing codes. We'll need to check whether this is a `Dependency` instance before we use the new code path.

selwin:
In addition to my comments, please add tests and docs too. Thanks for the quick revert :)

selwin:
Superseded by https://github.com/rq/rq/pull/1681

PR #1615, opened 2022-01-24
Fixed job not getting enqueued due to parallel modification

th3hamm0r:
I've found another bug in the dependency handling, which seems to be introduced with https://github.com/rq/rq/commit/456743b225c0b5a4339242a5b714c378b86f2fe0. It results in no job execution, if unfinished dependencies exist at the beginning, but all of them get finished in parallel.

**More details:**
If all of the unfinished dependencies get finished in parallel to the enqueuing of the job, the `execute()`-call will fire a WatchError [here](https://github.com/rq/rq/blob/c5a1ef17345e17269085e7f72858ac9bd6faf1dd/rq/queue.py#L368). This will leave the `job` instance in a changed state, where status is set to `DEFERRED` ([see](https://github.com/rq/rq/blob/c5a1ef17345e17269085e7f72858ac9bd6faf1dd/rq/queue.py#L363)).
In the following retry, no unfinished dependencies would be found, so the loop will be exited [here](https://github.com/rq/rq/blob/c5a1ef17345e17269085e7f72858ac9bd6faf1dd/rq/queue.py#L370) and the job will be returned.
This will then skip the actual `enqueue_job()` call [here](https://github.com/rq/rq/blob/c5a1ef17345e17269085e7f72858ac9bd6faf1dd/rq/queue.py#L405).

This was hard to find, because in that case `setup_dependencies()` does not save the `DEFERRED` job to redis, and since `enqueue_job()` was also not called, also no `QUEUED` job gets written.

For now I've only found the `_status` as the only local state that needs to be reset before enqueuing the job.

Related: #1605, #758

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -343,6 +343,7 @@ class Queue:
         # something else has modified either the set of dependencies or the
         # status of one of them. In this case, we simply retry.
         if len(job._dependency_ids) > 0:
+            orig_status = job.get_status(refresh=False)
             pipe = pipeline if pipeline is not None else self.connection.pipeline()
             while True:
                 try:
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

PR #1618, opened 2022-02-04
rq.worker: remove useless set_state call in horse

xavfernandez:
The state should already have been set properly by the worker in
`execute_job`

`prepare_job_execution` is only called by `perform_job` which should only be
called by `main_work_horse`/`fork_work_horse` (themselves only called by `execute_job`).
Let `execute_job` do the bookkeeping.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -922,7 +922,6 @@ class Worker:
         """
 
         with self.connection.pipeline() as pipeline:
-            self.set_state(WorkerStatus.BUSY, pipeline=pipeline)
             self.set_current_job_id(job.id, pipeline=pipeline)
             self.set_current_job_working_time(0, pipeline=pipeline)
 
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -1194,7 +1194,9 @@ class SimpleWorker(Worker):
 
     def execute_job(self, job, queue):
         """Execute job in same thread/process, do not fork()"""
-        return self.perform_job(job, queue)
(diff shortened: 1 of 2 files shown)


selwin:
Is there a reason why this change is made? Moving the `set_state()` call to `execute_job()` leads to an additional Redis call.

xavfernandez:
> Is there a reason why this change is made?

This is not needed per se but seems cleaner to me. Now only the worker is responsible for updating its state via `set_state` and the horse doesn't anymore.

> Moving the set_state() call to execute_job() leads to an additional Redis call.

I removed the `set_state` from `execute_job` since `SimpleWorker` doesn't seem to update its state.

xavfernandez:
We might want to update `SimpleWorker`'s `execute_job` to:

```
    def execute_job(self, job, queue):
        """Execute job in same thread/process, do not fork()"""
        self.set_state(WorkerStatus.BUSY)
        self.perform_job(job, queue)
        self.set_state(WorkerStatus.IDLE)
```

but I'm not sure if it is needed/wanted.

selwin:
@xavfernandez yes please! As much as possible, we should make it so that `SimpleWorker`'s behavior is identical to the default `Worker` class.

xavfernandez:
> yes please! As much as possible, we should make it so that SimpleWorker's behavior is identical to the default Worker class.

done :)

selwin:
Thanks!

PR #1629, opened 2022-02-28
Cross platform simple worker

caffeinatedMike:
Replaced `SimpleWorker`'s use of signals that are not available on Windows platform with a `threading.Timer` approach that should be platform-agnostic and allow for local testing.

Tests related to `SimpleWorker` pass:
`pytest ./tests/test_worker.py::TestWorker::test_work_via_simpleworker`
`pytest ./tests/test_worker.py::TestWorker::test_simpleworker_heartbeat_ttl`

Closes #1625

PR #1630, opened 2022-03-02
Added Dependency class with allow_failures (continued)

### LINKED ISSUE #1224: What is the idiomatic way in Python-RQ to handle dependent jobs, when the job they depend on fails?
What is the idiomatic way in Python-RQ to handle dependent jobs, when the job they depend on fails?

I work on a media transformation service. I have a download job, on this job depends a transformation job, on this job depends an upload job. If the download job fails for any reason, the depending jobs are obsolete. I can think of the following solutions:

1) Let the download job fail with a Python Exception. Python-RQ will file it as failed and remove the depending jobs.
2) Explicitly remove depending jobs with the job's delete function.
3) Return failure/success information as result of a job and interpret it in the depending jobs.
4) Do not communicate anything between jobs, just let each job check on its own, if its necessary prerequisites are met.

What solution would you choose, and why? Are there other, better solutions?

### LINKED ISSUE #1608: Child job dependencies?

caffeinatedMike:
Applies logic that was started in #1611, fixed a few typos, and added additional logic to handle the new `Dependency` class.

Tests have been updated/added to test functionality ~~, but docs have not been updated~~.

**Edit:** Docs have now been updated.

Closes #1224 
Closes #1608 
Closes #1611

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -155,8 +155,32 @@ baz_job = queue.enqueue(baz, depends_on=[foo_job, bar_job])
 ```
 
 The ability to handle job dependencies allows you to split a big job into
-several smaller ones. A job that is dependent on another is enqueued only when
+several smaller ones. By default, a job that is dependent on another is enqueued only when
 its dependency finishes *successfully*.
+
+_New in 1.11.0._
+
+If you want a job's dependencies to execute regardless if the job completes or fails, RQ provides
+the `Dependency` class that will allow you to dictate how to handle job failures.
+
+The `Dependency(jobs=...)` parameter accepts:
+- a string representing a single job id
+- a Job object
+- an iteratable of job id strings and/or Job objects
+
+Example:
+
+```python
+from redis import Redis
+from rq.job import Dependency
+from rq import Queue
+
+queue = Queue(connection=Redis())
+job_1 = queue.enqueue(div_by_zero, on_success=report_success, on_failure=report_failure)
+dependency = Dependency(jobs=[job_1], allow_failure=True)  # allow_failure defaults to False
+job_2 = queue.enqueue(say_hello, depends_on=dependency, on_success=report_success, on_failure=report_failure)
+# job_2 will execute even though its dependency (job_1) fails
+```
 
 
 ## Job Callbacks
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -176,9 +176,9 @@ from rq.job import Dependency
 from rq import Queue
 
 queue = Queue(connection=Redis())
-job_1 = queue.enqueue(div_by_zero, on_success=report_success, on_failure=report_failure)
+job_1 = queue.enqueue(div_by_zero)
 dependency = Dependency(jobs=[job_1], allow_failure=True)  # allow_failure defaults to False
-job_2 = queue.enqueue(say_hello, depends_on=dependency, on_success=report_success, on_failure=report_failure)
+job_2 = queue.enqueue(say_hello, depends_on=dependency)
 # job_2 will execute even though its dependency (job_1) fails
 ```
 
--- a/rq/job.py
+++ b/rq/job.py
@@ -48,6 +48,7 @@ class Dependency:
         if not all(
             isinstance(job, Job) or isinstance(job, str)
             for job in jobs
+            if job
         ):
             raise ValueError("jobs: must contain objects of type Job and/or strings representing Job ids")
         elif len(jobs) < 1:
--- a/rq/job.py
+++ b/rq/job.py
@@ -13,7 +13,7 @@ from collections.abc import Iterable
 from datetime import datetime, timedelta, timezone
 from enum import Enum
 from functools import partial
-from typing import List
+from typing import List, Union
 from uuid import uuid4
 
 from redis import WatchError
@@ -44,10 +44,13 @@ class JobStatus(str, Enum):
 
 
 class Dependency:
-    def __init__(self, jobs: List[Job], allow_failure: bool = False):
+    def __init__(self, jobs: Union[List[Job], List[str]], allow_failure: bool = False):
         jobs = ensure_list(jobs)
-        if not all(isinstance(job, Job) for job in jobs):
-            raise ValueError("jobs: must contain objects of type Job")
+        if not any([
+            all(isinstance(job, Job) for job in jobs),
+            all(isinstance(job, str) for job in jobs)
+        ]):
+            raise ValueError("jobs: must contain objects of type Job or strings representing Job ids")
         elif len(jobs) < 1:
             raise ValueError("jobs: cannot be empty.")
 
@@ -91,7 +94,7 @@ class Job:
     @classmethod
     def create(cls, func, args=None, kwargs=None, connection=None,
                result_ttl=None, ttl=None, status=None, description=None,
-               depends_on: Dependency = None, timeout=None, id=None, origin=None, meta=None,
+               depends_on=None, timeout=None, id=None, origin=None, meta=None,
                failure_ttl=None, serializer=None, *, on_success=None, on_failure=None):
         """Creates a new Job instance for the given function, arguments, and
         keyword arguments.
@@ -151,7 +154,14 @@ class Job:
 
         # dependency could be job instance or id, or iterable thereof
         if depends_on is not None:
-            job._dependency_ids = [dep.id for dep in depends_on.dependencies]
+            if isinstance(depends_on, Dependency):
+                depends_on_list = depends_on.dependencies
+            else:
+                depends_on_list = ensure_list(depends_on)
+            job._dependency_ids = [
+                dep.id if isinstance(dep, Job) else dep
+                for dep in depends_on_list
+            ]
 
         return job
 
--- a/rq/job.py
+++ b/rq/job.py
@@ -588,8 +588,7 @@ class Job:
         dep_id = obj.get('dependency_id')  # for backwards compatibility
         self._dependency_ids = (json.loads(dep_ids.decode()) if dep_ids
                                 else [dep_id.decode()] if dep_id else [])
-
-        self.allow_failure = bool(obj.get('allow_failure')) if obj.get('allow_failure') else None
+        self.allow_failure = bool(int(obj.get('allow_failure'))) if obj.get('allow_failure') else None
         self.ttl = int(obj.get('ttl')) if obj.get('ttl') else None
         self.meta = self.serializer.loads(obj.get('meta')) if obj.get('meta') else {}
 
@@ -759,7 +758,7 @@ class Job:
                     continue
                 else:
                     # if the pipeline comes from the caller, we re-raise the
-                    # exception as it it the responsibility of the caller to
+                    # exception as it is the responsibility of the caller to
                     # handle it
                     raise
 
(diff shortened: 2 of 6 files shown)


caffeinatedMike:
@selwin Everything seems to be in order according to the logic discussed in [your comment](https://github.com/rq/rq/issues/1608#issuecomment-1013542501) from #1608. However, `test_job_dependency` is failing for the newly-added cases that use `Dependency`. The dependent job seems to be stuck in the `DEFERRED` status.

Could you have a look?

rq/job.py
```diff
@@ -988,6 +1016,7 @@ def dependencies_are_met(self, exclude_job_id=None, pipeline=None):
 
         return all(
             status.decode() == JobStatus.FINISHED
+            or (status.decode() in (JobStatus.FAILED, JobStatus.STOPPED) and self.allow_failure)
```
selwin, line 1019: Can we make in `if` statement here? This is becoming hard to read.

rq/job.py
```diff
@@ -42,6 +42,21 @@ class JobStatus(str, Enum):
...
+        if not all(
+            isinstance(job, Job) or isinstance(job, str)
+            for job in jobs
+        ):
+            raise ValueError("jobs: must contain objects of type Job and/or strings representing Job ids")
+        elif len(jobs) < 1:
+            raise ValueError("jobs: cannot be empty.")
+
+        self.dependencies = jobs
+        self.allow_failure = allow_failure
```
selwin, line 57: Can we make this into an enum to accommodate future variations? Some possible variations:
1. Execute immediately regardless of retries, or wait until retries are done
2. If a job has multiple dependencies, execute immediately after any one is successful3. 

Something like `Dependency(job, execute_when=EXECUTE_WHEN.SUCCESS)` or something like that (but with better naming possibly)?

selwin: A few things that may come into consideration with regards to dependency logic:
1. If it has multiple dependencies
2. Retries

We don't need to address these in this PR, but preferably we design the dependency API so that it's more future proof.

caffeinatedMike:
@selwin can you please address my previous comment before I proceed to make these requested changes?

selwin:
@caffeinatedMike I just looked at the tests results. Many tests are failing so I think those will need to be fixed first. https://github.com/rq/rq/runs/5432544475?check_suite_focus=true

If you can make sure that existing test cases pass, I can then help you debug why the new code paths don't work. The comments I made in [this issue](https://github.com/rq/rq/issues/1608#issuecomment-1013542501) are just general guidelines, there could well be other implementation details that I missed.

caffeinatedMike:
@selwin I'm not sure where to start with resolving these failed tests because it seems many are unrelated to the code changes. From what I saw in other issues there are systemically-failing tests prior to my changes.

selwin:
Your [other pull request](https://github.com/rq/rq/pull/1629/files) seems to have passing tests. Could it be that you're basing your work off other branches that have test failures?

caffeinatedMike:
@selwin Actually it looks like the majority of tests didn't run because python 3.5 tests [failed](https://github.com/rq/rq/runs/5432541797?check_suite_focus=true) (due to use of f-string).

selwin:
Good catch, do you mind removing the usage of f string from your other PR?

Win~

caffeinatedMike:
@selwin I have the Dependency functionality working now with all-but-one test passing (`tests/test_job.py::TestJob::test_create_and_cancel_job_enqueue_dependents_in_registry`). Could you please run the workflows to verify this and assist in figuring out the final failure? Also, I'm not sure what additional changes you want (enum aside).

selwin:
@caffeinatedMike the tests are running. In the meantime, can we try to get https://github.com/rq/rq/pull/1629 merged first?

caffeinatedMike:
@selwin Shoot, I just pushed a commit that fixes the issue that causes that test to fail. As for #1629, unfortunately I'm running into some issues. Since `threading.Timer` is subclass of a `Thread` the exception on timeout occurs outside of the main thread, making it uncatchable (failure test fails even though the exception is thrown). I'm much rather focus on this major feature before circling back to that one.

caffeinatedMike:
@selwin my latest change [011c1f3](https://github.com/rq/rq/pull/1630/commits/011c1f3e1f0bf992566779b49101bec610fc1687) takes care of the last failing test. Please re-run the workflows. 

As for the enum, should this wait since we aren't sure what other variations might be needed? I'm hoping this feature can be merged asap.

caffeinatedMike:
@selwin my latest change [011c1f3](https://github.com/rq/rq/pull/1630/commits/011c1f3e1f0bf992566779b49101bec610fc1687) takes care of the last failing test. Please re-run the workflows. 

As for the enum, should this wait since we aren't sure what other variations might be needed? I'm hoping this feature can be merged asap.

selwin:
> As for the enum, should this wait since we aren't sure what other variations might be needed? I'm hoping this feature can be merged asap.

Ok, let's try to get this one in first.

docs/docs/index.md
```diff
@@ -155,9 +155,33 @@ baz_job = queue.enqueue(baz, depends_on=[foo_job, bar_job])
...
+
+Example:
+
+```python
+from redis import Redis
+from rq.job import Dependency
+from rq import Queue
+
+queue = Queue(connection=Redis())
+job_1 = queue.enqueue(div_by_zero, on_success=report_success, on_failure=report_failure)
```
selwin, line 179: Can we remove the `on_success` and `on_failure` arguments here? They're not relevant in this context.

tests/test_worker.py
```diff
@@ -700,21 +700,58 @@ def test_worker_sets_job_status(self):
...
+        parent_job = q.enqueue(div_by_zero)
+        dependency = Dependency(jobs=parent_job, allow_failure=True)
+        job = q.enqueue_call(say_hello, depends_on=dependency)
+        w.work(burst=True)
+        job = Job.fetch(job.id, connection=self.testconn)
+        self.assertEqual(job.get_status(), JobStatus.FINISHED)
+
+        # only enqueue dependent job when all dependencies have finished/failed
+        first_parent_job = q.enqueue(div_by_zero)
+        second_parent_job = q.enqueue(long_running_job, args=(5,))
```
selwin, line 741: Can we avoid using `long_running_job` for test cases where they're not needed? This will make tests run for longer.

caffeinatedMike: What should this be replaced with here then?

rq/queue.py
```diff
@@ -635,7 +638,9 @@ def enqueue_dependents(self, job, pipeline=None):
                     )
                 ]
 
-                pipe.multi()
+                # only invoke .multi() if all dependent jobs' allow_failure are False-y
+                if not all([job.allow_failure for job in jobs_to_enqueue]):
+                    pipe.multi()
```
selwin, line 643: Sorry, I'm not sure I understand why this is needed.

rq/worker.py
```diff
@@ -974,6 +977,12 @@ def handle_job_failure(self, job, queue, started_job_registry=None,
                                                         job_class=self.job_class, serializer=job.serializer)
                 failed_job_registry.add(job, ttl=job.failure_ttl,
                                         exc_string=exc_string, pipeline=pipeline)
+                try:
+                    pipeline.execute()
+                except Exception:
+                    # Ensure that custom exception handlers are called
+                    # even if Redis is down
+                    pass
```
selwin, line 985: Can we remove this code block? This is not relevant for this PR and I'm not sure calling `pipeline.execute()` here is the right thing to do.

caffeinatedMike: This is required to make `test_job.py::TestJob::test_create_and_cancel_job_enqueue_dependents_in_registry` pass, because originally the job status would be stuck in the `started` state (maybe due to the addition of unconditionally enqueueing dependents below that code block in the same function)

rq/queue.py
caffeinatedMike: Calling `pipe.multi()` here when any of the `jobs_to_enqueue` has `allow_failure=True` will cause `test_worker.py::TestWorker::test_job_dependency` to fail.

caffeinatedMike:
@selwin 

- Addressed (made changes regarding) 2/4 of the latest round of review comments.
- For the other 2, I've responded with clarification on the reasons for the additional logic. Let me know if you need any further clarification or would like to offer an alternative approach.
- Added an additional test to improve coverage.

All checks are now passing 🎉

rq/worker.py
```diff
@@ -974,6 +977,12 @@ def handle_job_failure(self, job, queue, started_job_registry=None,
                                                         job_class=self.job_class, serializer=job.serializer)
                 failed_job_registry.add(job, ttl=job.failure_ttl,
                                         exc_string=exc_string, pipeline=pipeline)
+                try:
+                    pipeline.execute()
```
selwin, line 981: Can we change this to use `with suppress(Exception)`? It will give us better test coverage since the `pass` line will not be tested.

Also, is there a specific exception that covers Redis connection errors? Just so that we don't accidentally swallow other classes of exceptions.

rq/queue.py
selwin: I think this may be caused by the fact that there are [other commands](https://github.com/rq/rq/pull/1630/files#diff-50984f276e8a6fea5f4ba4938afb6f4fbc544fd3751fbd16027dc0a326d6405eR987) in the pipeline that are not yet executed.

I think we can simplify this by ensuring that `enqueue_dependents` are always called in a clean pipeline. Can you call `pipeline.execute()` prior to calling `enqueue_dependents`? This will make the logic easier to follow and troubleshoot.

caffeinatedMike: Sure thing, will make this change and see how it does Monday.

rq/worker.py
caffeinatedMike: Change completed. Logic now suppresses on redis connection errors with `with suppress(redis.exceptions.ConnectionError):`

rq/queue.py
caffeinatedMike: @selwin I attempted this change (added a `pipeline.execute()` immediately before `queue.enqueue_dependents(job, pipeline=pipeline)` and removed the conditional statement), but it didn't not behave as expected (job still ends up as deferred). So far, the only solution was to conditionally execute the `pipe.multi()` command. Thoughts?

selwin: Did you put `pipeline.execute()` before or after `pipeline.watch()`? I think it has to be called before `watch()`. https://github.com/redis/redis-py#pipelines

selwin: The other alternative here is to call `enqueue_dependents` without `pipeline` so it can call `watch()` on its own (since there will always be two round trips to Redis anyway).

caffeinatedMike: > Did you put `pipeline.execute()` before or after `pipeline.watch()`? I think it has to be called before `watch()`. https://github.com/redis/redis-py#pipelines
> The other alternative here is to call enqueue_dependents without pipeline so it can call watch() on its own (since there will always be two round trips to Redis anyway).

I've tried both ways (with removing the conditional statement around the `pipe.multi()` call in `enqueue_dependents`):
```python3
with suppress(redis.exceptions.ConnectionError):
    pipeline.execute()
# if dependencies are inserted after enqueue_dependents
# a WatchError is thrown by execute()
# enqueue_dependents calls multi() on the pipeline!
queue.enqueue_dependents(job, pipeline=pipeline)
```
And
```python3
with suppress(redis.exceptions.ConnectionError):
    pipeline.execute()
# if dependencies are inserted after enqueue_dependents
# a WatchError is thrown by execute()
# enqueue_dependents calls multi() on the pipeline!
queue.enqueue_dependents(job)
```

I think I'm going to have to call it quits on my efforts soon because I've already spent too much time trying to fix the dependency issues with this package. I'm sadly probably going to have to move on to another package.

selwin: @caffeinatedMike no worries, thanks for your efforts!

selwin: I'll try to look into this when I can, am super busy now.

selwin:
Superseded by https://github.com/rq/rq/pull/1681

PR #1650, opened 2022-04-27
[worker.kill_horse] Default to SIGTERM to exit worker gracefully, send SIGKILL after timeout.

luisbc92:
I have a use case where I need to terminate a currently running job with `send_stop_job_command(redis, job_id)`, but I want to give the worker the opportunity to exit cleanly.

Currently, this is not possible as this command will trigger `Worker.kill_horse` which defaults to sig=SIGKILL, which immediately kills the worker. We could fix it by calling `Worker.request_stop`, however, we still want to forcefully kill the worker if it does not respond.

As implemented, my solution extends `Worker.kill_horse` to attempt to kill the worker by sending the requested signal first. If the worker does not exit within a timeout, a SIGKILL signal is sent to force termination.

### DIFF
--- a/rq/timeouts.py
+++ b/rq/timeouts.py
@@ -70,7 +70,7 @@ class UnixSignalDeathPenalty(BaseDeathPenalty):
         an exception after the timeout amount (expressed in seconds).
         """
         signal.signal(signal.SIGALRM, self.handle_death_penalty)
-        signal.alarm(self._timeout)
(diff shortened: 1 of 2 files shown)


selwin:
You can use `send_shutdown_command()` for this. https://github.com/rq/rq/blob/b80045d615d539cf4c565c5541694efe13ea648e/rq/command.py#L25

PR #1657, opened 2022-05-20
Support at_front requeue support

buroa:
This allows people to specify if they want a requeue to place at the front (which at_front already has support).


buroa:
@selwin One for you :)

selwin:
Thanks, will merge if CI passes

Issue #1646, opened 2022-04-14
use unittest.mock instead of mock


pgajdos:
Could you please consider to utilize unittest.mock, as a fallback at least?

selwin:
PR for this is welcome :)

pgajdos:
Thank you!
And sorry I get not it sooner.

PR #1673, opened 2022-07-11
Use unittest.mock instead of mock

WhyNotHugo:
This module has been included in Python itself since 3.3.

Fixes: https://github.com/rq/rq/issues/1646


selwin:
Thanks!

PR #1686, opened 2022-08-02
Support callbacks in enqueue_many

### LINKED ISSUE #1680: Bulk Enqueue Documentation Inconsistent with API
In the documentation for [Bulk Job Enqueuing](https://python-rq.org/docs/#bulk-job-enqueueing), it states that

> Queue.prepare_data accepts all arguments that Queue.parse_args does EXCEPT for depends_on, which is not supported at this time, so dependencies will be up to you to setup.

This does not appear to be the case as `Queue.parse_args` accepts arguments for `on_success` and `on_failure` callbacks added in 1.9.0 but `Queue.prepare_data` does not.

https://github.com/rq/rq/blob/145884fcd5a016ee353da9ce1a586044bf058f45/rq/queue.py#L413-L425

It would be nice if this could get added (and subsequently plumbed through to `enqueue_many`) so that bulk job creation and single job creation have feature parity.

y4n9squared:
The `on_success` and `on_failure` callbacks are currently supported in `Queue.enqueue` but not `Queue.enqueue_many`, despite documentary indicating the contrary.

This commit brings the bulk enqueuing API back into parity.

Fixes #1680

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -23,7 +23,7 @@ def compact(lst):
 class EnqueueData(namedtuple('EnqueueData', ["func", "args", "kwargs", "timeout",
                                              "result_ttl", "ttl", "failure_ttl",
                                              "description", "job_id",
-                                             "at_front", "meta", "retry"])):
+                                             "at_front", "meta", "retry", "on_success", "on_failure"])):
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

Issue #1690, opened 2022-08-05
Is there a way to make sure dependants are automatically queued at the front of the queue?


jtfidje:
Currently I have a custom Queue class that implements this feature by overriding the `enqueue_dependents` method, but after upgrading to 1.11 stuff breaks as a consequence of my overrides.

Is there a way to achieve this functionality without custom code?

selwin:
I'm happy to accept a PR for this. The newly introduced `Dependency` class gives us a good way of controlling how dependencies are handled. https://github.com/rq/rq/blob/5b95725dc4410634a658bda9aa59188eaeaf87ea/rq/job.py#L41

The API should be as follows:
```python
queue.enqueue(foo, depends_on=Dependency(bar, enqueue_at_front=True))
```

Mind opening a PR for this?

jtfidje:
Sure!

lowercase00:
Implemented with #1696. Closing.

Issue #1693, opened 2022-08-05
Dependent Job Queueing Order Issue


jpbakke:
RQ uses redis SETS store the dependent jobs in the deferred registry.  When the set of dependents is retrieved, the SMEMBER function in redis is used.  Unfortunately, this set function does not guarantee the order of dependents.  I have situation where I want the dependents to be queued in a specific order after their dependency completes and currently this order is non-deterministic due redis using hash storage for sets.  

I would request that the dependent list be stored in redis using ZADD and ZRANGE to retrieve to maintain queuing order for dependents.

selwin:
I'm ok with this change. Do you mind opening a PR for this?

jpbakke:
Yes, I can make the changes and test in a few days and generate a PR for it.

selwin:
Remember to take backwards compatibility into account. If there are dependencies, you'll need to first check the sorted set before falling back to the older set structure.

PR #1696, opened 2022-08-09
Add feature to enqueue dependents at the front of queues

jtfidje:
Suggested solution to #1690.

### DIFF
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -167,8 +167,7 @@ The `Dependency(jobs=...)` parameter accepts:
 - a string representing a single job id
 - a Job object
 - an iteratable of job id strings and/or Job objects
-
-In addition, the `Dependency` class can be configured to enqueue the dependents at the front of the queue.
+- `enqueue_at_front` boolean parameter to put dependents at the front when they are enqueued
 
 Example:
 
@@ -279,10 +278,10 @@ There are two options:
 
 #### Arguments:
 
-| | plain text | json | [literal-eval](https://docs.python.org/3/library/ast.html#ast.literal_eval) |
-|-|-|-|-|
-| keyword | `[key]=[value]` | `[key]:=[value]` | `[key]%=[value]` |
-| no keyword | `[value]` | `:[value]` | `%[value]` |
+|            | plain text      | json             | [literal-eval](https://docs.python.org/3/library/ast.html#ast.literal_eval) |
+| ---------- | --------------- | ---------------- | --------------------------------------------------------------------------- |
+| keyword    | `[key]=[value]` | `[key]:=[value]` | `[key]%=[value]`                                                            |
+| no keyword | `[value]`       | `:[value]`       | `%[value]`                                                                  |
 
 Where `[key]` is the keyword and `[value]` is the value which is parsed with the corresponding
 parsing method.
--- a/docs/docs/index.md
+++ b/docs/docs/index.md
@@ -168,6 +168,8 @@ The `Dependency(jobs=...)` parameter accepts:
 - a Job object
 - an iteratable of job id strings and/or Job objects
 
+In addition, the `Dependency` class can be configured to enqueue the dependents at the front of the queue.
+
 Example:
 
 ```python
@@ -177,9 +179,17 @@ from rq import Queue
 
 queue = Queue(connection=Redis())
 job_1 = queue.enqueue(div_by_zero)
-dependency = Dependency(jobs=[job_1], allow_failure=True)  # allow_failure defaults to False
+dependency = Dependency(
+    jobs=[job_1],
+    allow_failure=True,    # allow_failure defaults to False
+    enqueue_at_front=True  # enqueue_at_front defaults to False  
+)
 job_2 = queue.enqueue(say_hello, depends_on=dependency)
-# job_2 will execute even though its dependency (job_1) fails
+
+"""
+  job_2 will execute even though its dependency (job_1) fails,
+  and it will be enqueued at the front of the queue.
+"""
 ```
 
 
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -648,16 +648,22 @@ class Queue:
                     break
 
                 for dependent in jobs_to_enqueue:
+                    if dependent.enqueue_at_front is not None:
+                        enqueue_at_front = dependent.enqueue_at_front
+                    else:
+                        enqueue_at_front = False
+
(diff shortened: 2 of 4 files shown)


selwin:
@jtfidje looks good. Do you mind updating the docs to add `enqueue_at_front` parameter to `Dependency` class? Also please make sure all the tests pass.

jtfidje:
Aah yes - forgot docs! 

All tests passed when running `run_tests_in_docker.sh` - is there a way to run the CI tests locally as well?

selwin:
@jtfidje can you check your tests and ensure that they pass? I'll merge if they can pass the tests.

docs/docs/index.md
```diff
@@ -168,6 +168,8 @@ The `Dependency(jobs=...)` parameter accepts:
 - a Job object
 - an iteratable of job id strings and/or Job objects
```
selwin, line 170: Can you add this here (in the list of parameters)?
```
`enqueue_at_front` boolean parameter to put dependents at the front when they are enqueued
```

jtfidje:
Looks like this PR can/will also benefit from #1693 - maybe we should wait and see?

selwin:
I'd rather get this PR merged in first. We don't know when https://github.com/rq/rq/issues/1693 will be done.

If you can make all the tests pass, I'll merge this in and release a patch release along with a few bug fixes and improvements sometime in the next week or two.

jtfidje:
No problem :) I'm facing a problem in my test tho: It seems that I cannot trust that the list of queued jobs I retrieve from Redis is ordered correctly. Even though I queue my `job_2` with `at_front=True`, the list I get back does not always have `job_2` at index 0. 
I've also tried to change my test so that I call `.work(...)` two times, one for parent and then once more to get the next job, but this has the same issue. Sometimes my `job_2` which is supposed to be at the front of the queue gets pulled and finished, but sometimes `job_1` gets pulled instead.

selwin:
Ah ok, if that’s the problem, there’s a simple fix for that.

After you create the parent job along with the dependents, create another regular job so the queue is not empty.

Afterwards, run worker.work(max_jobs=1). This way, the one dependent will always be enqueued to the end of the queue and the other one always to the front.

This way we don’t depend on the order at which the dependents get enqueued.

jtfidje:
Thanks for your time and quick responses!

I think that I don't quite follow what you mean, or maybe I've written a bad test. If we have the following setup:
```python
q.empty()
parent_job = q.enqueue(say_hello)
job_1 = q.enqueue(
    say_hello,
    job_id='fake_job_id_1',
    depends_on=Dependency(jobs=[parent_job])
)
job_2 = q.enqueue(
    say_hello,
    job_id='fake_job_id_2',
    depends_on=Dependency(jobs=[parent_job],enqueue_at_front=True)
)
w.work(burst=True, max_jobs=1)
```

I need to reliably test that `job_2` will run before `job_1`. I did this with the following assertion:

```python
self.assertEqual(q.job_ids, ["fake_job_id_2", "fake_job_id_1"])
```

How should I do this instead?

selwin:
Try doing this:
```python
parent_job = q.enqueue(say_hello)
job_1 = q.enqueue(
    say_hello,
    job_id='fake_job_id_1',
    depends_on=Dependency(jobs=[parent_job])
)
job_2 = q.enqueue(
    say_hello,
    job_id='fake_job_id_2',
    depends_on=Dependency(jobs=[parent_job],enqueue_at_front=True)
)

q.enqueue(say_hello) # This is a filler job that will act as a separator for jobs, one will be enqueued at front while the other one at the end of the queue
w.work(burst=True, max_jobs=1)
```

jtfidje:
I've added your suggestion and run the test three times. This is the result from `q.job_ids` in order:

```
['1bf9a09a-9b8c-4c31-9737-b61b970645bc', 'fake_job_id_1', 'fake_job_id_2']
['cd9bab30-9e45-4f03-b234-3c1ea080eed8', 'fake_job_id_2', 'fake_job_id_1']
['a2669abd-3eb7-42f5-b805-a122b88f11e9', 'fake_job_id_1', 'fake_job_id_2']
```

It seems the order is still random after `w.work(burst=True, max_jobs=1)` has been executed. I'm at a loss for how I could test if `at_front=True` is actually performed.

selwin:
Then it seems like enqueue_at_front did not work. Did you check the logic and whether the enqueue_at_front variable was saved properly?

selwin:
To recap, it seems like the issue is not in the random ordering but the job was never enqueued at front.

jtfidje:
I'll dive back into the code again then :) Again, thank you for your time :)

jtfidje:
You were correct - I had overlooked something critical. It now looks like everything is ok :) Please double check tho 😅

selwin:
Sorry I missed your message. Will check in a bit.

selwin:
`job.meta` is actually used for storing non core RQ data. Storing `enqueue_at_front` attribute on the actual job hash is actually the correct way.

If `job.enqueue_at_front` was not persisted correctly, you should look at:
1. `job.to_dict()` - serializes job attributes to be saved in Redis. Tackle this first so `enqueue_at_front` attribute gets correctly saved. https://github.com/rq/rq/blob/master/rq/job.py#L616
2. `job.restore()` - populates job attributes based on Redis hash. After it's saved properly, use `Job.fetch()` to test that `enqueue_at_front` gets properly restored. https://github.com/rq/rq/blob/master/rq/job.py#L545

jtfidje:
Oh I understand! I'll fix it right away :)

rq/queue.py
```diff
@@ -325,9 +325,6 @@ def create_job(self, func, args=None, kwargs=None, timeout=None,
             job.retries_left = retry.max
             job.retry_intervals = retry.intervals
 
-        if isinstance(depends_on, Dependency):
```
jtfidje, line 328: I took the liberty and removed these lines as this check is already done inside job.create.

jtfidje:
Hi! I apologize for taking so long to fix this. Here is an updated proposal :)

I also took the liberty to remove some duplicate code - see: https://github.com/rq/rq/pull/1696/files#r954568688

selwin:
Hey sorry I totally missed this update to your PR. This looks good, thanks!

selwin:
Now I'm just re-running all the tests and if they pass, I can merge this in.

jtfidje:
No problem! I should've been more active myself - sorry :)
Thanks for merging!

PR #1698, opened 2022-08-15
Typing

lowercase00:
With 3.5 EOL should be safe to add type hints without much side effects.
Haven't done everything, but it's already helpful with autocomplete and some useful Mypy warnings.
There are a few arguments that are very flexible (accept several different types) and others that I'm not particularly familiar with, but this should already help.


selwin:
Thanks, can you make sure that all tests pass so I can merge this in?

lowercase00:
@selwin sure. I used `list` as a type that only works on `3.9` forward if I'm not mistaken. I'll update the code.
I guess I could also remove 3.5 from Github Actions and tox, right? Tests are passing locally with docker, but there seems to be a problem with the `test_command`, for some reason can't pickle the job. I'll check what changes might be causing this.

lowercase00:
@selwin I made a few changes with this last commit, nothing big, but would be great if you could take a look:

- Fixed the usage of typing (3.9+ accepts `list` as such, changed everything back to `typing.List`)
- Removed `annotations` import from `__future__`, since this is python3.8+ 
- removed a Py2 compatibility from tests
- Changed the Tests Dockerfile to run Tox with all environments.
- Explicitly added python version to `tox` file, it seems some versions fail when using the shortcut `py36` for example
- Removed 3.5 from github actions flow declaration (this version won't work with 3.5 anyway)
- Moves the script `run_tests_in_docker` to the Makefile (`make test`)
- Wrote a couple of docstrings (Using Google style)

I guess we could also remove everything related to Py2 compatibility (there's code related to compat in a few places, if you give me green light I'll clean this up). Actually the whole `compat` module I think.

The tests are a bit slower when using `tox` since they run in every environment (from 3.6 to 3.10), but a lot more complete than they were. Everything passing (using Docker & tox). Still need to check GH's automated tests. I can't seem to make the `test_command` pass in my mac (macOS 11.6), but I can't make them pass in the master branch either. Not sure if it's something platform specific that could justify a `skip` if `darwin`

lowercase00:
@selwin had a typo and a couple of tests failed, fixed it in the last commit, should be ok now. Just not sure how you guys want to keep the Github actions (Pytest directly vs tox), let me know

lowercase00:
@selwin had the chance to take a look at this? Any other adjustments needed?

cyphase:
Just came looking for anything about typing for RQ. I'd love to see this merged. Thanks @lowercase00!

selwin:
@lowercase00 this looks good, do you mind fixing the merge conflict so I can merge this in?

Additionally I'm also wondering whether we can setup CI such that additional codes without proper type hints are flagged?

lowercase00:
@selwin hmmm I'm not really sure, even though I think it would be a great way to reinforce typed code in the future.

There are a few options I can think of:
- Adding MyPy checks to GH Actions (https://github.com/marketplace/actions/mypy-action)
- Have a pre-commit hook that checks for types (https://github.com/pre-commit/mirrors-mypy)
- Add a new tool to only lint diff (like https://github.com/akaihola/darker)

There's still a lot of small changes to make: MyPy still points to 84 problems, I tried to fix all low hanging fruits, but there are somethings that we would have to take a closer look on how to fix.

Some ideas: I would probably merge this asap to avoid having it lagging behind the master every now and then. Have a new branch "types" that closely follows the master and that goes for the 100% type coverage, after that's working add MyPy to GH actions and the pre-commit hook. This might take a few iterations (there's not much left), but would make the repo 100% typed before long, and with additional tools being enforced in CI.

selwin:
I'm re-running CI, once all the tests pass, I'll merge this PR in.

PR #1704, opened 2022-09-06
doc: default queues order is a priority order

alexgarel:
In introduction, we say that queue order matters, but we are not very specific.
Added a paragraph to explicit that.

### DIFF
--- a/docs/docs/workers.md
+++ b/docs/docs/workers.md
@@ -319,7 +319,7 @@ $ rq worker -w 'path.to.GeventWorker'
 
 ## Round Robin and Random strategies for dequeuing jobs from queues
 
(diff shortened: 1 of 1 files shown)


docs/docs/workers.md
```diff
@@ -319,6 +319,10 @@ $ rq worker -w 'path.to.GeventWorker'
 
 ## Round Robin and Random strategies for dequeuing jobs from queues
 
+The default worker consider the order of queues as their priority order,
```
alexgarel, line 322:
```suggestion
The default worker considers the order of queues as their priority order, 
```

selwin:
Thanks!

PR #1709, opened 2022-09-20
Fix log on job stop

ronlut:
Job status is not yet "stopped" when determining the log to print in `monitor_work_horse`


selwin:
Thanks!

PR #1710, opened 2022-09-22
scheduler heartbeat for acquired locks only

### LINKED ISSUE #1707: Scheduler heartbeats for queues it doesn't have locks on
When managing multiple queues, the `RQScheduler` can heartbeat for queues it does not have a lock on. This prevents other workers scheduler's from acquiring locks and starting a new scheduler in the event of a schedulers death. 
https://github.com/rq/rq/blob/v1.11/rq/scheduler.py#L181
```python
    def heartbeat(self):
        """Updates the TTL on scheduler keys and the locks"""
        self.log.debug("Scheduler sending heartbeat to %s",
                       ", ".join(self.acquired_locks))
        if len(self._queue_names) > 1:
                for name in self._queue_names:
                    key = self.get_locking_key(name)
                    pipeline.expire(key, self.interval + 60)
                pipeline.execute()
        else:
            key = self.get_locking_key(next(iter(self._queue_names)))
            self.connection.expire(key, self.interval + 60)
```
`for name in self._queue_names` should be  `for name in self._acquired_locks`.

milesjwinter:
closes #1707

### DIFF
--- a/rq/scheduler.py
+++ b/rq/scheduler.py
@@ -184,7 +184,7 @@ class RQScheduler:
                        ", ".join(self.acquired_locks))
         if len(self._queue_names) > 1:
             with self.connection.pipeline() as pipeline:
-                for name in self._queue_names:
+                for name in self._acquired_locks:
                     key = self.get_locking_key(name)
                     pipeline.expire(key, self.interval + 60)
                 pipeline.execute()


milesjwinter:
ran `./run_tests_in_docker.sh` locally, all test passing.

selwin:
Thanks!

PR #1727, opened 2022-11-09
pre-warm redis server version cache of job from queue

rggjan:
`rq.queue.Queue.enqueue_many` is used to enqueue many jobs in one go on `rq` in a bulk operation. 

The issue is that it calls `rq.queue.Queue.enqueue_job` which in turn calls `rq.job.Job.save` which calls `rq.job.Job.get_redis_server_version` which does a round-trip to the server to get the server version. This is cached in `Job`, however when enqueuing 1000 jobs all of these caches are empty. So calling `enqueue_job` on 1000 jobs will do 1000 round-trips to the server to get the server version, which is very slow. 

The patch pre-warms the `Job` cache, causing the round-trip to the server to only happen once.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -562,6 +562,7 @@ class Queue:
 
         # Add Queue key set
         pipe.sadd(self.redis_queues_keys, self.key)
+        job.redis_server_version = self.get_redis_server_version()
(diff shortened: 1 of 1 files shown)


rggjan:
@aparcar maybe you can have a look, since you added the server version check logic back in 2020?

PR #1729, opened 2022-11-13
Enhance worker termination logic

ronlut:
* use `wait4` instead of `waitpid` to get `rusage`. This can be used by handlers to detect OOM for example (our use-case)
* allow custom handling of workhorse terminated situation
* extract signal on termination (if any)


rq/worker.py
```diff
@@ -175,7 +176,8 @@ def __init__(self, queues, name: t.Optional[str] = None, default_result_ttl=DEFA
                  queue_class=None, log_job_description: bool = True,
                  job_monitoring_interval=DEFAULT_JOB_MONITORING_INTERVAL,
                  disable_default_exception_handler: bool = False,
-                 prepare_for_work: bool = True, serializer=None):  # noqa
+                 prepare_for_work: bool = True, serializer=None,
+                 workhorse_terminated_handler=None):  # noqa
```
selwin, line 180: Mind updating the docs for this new argument?

ronlut: @selwin sure, where is it documented? I didn't find any place documenting arguments to the constructor

selwin: Thanks for updating this PR. Mind adding some documentation for this handler in the exceptions section?

This is how it looks like on the web. https://python-rq.org/docs/exceptions/#custom-exception-handlers

selwin:
@ronlut still interested in getting this PR merged?

ronlut:
I do, will get to the documentation change soon :)

ronlut:
done

docs/docs/exceptions.md
```diff
@@ -119,3 +119,27 @@ use a custom exception handler that doesn't fall through, for example:
 def black_hole(job, *exc_info):
     return False
 ```
+
+## Workhorse Termination Handler
```
selwin, line 123: Can you add the text `_New in version 1.13.0_`? An example is here: https://github.com/rq/rq/blob/a02ad29cef71712376aa0b284b4cb2e3aea576fd/docs/docs/index.md?plain=1#L239

rq/worker.py
```diff
@@ -221,6 +223,7 @@ def __init__(
         disable_default_exception_handler: bool = False,
         prepare_for_work: bool = True,
         serializer=None,
+        workhorse_terminated_handler: Optional[Callable[[Job, int, int, resource.struct_rusage], None]] = None
```
selwin, line 226: Any ideas for alternative arg names? Maybe `forced_termination_handler`?

`workhorse_terminated_handler` does not imply that this callable is to handle cases where the workhorse is forcibly terminated.

ronlut: By forcibly you mean something stopped it from outside (hard/stop)? Just to clarify - We are handling a case in which the workhorse was terminated unexpectedly, while running the job. The case for stopped is handled [differently here](https://github.com/rq/rq/pull/1729/files#diff-50984f276e8a6fea5f4ba4938afb6f4fbc544fd3751fbd16027dc0a326d6405eR949).
LMKWYT

ronlut: How do you like `workerhorse_crash_handler`? See latest commit changes

selwin: `workerhorse_crash_handler` is ok, but I'd revise is a bit to `horse_crash_handler`. What do you think about `horse_killed_handler` or `process_killed_handler`?

The term kill is quite specific to a process being killed.

ronlut: For me killed is intuitively connected to a process being killed from the outside, like when stopping or force stopping

selwin: Crash could also mean an exception raised from inside the job though. Let's go with `work_horse_killed_handler`, I think that's the most descriptive term (and already reinforced well in the docs), we already have `send_stop_job_command` so users shouldn't mistake it for job being stopped.

ronlut:
A test fails here so let me review it tomorrow

ronlut:
@selwin ready for review

selwin:
Thanks!

Issue #1733, opened 2022-11-18
Make rq website more accessible


krzysztofjeziorny:
Please make the rq website more acccessible - setting body text in Lato light weight makes it look like set in light gray on white, making it completely [unreadable](https://color.a11y.com/ContrastPair/?bgcolor=ffffff&fgcolor=B1B1B1). 

Just turning off `font-weight: 300` makes it way easier to the eyes.

Thank you!

selwin:
Thanks for the feedback. Mind opening a PR for this? :)

krzysztofjeziorny:
@selwin Sure thing :) https://github.com/rq/rq/pull/1734 

The problem is I don't know how to compile the docs, so I if my changes break something please ignore or adjust if needed.

PR #1734, opened 2022-11-21
#1733 Typography adjustments for docs

krzysztofjeziorny:
- html font size removed, body font size set in rem w/o weight for better readability 
- font stacks with system fonts defined - Lato font wasn't even imported, so it wouldn't load for most users
- code formatted w/ prettier rules


selwin:
Thanks!

Issue #1595, opened 2021-11-19
Pass dot-separated string of package, module and function to success / failure callbacks


erikrichardlarson:
Could we allow the success / failure callbacks to accept a dot-separated string in addition to building it like below? https://github.com/rq/rq/blob/76ba690aafaa0fffbab1ee8f5db7c27c5327d792/rq/job.py#L128

Happy to create a PR if it makes sense.

selwin:
Yes, would be happy to accept a PR for this. Please include tests also :)

PR #1738, opened 2022-11-30
Support string callbacks in on_success/on_failure

ankush:
`on_success` and `on_failure` callbacks can now take dotted function string in addition to function objects.

Internally this isn't quite different as during serialization phase functions seem to be converted to their dotted path string representation anyway.

Note:
- I've tested functionality locally (and with our projects that use rq)
- Added test and ran tests locally. 

closes https://github.com/rq/rq/issues/1595

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -138,14 +138,20 @@ class Job:
         job._kwargs = kwargs
 
         if on_success:
-            if not inspect.isfunction(on_success) and not inspect.isbuiltin(on_success):
-                raise ValueError('on_success callback must be a function')
-            job._success_callback_name = '{0}.{1}'.format(on_success.__module__, on_success.__qualname__)
+            if inspect.isfunction(on_success) or inspect.isbuiltin(on_success):
+                job._success_callback_name = '{0}.{1}'.format(on_success.__module__, on_success.__qualname__)
+            elif isinstance(on_success, string_types):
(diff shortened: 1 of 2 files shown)


rq/job.py
```diff
@@ -138,14 +138,20 @@ def create(cls, func: t.Callable[..., t.Any], args=None, kwargs=None, connection
...
+            elif isinstance(on_success, string_types):
+                job._success_callback_name = as_text(on_success)
+            else:
+                raise ValueError('on_success callback must be a function or a string but got: {0}'.format(on_failure))
 
         if on_failure:
-            if not inspect.isfunction(on_failure) and not inspect.isbuiltin(on_failure):
-                raise ValueError('on_failure callback must be a function')
-            job._failure_callback_name = '{0}.{1}'.format(on_failure.__module__, on_failure.__qualname__)
+            if inspect.isfunction(on_failure) or inspect.isbuiltin(on_failure):
```
ankush, line 149: Note: I thought of writing a utils to avoid duplication here and in job which uses fairly similar logic but there are not that many common parts and utils will probably give bad error messages. 

Open to suggestions.

selwin: Mind refactoring the callable inspection part into a separate function that raises a type of exception? We can then catch this exception and reraise it with a descriptive error message.

PR #1742, opened 2022-12-19
Cache redis version in the connection object

odarbelaeze:
Hi there, long time `rq` user here.

I was recently running some annual disaster recovery tests in our infra. These tests involve enqueuing hundreds of thousands of objects in our `rq` and then running scripts to evict duplicated jobs. To my surprise, enqueuing the objects would take about a day or so, while last years test took a few seconds.

Upon further inspection and profiling it became aparent that the reason for the low performance was a `get_redis_server_version` call for every job. And caching or hard-coding that value would help the performance issue.

Here's a proposal to sneakily store the server version on the connection object, this would make enqueue calls fast again as long as they are made from the same connection.

### DIFF
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -303,7 +303,14 @@ def get_version(connection: 'Redis'):
         connection (Redis): The Redis connection.
     """
     try:
-        return tuple(int(i) for i in connection.info("server")["redis_version"].split('.')[:3])
+        # Getting the connection info for each job tanks performance, we can cache it on the connection object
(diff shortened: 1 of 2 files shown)


selwin:
This is an interesting PR. Have you considered storing this information in thread local instead?

odarbelaeze:
I think you could have multiple connections in one thread. However, the PR assumes you're not wildly changing the redis target version reusing the same connection.

selwin:
Thanks!

PR #1752, opened 2023-01-16
Enhance debug logs

lowercase00:
There were reports where a worker wouldn't fetch new jobs.
Since one possible cause is the infinite hanging of the `BLPOP` command,
this adds a couple of logs surrounding the `dequeue_any` function,
that interacts with Redis using `BLPOP`,

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -79,7 +79,7 @@ class Queue:
         return cls(name, connection=connection, job_class=job_class, serializer=serializer)
 
     def __init__(self, name='default', default_timeout=None, connection: t.Optional['Redis'] = None,
-                 is_async=True, job_class=None, serializer=None, **kwargs):
+                 is_async=True, job_class=None, serializer=None, debug=False, **kwargs):
         self.connection = resolve_connection(connection)
         prefix = self.redis_queue_namespace_prefix
         self.name = name
@@ -87,6 +87,7 @@ class Queue:
         self._default_timeout = parse_timeout(default_timeout) or self.DEFAULT_TIMEOUT
         self._is_async = is_async
         self.log = logger
+        self.debug_mode = debug
 
         if 'async' in kwargs:
             self._is_async = kwargs['async']
@@ -305,13 +306,15 @@ class Queue:
     def push_job_id(self, job_id: str, pipeline: t.Optional['Pipeline'] = None, at_front=False):
         """Pushes a job ID on the corresponding Redis queue.
         'at_front' allows you to push the job onto the front instead of the back of the queue"""
-        self.log.debug(f"Pushing job {blue(job_id)} into the queue {green(self.name)}, current job count {self.count}")
+        self.log.debug(f"Pushing job {blue(job_id)} into the queue {green(self.name)}.")
+        if self.debug_mode:
+            self.log.debug(f"Current job count for queue {green(self.name)} is {self.count}.")
         connection = pipeline if pipeline is not None else self.connection
         if at_front:
             result = connection.lpush(self.key, job_id)
         else:
             result = connection.rpush(self.key, job_id)
-        self.log.debug(f"Pushed job {blue(job_id)} into the queue {green(self.name)}, total job count after pushing job: {result}")
+        self.log.debug(f"Pushed job {blue(job_id)} into {green(self.name)}, {result} job(s) are in queue.")
 
     def create_job(self, func: t.Callable[..., t.Any], args=None, kwargs=None, timeout=None,
                    result_ttl=None, ttl=None, failure_ttl=None,
@@ -698,7 +701,7 @@ class Queue:
         return as_text(self.connection.lpop(self.key))
 
     @classmethod
-    def lpop(cls, queue_keys, timeout, connection: t.Optional['Redis'] = None):
+    def lpop(cls, queue_keys, timeout: int, connection: t.Optional['Redis'] = None):
         """Helper method.  Intermediate method to abstract away from some
         Redis API details, where LPOP accepts only a single key, whereas BLPOP
         accepts multiple.  So if we want the non-blocking LPOP, we need to
@@ -715,7 +718,7 @@ class Queue:
         if timeout is not None:  # blocking variant
             if timeout == 0:
                 raise ValueError('RQ does not support indefinite timeouts. Please pick a timeout value > 0')
-            logger.debug(f"Starting blocking pop operation for queues {green(queue_keys)} with timeout of {timeout}")
+            logger.debug(f"Starting BLPOP operation for queues {green(queue_keys)} with timeout of {timeout}")
             result = connection.blpop(queue_keys, timeout)
             if result is None:
                 logger.debug(f"BLPOP Timeout, no jobs found on queues {green(queue_keys)}")
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -720,7 +720,7 @@ class Queue:
                 logger.debug(f"BLPOP Timeout, no jobs found on queues {green(queue_keys)}")
                 raise DequeueTimeout(timeout, queue_keys)
             queue_key, job_id = result
-            logger.debug(f"Found job {blue(job_id)} on queue {green(queue_key)}")
+            logger.debug(f"Dequeued job {blue(job_id)} from queue {green(queue_key)}")
             return queue_key, job_id
         else:  # non-blocking variant
             for queue_key in queue_keys:
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -79,7 +79,7 @@ class Queue:
         return cls(name, connection=connection, job_class=job_class, serializer=serializer)
 
     def __init__(self, name='default', default_timeout=None, connection: t.Optional['Redis'] = None,
-                 is_async=True, job_class=None, serializer=None, debug=False, **kwargs):
+                 is_async=True, job_class=None, serializer=None, **kwargs):
         self.connection = resolve_connection(connection)
         prefix = self.redis_queue_namespace_prefix
         self.name = name
@@ -87,7 +87,6 @@ class Queue:
         self._default_timeout = parse_timeout(default_timeout) or self.DEFAULT_TIMEOUT
         self._is_async = is_async
         self.log = logger
-        self.debug_mode = debug
 
         if 'async' in kwargs:
             self._is_async = kwargs['async']
@@ -306,9 +305,6 @@ class Queue:
     def push_job_id(self, job_id: str, pipeline: t.Optional['Pipeline'] = None, at_front=False):
         """Pushes a job ID on the corresponding Redis queue.
         'at_front' allows you to push the job onto the front instead of the back of the queue"""
-        self.log.debug(f"Pushing job {blue(job_id)} into the queue {green(self.name)}.")
-        if self.debug_mode:
-            self.log.debug(f"Current job count for queue {green(self.name)} is {self.count}.")
         connection = pipeline if pipeline is not None else self.connection
         if at_front:
             result = connection.lpush(self.key, job_id)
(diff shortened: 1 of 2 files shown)


rq/worker.py
```diff
@@ -690,10 +690,12 @@ def dequeue_job_and_maintain_ttl(self, timeout):
                 if self.should_run_maintenance_tasks:
                     self.run_maintenance_tasks()
 
+                self.log.debug(f"Dequeing jobs on queues {self._ordered_queues} and timeout {timeout}")
```
selwin, line 693: Sorry, minor nitpick `Dequeueing job from self._ordered_queues with timeout of {}`

rq/worker.py
```diff
@@ -690,10 +690,12 @@ def dequeue_job_and_maintain_ttl(self, timeout):
...
                 result = self.queue_class.dequeue_any(self._ordered_queues, timeout,
                                                       connection=self.connection,
                                                       job_class=self.job_class,
                                                       serializer=self.serializer)
+                self.log.debug(f"Dequeued job ID {result[1]} on queue {result[0]}")
```
selwin, line 698: `Dequeued job {result[1]} from {result[0]}`

rq/worker.py
```diff
@@ -1098,17 +1105,23 @@ def perform_job(self, job: 'Job', queue: 'Queue'):
         """Performs the actual work of a job.  Will/should only be called
         inside the work horse's process.
         """
+        self.log.debug("Starting process to perform a job...")
         push_connection(self.connection)
+        self.log.debug("Connection pushed.")
```
selwin, line 1110: I think we can get rid of these two statements.

rq/worker.py
```diff
@@ -1084,12 +1089,14 @@ def handle_job_success(self, job: 'Job', queue: 'Queue', started_job_registry: S
 
     def execute_success_callback(self, job: 'Job', result):
         """Executes success_callback with timeout"""
+        self.log.debug("Running Success Callbacks.")
```
selwin, line 1092: `Running success callbacks for {job.id}`

rq/worker.py
```diff
@@ -1084,12 +1089,14 @@ def handle_job_success(self, job: 'Job', queue: 'Queue', started_job_registry: S
...
         job.heartbeat(utcnow(), CALLBACK_TIMEOUT)
         with self.death_penalty_class(CALLBACK_TIMEOUT, JobTimeoutException, job_id=job.id):
             job.success_callback(job, self.connection, result)
 
     def execute_failure_callback(self, job):
         """Executes failure_callback with timeout"""
+        self.log.debug("Running Failure Callbacks.")
```
selwin, line 1099: `Running Failure Callbacks for {job.id}`

rq/worker.py
```diff
@@ -1098,17 +1105,23 @@ def perform_job(self, job: 'Job', queue: 'Queue'):
...
         """
+        self.log.debug("Starting process to perform a job...")
         push_connection(self.connection)
+        self.log.debug("Connection pushed.")
 
         started_job_registry = queue.started_job_registry
+        self.log.debug("Started Job Registry set.")
 
         try:
+            self.log.debug("Preparing for execution...")
```
selwin, line 1116: I think we can remove this statement, there's already a logging statement in `prepare_job_execution`

rq/worker.py
```diff
@@ -1123,6 +1136,7 @@ def perform_job(self, job: 'Job', queue: 'Queue'):
                                     queue=queue,
                                     started_job_registry=started_job_registry)
         except:  # NOQA
+            self.log.debug("Job Execution failed.")
```
selwin, line 1139: `Job <ID> raised an exception`

rq/worker.py
```diff
@@ -1140,6 +1154,7 @@ def perform_job(self, job: 'Job', queue: 'Queue'):
 
             self.handle_job_failure(job=job, exc_string=exc_string, queue=queue,
                                     started_job_registry=started_job_registry)
+            self.log.debug("Handling Exception.")
```
selwin, line 1157: Can we move this statement into `self.handle_exception`?

lowercase00:
@selwin I think all comments are resolved now. I'm thinking that we should also add some logs to the `queue` to help capture the application's state when pushing the job to a queue, especially at `push_job_id` that way we could check whether the `rpush` has any issues. Thoughts?

lowercase00:
@selwin take a look at the suggestion I made on https://github.com/rq/rq/pull/1752/commits/55dee09dd7fe4958b01a9ae446f385de38f97862. There are a log of logs, but they should help understanding where the issue may be happening.

rq/queue.py
```diff
@@ -291,11 +305,13 @@ def compact(self):
     def push_job_id(self, job_id: str, pipeline: t.Optional['Pipeline'] = None, at_front=False):
         """Pushes a job ID on the corresponding Redis queue.
         'at_front' allows you to push the job onto the front instead of the back of the queue"""
+        self.log.debug(f"Pushing job {blue(job_id)} into the queue {green(self.name)}, current job count {self.count}")
```
selwin, line 308: `self.count` incurs a Redis operation so I'd rather not have job count displayed here.

rq/queue.py
```diff
@@ -291,11 +305,13 @@ def compact(self):
...
         connection = pipeline if pipeline is not None else self.connection
         if at_front:
-            connection.lpush(self.key, job_id)
+            result = connection.lpush(self.key, job_id)
         else:
-            connection.rpush(self.key, job_id)
+            result = connection.rpush(self.key, job_id)
+        self.log.debug(f"Pushed job {blue(job_id)} into the queue {green(self.name)}, total job count after pushing job: {result}")
```
selwin, line 314: `Pushed job {blue(job_id)} into {green(self.name)}, {result} job(s) are in in queue.`

rq/queue.py
lowercase00: We could have a toggle for "debug mode" to avoid the extra request while not debugging, what do you think? Without the count we would lose the ability to check whether the list was actually increased as it should.

rq/worker.py
```diff
@@ -1089,14 +1089,14 @@ def handle_job_success(self, job: 'Job', queue: 'Queue', started_job_registry: S
...
-        self.log.debug("Running Success Callbacks.")
+        self.log.debug(f"Running success callbacks for {job.id}")
         job.heartbeat(utcnow(), CALLBACK_TIMEOUT)
         with self.death_penalty_class(CALLBACK_TIMEOUT, JobTimeoutException, job_id=job.id):
             job.success_callback(job, self.connection, result)
 
     def execute_failure_callback(self, job):
         """Executes failure_callback with timeout"""
-        self.log.debug("Running Failure Callbacks.")
+        self.log.debug(f"Running Failure Callbacks for {job.id}")
```
selwin, line 1099: Nitpick: `failure callbacks` should not have uppercase as first letters :)

rq/queue.py
```diff
@@ -699,10 +715,13 @@ def lpop(cls, queue_keys, timeout, connection: t.Optional['Redis'] = None):
         if timeout is not None:  # blocking variant
             if timeout == 0:
                 raise ValueError('RQ does not support indefinite timeouts. Please pick a timeout value > 0')
+            logger.debug(f"Starting blocking pop operation for queues {green(queue_keys)} with timeout of {timeout}")
```
selwin, line 718: Nitpick: we should make the blocking pop terminology consistent. `Starting BLPOP for queues {green(queue_keys)} with timeout of {timeout}`

rq/worker.py
lowercase00: "Should" is a strong word to use :) But got your point will change that.

rq/queue.py
lowercase00: Introduced a `debug` param to the queue's `__init__`, and then logged the count only if the debug is `True` (defaults to `False`). This will avoid extra operations when not debugging while maintaining full visibility over the most relevant operations.

rq/queue.py
```diff
@@ -305,13 +306,15 @@ def compact(self):
     def push_job_id(self, job_id: str, pipeline: t.Optional['Pipeline'] = None, at_front=False):
         """Pushes a job ID on the corresponding Redis queue.
         'at_front' allows you to push the job onto the front instead of the back of the queue"""
-        self.log.debug(f"Pushing job {blue(job_id)} into the queue {green(self.name)}, current job count {self.count}")
+        self.log.debug(f"Pushing job {blue(job_id)} into the queue {green(self.name)}.")
+        if self.debug_mode:
```
selwin, line 310: Sorry, do you mind deleting this? I don't think we need this as we'll know the job count after job is pushed, sorry :(

rq/queue.py
lowercase00: Undone.

rq/queue.py
```diff
@@ -699,10 +714,13 @@ def lpop(cls, queue_keys, timeout, connection: t.Optional['Redis'] = None):
         if timeout is not None:  # blocking variant
             if timeout == 0:
                 raise ValueError('RQ does not support indefinite timeouts. Please pick a timeout value > 0')
+            logger.debug(f"Starting BLPOP operation for queues {green(queue_keys)} with timeout of {timeout}")
             result = connection.blpop(queue_keys, timeout)
             if result is None:
+                logger.debug(f"BLPOP Timeout, no jobs found on queues {green(queue_keys)}")
                 raise DequeueTimeout(timeout, queue_keys)
             queue_key, job_id = result
+            logger.debug(f"Found job {blue(job_id)} on queue {green(queue_key)}")
```
selwin, line 723: One more minor nitpick before I merge this in. Can we slightly rephrase this to `Dequeued job {blue(job_id)} from queue {green(queue_key)}`? This should make it clear that the job has been popped of the queue.

rq/worker.py
```diff
@@ -946,17 +948,20 @@ def prepare_job_execution(self, job: 'Job'):
...
+        self.log.debug(f"Preparing for execution of Job ID {job.id}")
         with self.connection.pipeline() as pipeline:
             self.set_current_job_id(job.id, pipeline=pipeline)
             self.set_current_job_working_time(0, pipeline=pipeline)
+            self.log.debug(f"Set Current Job ID & Working Time")
 
             heartbeat_ttl = self.get_heartbeat_ttl(job)
             self.heartbeat(heartbeat_ttl, pipeline=pipeline)
             job.heartbeat(utcnow(), heartbeat_ttl, pipeline=pipeline)
+            self.log.debug(f"Got Heartbeat. Current TTL is {heartbeat_ttl}")
```
selwin, line 960: `Heartbeat set with TTL of {heartbeat_ttl}`

rq/worker.py
```diff
@@ -946,17 +948,20 @@ def prepare_job_execution(self, job: 'Job'):
...
 
             job.prepare_for_execution(self.name, pipeline=pipeline)
             pipeline.execute()
+            self.log.debug(f"Job preparation finished.")
```
selwin, line 964: Actually I just realized that the commands for these three log statements are executed in a single `pipeline.execute()` call, meaning when the log is output, the current job ID, working time and heartbeat have not been set yet.

This means we should only log after a `pipeline.execute()`

lowercase00: Not sure I understand how you'd like this to be. My initial idea was just to set logs to understand what was the last active point before a potential issue - those particular logs don't necessarily depend on the pipeline execution. Anyways, fixed the logging text. This is the only one missing, let me know how you'd like this to be.

selwin: Sorry, then let's just delete these two lines:
1. `self.log.debug(f"Set Current Job ID & Working Time")`
2. `self.log.debug(f"Heartbeat set with TTL of {heartbeat_ttl}")`

I'll merge this PR in after these two are removed.

selwin:
Thanks!

PR #1261, opened 2020-05-25
Add exception catch for redis connection failure

corynezin:
Resolves #1153 
Resolves (Maybe?) #998

PR #1753, opened 2023-01-16
Enhanced Redis Connection Reliability

lowercase00:
The Redis connection may fail for several reasons. As the connection can be (1) explicitly passed to the worker or (2) implicity set, this will improve the Connection configuration by setting a timeout to the socket, and adding an ExponentialBackoff Retry logic.


rq/worker.py
```diff
@@ -19,6 +19,9 @@
 from uuid import uuid4
 from random import shuffle
 from typing import Callable, List, Optional
+from redis.retry import Retry as RedisRetry
+from redis.exceptions import TimeoutError
+from redis.backoff import ExponentialBackoff as RedisBackoff
```
lowercase00, line 24: It seems is only available with redis-py +4.0, need to double check.

rq/worker.py
```diff
@@ -262,6 +264,26 @@ def __init__(self, queues, name: t.Optional[str] = None, default_result_ttl=DEFA
...
+        if connection is None:
+            connection = get_current_connection()    
+        retry = RedisRetry(RedisBackoff(), 3)
+        timeout = default_worker_ttl - 5
+        keepalive = default_worker_ttl - 10
+        connection.socket_timeout = timeout
+        connection.set_retry(retry)
+        connection.socket_connect_timeout = timeout
+        connection.socket_keepalive = keepalive
+        return connection
```
lowercase00, line 285: Some of these params should be moved to `Redis.connection` attribute.

rq/worker.py
```diff
@@ -641,6 +664,13 @@ def work(self, burst: bool = False, logging_level: str = "INFO", date_format=DEF
                             )
                             break
 
+                except TimeoutError:
+                    if max_retry > retry_num:
+                        break
+                    retry_num += 1
+                    time.sleep(retry_num ** 3)
+                    continue
```
lowercase00, line 672: Simple Exponential Backoff connection retry, can be a parameter to the worker.

lowercase00:
Maybe related to PR #1753

lowercase00:
May be related to #1261

rq/worker.py
lowercase00: It seems the `Redis` class holds a `connection` attributes, that is being filled with a connection from the `connection_pool`, but every option passed to `Redis` goes directly to the connection pool, so the new version updates the `connection_kwargs` dictionary.

rq/worker.py
```diff
@@ -262,6 +263,26 @@ def __init__(self, queues, name: t.Optional[str] = None, default_result_ttl=DEFA
...
+        set a `socket_timout` in case the `BLPOP` command hangs (or any other command).
+
+        Args:
+            connection (Optional[Redis]): The Redis Connection.
+            default_worker_ttl (int): The Default Worker TTL
+        """
+        if connection is None:
+            connection = get_current_connection()    
+        retry = RedisRetry(RedisBackoff(), 3)
+        timeout = default_worker_ttl - 5
```
selwin, line 278: Take a look at how the actual timeout is calculated here. We need to base `socket_timeout` based on this. https://github.com/rq/rq/blob/fa7a635aa42b9626abf488d51863b2333d7837e5/rq/worker.py#L623

rq/worker.py
```diff
@@ -607,7 +628,8 @@ def work(self, burst: bool = False, logging_level: str = "INFO", date_format=DEF
                     self.scheduler.start()
 
         self._install_signal_handlers()
-
+        retry_num = 0
+        max_retry = 3
```
selwin, line 632: Let's not do anything with retry logic first, at least in this patch. Let's implement `socket_timeout` first and see how it goes. I want to be extra careful to not introduce any new behavior changes.

lowercase00:
I missed the version compatibility. It seems that `get_connection_kwargs()` was only added in 4.1.0, we currently use 3.5.0 for GH tests. This new commit changes the setup and gets the `connection_kwargs` directly from the `connection_pool` without any shortcuts, considers existing `socket_timeout` values set by the user, and adds a small session to the comments talking about this specific parameter.

rq/worker.py
```diff
@@ -262,6 +261,24 @@ def __init__(self, queues, name: t.Optional[str] = None, default_result_ttl=DEFA
...
+
+        Args:
+            connection (Optional[Redis]): The Redis Connection.
+            default_worker_ttl (int): The Default Worker TTL
+        """
+        if connection is None:
+            connection = get_current_connection()
+        current_socket_timeout = connection.connection_pool.connection_kwargs.get("socket_timeout")
+        if current_socket_timeout is None:
+            timeout = max(1, default_worker_ttl - 15) + 20
```
selwin, line 277: Nitpick: 10 seconds higher is enough.

Also, can we move the timeout calculation to a function so we can put the timeout calculation logic in one place?

docs/docs/connections.md
```diff
@@ -142,3 +142,23 @@ SENTINEL: {'INSTANCES':[('remote.host1.org', 26379), ('remote.host2.org', 26379)
            'DB': 2,
            'MASTER_NAME': 'master'}
 ```
+
+
+### Timeout
+
+To avoid potential issues with hanging Redis commands, specially the blocking `BLPOP`,
```
selwin, line 149: `specially the blocking BLPOP` -> `specifically the blocking BLPOP command`

docs/docs/connections.md
```diff
@@ -142,3 +142,23 @@ SENTINEL: {'INSTANCES':[('remote.host1.org', 26379), ('remote.host2.org', 26379)
...
+RQ automatically sets a `socket_timeout` value that is 20 seconds higher than the `default_worker_ttl`.
```
selwin, line 150: `that is 10 seconds higher than `default_worker_ttl``

selwin:
Friendly reminder to update this PR :)

docs/docs/connections.md
selwin: Mind making this change?

rq/worker.py
selwin: This comment also has not been addressed :)

lowercase00: Yeah, still need to fix this, since this logic is used when setting the connection (before this attribute is set on the class). Will get back to this in a couple of days.

selwin:
@lowercase00 tests didn't pass, mind taking a look? I'd like to include this in the next point release.

lowercase00:
Yes, I left a review comment open, moving the timeout calculation to a function breaks a few things because some attributes are not available when we actually use that method. I'll get back to this soon.

selwin:
Ok, no pressure. Even without this, we have good enough improvements to warrant a minor release. Thanks 🙏🏻 On Jan 31, 2023, at 6:15 PM, lowercase00 ***@***.***> wrote:﻿
Yes, I left a review comment open, moving the timeout calculation to a function breaks a few things because some attributes are not available when we actually use that method. I'll get back to this soon.

lowercase00:
@selwin should be good to go now.

selwin:
Thanks!

ronlut:
@lowercase00 Why is the private method `_get_timeout` not taking `self.default_worker_ttl` into account? Why does it use the const `DEFAULT_WORKER_TTL`? This ignores the actual worker ttl, is that on purpose?

lowercase00:
@ronlut this method is needed while setting the connection, at that moment, the `self.default_worker_ttl` still doesn't exist. It uses the `DEFAULT_WORKER_TTL` only if the `worker_ttl` param is not set, which is the desired behavior. I think we might have an issue here though: `timeout = None if burst else self._get_timeout()` -> we should be passing the `self.default_worker_ttl` as a param to the this method.

lowercase00:
Actually, I think what we should do here is `time = getattr(self, 'default_worker_ttl', DEFAULT_WORKER_TTL)` this would be more reliable (if we forget to pass the param to this function somewhere).

ronlut:
I don't think the `DEFAULT_WORKER_TTL` const should be used in that context at all, as it ignores the real worker ttl. The line you mentioned (for `burst`) is exactly the reason I asked that :)
I opened a PR with a suggested fix, WDYT? @lowercase00 @selwin 
https://github.com/rq/rq/pull/1794

rq/worker.py
lowercase00: @ronlut here

lowercase00:
So this function is only concentrating a logic that already existed. I'm pushing a suggested fix that only fixes this bug - we might want to review the logic itself though.

Issue #1758, opened 2023-01-19
Retrying failed tasks doesn't work when a rqworker goes down


lcosmin:
Running rq version 1.12.0.

I have a setup consisting of two `rqworker`s (running with `--with-scheduler`) on two sepparated machines and a task that fails constantly by raising an exception (for testing purposes). 

I enqueue the task using `enqueue_call(my_task, args=(arg1, arg2), retry=Retry(max=60, interval=10))`.

Watching the logs on both machines, I notice the task being executed, the exception raised, re-executed after the specified interval and so on. All seems to be working properly and **both** `rqworker`s can pick up the task to run it. 

However, if I shut down the worker **that first picked up** the task (ctrl-c), it seems that the task doesn't get executed by the second worker anymore. 

It's like the first worker coordinates the execution of the retry and once it's gone, so is the task. Of course, after I start it back up, the execution resumes. I would expect the second worker to continue running it. 

So, I wonder, is this the expected/desired behavior? Because in a cloud environment (where machines get downscaled from time to time) this would lead to losing tasks (if the "coordinator" goes down).

lowercase00:
@lcosmin just to make sure I can replicate this, you're stopping the worker (1) after the first exception right?

As far as I understand, the job (and the retry) shouldn't be attached to a worker, but rather to a queue, if both workers are listening to the same queue I guess they should be able to handle this situation. 

I'll double check and try to replicate this.

lowercase00:
@lcosmin I couldn't reproduce this. My current setup:

```python
# local/app.py
from redis import Redis
from rq import Queue, Retry
q = Queue(connection=Redis())
q.enqueue("local.task.mytask", retry=Retry(max=5))
```

```python
# local/task.py
import time
def mytask():
    time.sleep(2)
    raise Exception("This is an error")
```

Spin up two workers. Run app.py. After the Worker 1 get's the task and raises the exception, stop the worker with `ctrl + c` , worker two still gets the second task. Perhaps you're running workers without `--with-scheduler`? This is required when you are using `interval`.

> If you use `interval` argument with `Retry`, don't forget to run your workers using the `--with-scheduler` argument.

Feel free to add more details to reproduce this.

lcosmin:
I've managed to reproduce it locally using your example as a starting point. I did a slight variation so that the example matches my real world use case (i.e. use `enqueue_call` and have some arguments), but it seems the thing that triggers the behavior is using `interval=XX` as an argument for `Retry`. 

My code:

```python
[...]

PR #1759, opened 2023-01-22
Allow for configurable scheduler fallback period

lowercase00:
This adds a parameter to the Scheduler called
`fallback_period`, which determines the period
before work falls back to a new scheduler. 
Also reduces the default from 600 seconds to 300.
This improves #1758


rq/scheduler.py
```diff
@@ -98,7 +101,7 @@ def should_reacquire_locks(self):
             return False
         if not self.lock_acquisition_time:
             return True
-        return (datetime.now() - self.lock_acquisition_time).total_seconds() > 600
+        return (datetime.now() - self.lock_acquisition_time).total_seconds() > self.fallback_period
```
selwin, line 104: I think we can reduce this even further to 3 minutes. What do you think?

rq/scheduler.py
```diff
@@ -36,11 +37,13 @@ class RQScheduler:
 
     def __init__(self, queues, connection, interval=1, logging_level=logging.INFO,
                  date_format=DEFAULT_LOGGING_DATE_FORMAT,
-                 log_format=DEFAULT_LOGGING_FORMAT, serializer=None):
+                 log_format=DEFAULT_LOGGING_FORMAT,
+                 fallback_period=DEFAULT_SCHEDULER_FALLBACK_PERIOD, serializer=None):
```
selwin, line 41: I don't think we need to expose this parameter as this is a strictly internal thing. After we lower this threshold to 120 seconds, I think we no longer need this parameter, keeping things simpler.

lowercase00: I made a new commit removing the parameter, but keeping the value as a configuration (coming from the `DEFAULT_SCHEDULER_FALLBACK_PERIOD`). That way it should be more flexible to change in the future. Agreed that 120 would be better as well.

selwin:
Thanks!

PR #1760, opened 2023-01-23
Remove unused dummy tasks

lowercase00:
Dummy tasks were probably used for load testing
in the past, but currently are not being imported from anywhere.
Also, similar task utilities are already present on the tests fixtures, making the `dummy` module irrelevant.


selwin:
Thanks!

PR #1761, opened 2023-01-23
Remove compatibility layer for < 3.5

lowercase00:
Previously, with RQ supporting older Python versions (<3.5) a few helpers were necessary to ensure compatibility.
With `1.12` dropping 3.5 support, this becomes unnecessary.
This PR removes the `compat` layer completely.


selwin:
👍 I'd like to merge this in if we can get all the tests to pass.

lowercase00:
@selwin sure, it is as if the workflow was using an old version, the error is `ModuleNotFoundError: No module named 'rq.compat'` but this import does not exists in this branch anymore. I'll try to understand why.

selwin:
Thanks!

Issue #673, opened 2016-03-11
The example in the docs about worker


f0r:
In the docs about worker(http://python-rq.org/docs/workers/), the example like this:

```
#!/usr/bin/env python
import sys
from rq import Connection, Worker

# Preload libraries
import library_that_you_want_preloaded

# Provide queue names to listen to as arguments to this script,
# similar to rq worker
with Connection():
    qs = sys.argv[1:] or ['default']

    w = Worker(qs)
    w.work()
```

but it doesn't work, so I think maybe 

```
qs = sys.argv[1:] or ['default']
```

should be 

```
qs = Queue('default',connection=Redis())
```

then it works fine.

Issue #67, opened 2012-05-15
pypi installation is broken


selwin:
Hi Vincent,

Currently `rqinfo` and `rqworker` is broken if rq is installed with pip.

```
selwin@selwin:~$ rqinfo 
Traceback (most recent call last):
  File "/home/selwin/.virtualenvs/rq/bin/rqinfo", line 8, in <module>
    load_entry_point('rq==0.1.1', 'console_scripts', 'rqinfo')()
  File "/home/selwin/.virtualenvs/rq/local/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/pkg_resources.py", line 318, in load_entry_point
  File "/home/selwin/.virtualenvs/rq/local/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/pkg_resources.py", line 2221, in load_entry_point
  File "/home/selwin/.virtualenvs/rq/local/lib/python2.7/site-packages/setuptools-0.6c11-py2.7.egg/pkg_resources.py", line 1954, in load
ImportError: No module named scripts.rqinfo
```

This is caused by missing `scripts` directory from the pypi package.

PR #1765, opened 2023-01-27
Fix worker example

lowercase00:
Fixes [#67](https://github.com/rq/rq/issues/673)


docs/docs/workers.md
```diff
@@ -116,18 +116,14 @@ A simple implementation example:
...
-import sys
 from rq import Connection, Worker
 
 # Preload libraries
 import library_that_you_want_preloaded
 
-# Provide queue names to listen to as arguments to this script,
-# similar to rq worker
+# Provide the worker with the list of queues (str) to listen to.
 with Connection():
```
selwin, line 125: Can we eliminate this pattern from docs altogether? I want to slowly deprecate this context manager, would prefer explicitly passing in Redis connection to `Queue` and `Worker` objects.

lowercase00: I agree with removing the context manager, we already have the connection being injected, I think it's a lot cleaner to just have it as a parameter handled by the user.

I would keep the example (preloading library) and change the context manager to a explicit connection, what do you think?

selwin: Yes, this sounds good. Thanks!

PR #1769, opened 2023-01-27
Moves the `compact` helper to `utils`

lowercase00:
There is a helper funcion that excludes `None` values from a list. This was being declared both in the Queue and in the worker. This centralizes this helper in the `utils` Importing


selwin:
Thanks!

PR #1772, opened 2023-01-28
Job methods docstrings

lowercase00:
- Improves docstrings on the `connections` module.
- Adds docstrings to all methods on the `Job` class.
- ~Adds a `Protocol` to define serializers (`dumps` and `loads` methods)~ (this only exists in 3.8, excluded)

Still missing (next PR)
- Standardized docstrings for `Queue`
- Standardized docstrings for `Worker`

@selwin sorry this is fairly big, but it's a lot easier to do it all at once than N PRs with a single docstring each.
There's little to no change in logic, so this should be mostly reviewing the docs.

As this is fairly large, it would be preferable to not leave this hanging for too long, since it would cause quite a bit of conflict on future PR's which would be a huge pain to deal with.

### DIFF
--- a/rq/connections.py
+++ b/rq/connections.py
@@ -11,6 +11,19 @@ class NoRedisConnectionException(Exception):
 
 @contextmanager
 def Connection(connection: t.Optional['Redis'] = None):  # noqa
+    """The context manager for handling connections in a clean way.
+    It will push the connection to the LocalStack, and pop the connection
+    when leaving the context
+    Example:
+        ..codeblock:python::
+
+            with Connection():
+                w = Worker()
+                w.work()
+
+    Args:
+        connection (Optional[Redis], optional): A Redis Connection instance. Defaults to None.
+    """    
     if connection is None:
         connection = Redis()
     push_connection(connection)
@@ -33,9 +46,12 @@ def push_connection(redis: 'Redis'):
     _connection_stack.push(redis)
 
 
-def pop_connection():
+def pop_connection() -> 'Redis':
     """
     Pops the topmost connection from the stack.
+
+    Returns:
+        redis (Redis): A Redis connection
     """
     return _connection_stack.pop()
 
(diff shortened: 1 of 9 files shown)


rq/connections.py
```diff
@@ -11,6 +11,19 @@ class NoRedisConnectionException(Exception):
...
+        ..codeblock:python::
+
+            with Connection():
+                w = Worker()
+                w.work()
+
+    Args:
+        connection (Optional[Redis], optional): A Redis Connection instance. Defaults to None.
+    """    
     if connection is None:
```
selwin, line 27: Can we add a deprecation warning here?

lowercase00: Sure. I'm just leaving this PR as is, and continue the docstrings on a new one. Don't want to fall behind master too much. Already added that on the new one, so I'll resolve this for now.

lowercase00:
@selwin I think we can consider this finished and ready for review, I stashed the rest of the code I was working on for a next PR, don't want this to get too big.

rq/job.py
```diff
@@ -1054,30 +1355,38 @@ def register_dependency(self, pipeline: Optional['Pipeline'] = None):
...
         connection = pipeline if pipeline is not None else self.connection
 
         if pipeline is not None:
             connection.watch(*[self.key_for(dependency_id)
                                for dependency_id in self._dependency_ids])
 
-        dependencies_ids = {_id.decode()
-                            for _id in connection.smembers(self.dependencies_key)}
+        saved_job_dependents = connection.smembers(self.dependencies_key)
+        dependencies_ids = {_id.decode() for _id in saved_job_dependents}
```
selwin, line 1389: I think it's better if we stick with set comprehension here, it avoids the variable assignment `saved_job_dependents`

selwin:
@lowercase00 appreciate your help in doing this type of PRs. This really helps in the long term maintanability of RQ's code base.

rq/job.py
lowercase00: I personally feel separating this is a great win for style, easy of understanding, readability and maintenance, but I can roll this back.

selwin:
Thanks!

PR #1782, opened 2023-01-30
Queue Docstrings

lowercase00:
Adds more docstrings to a few modules.


lowercase00:
@selwin can we look into this before #1753 please? I know #1753 is more relevant, but this has a lot more changes (although no functionality changes), it will be easier to work on #1753 on top of this... thanks!

selwin:
Merged, thanks!

PR #1785, opened 2023-02-01
Bugfix: Fix bug when running worker on debug mode

### LINKED ISSUE #1786: Worker crashes when running on debug
We were printing some colored information on debug mode, but some of the data weren't suitable for the `green(...)` `blue(...)` functions: there was a `list[str]` and `bytes`. #1785 fixes it.

lowercase00:
When running on debug mode, some debug logs were
using colorizers on lists, which would raise an unhandled exception on the worker.
The same happened for a debug log that was using colors on a response from Redis (bytes).

To avoid changing the colorizer functions signature, I chose to fix this manually.
In the future we could have more robust logic for the colors (deconstruct lists, convert bytes to string, etc).

Fixes #1786

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -983,13 +983,14 @@ class Queue:
         if timeout is not None:  # blocking variant
             if timeout == 0:
                 raise ValueError('RQ does not support indefinite timeouts. Please pick a timeout value > 0')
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

Issue #1736, opened 2022-11-21
Add rq-scheduler missing functions to rq


gabriels1234:
Hi!
The --with-scheduler option is a great feature. and it has helped a lot in cases where the main scheduler fails.

I use django-rq-scheduler (the new repo) that relies on rq-scheduler for a few main functionalities. It could easily fallback to vanilla rq if some of the rq-scheduler functions are ported over rq.
(such as cron(), __contains__(), etc... )

The end goal is to avoid having a separate container (extra $$$) to run the scheduler only.

Thanks!

selwin:
Yes, that's the end goal. RQ also very recently gained a feature to track historical runs and results here so you can track the result of jobs that are run multiple times. https://github.com/rq/rq/pull/1725

PR #1787, opened 2023-02-01
Add to worker redis record scheduler info

### LINKED ISSUE #1783: Way to check if a worker is `with_scheduler`
Currently `with_scheduler` is passed only in `work(...)`, so when querying workers, there is no way to tell whether a worker has scheduler or not.

Proposed solution:
The workers that have scheduler should be stored in redis (eg, in a separate key, like `REDIS_WORKER_KEYS`).
And then when querying workers (using `Worker.all()` or `Worker.find_by_key(..)` it should instantiate the Worker class with a `with_scheduler` param.

cunla:
Fix #1783

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -196,7 +196,7 @@ class Queue:
         return 'rq:clean_registries:%s' % self.name
 
     @property
-    def scheduler_pid(self):
+    def scheduler_pid(self) -> int:
         from rq.scheduler import RQScheduler
         pid = self.connection.get(RQScheduler.get_locking_key(self.name))
         return int(pid.decode()) if pid is not None else None
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -3,11 +3,11 @@ import sys
 import traceback
 import uuid
 import warnings
-
 from collections import namedtuple
 from datetime import datetime, timezone, timedelta
 from functools import total_ordering
 from typing import TYPE_CHECKING, Dict, List, Any, Callable, Optional, Tuple, Type, Union
+
 from redis import WatchError
 
 if TYPE_CHECKING:
@@ -23,7 +23,6 @@ from .job import Job, JobStatus
 from .types import FunctionReferenceType, JobDependencyType
 from .serializers import resolve_serializer
 from .utils import backend_class, get_version, import_attribute, make_colorizer, parse_timeout, utcnow, compact
-
 
 green = make_colorizer('darkgreen')
 yellow = make_colorizer('darkyellow')
@@ -69,7 +68,7 @@ class Queue:
 
     @classmethod
     def all(
-        cls, connection: Optional['Redis'] = None, job_class: Optional[Type['Job']] = None, serializer=None
+            cls, connection: Optional['Redis'] = None, job_class: Optional[Type['Job']] = None, serializer=None
     ) -> List['Queue']:
         """Returns an iterable of all Queues.
 
@@ -94,11 +93,11 @@ class Queue:
 
     @classmethod
     def from_queue_key(
-        cls,
-        queue_key: str,
-        connection: Optional['Redis'] = None,
-        job_class: Optional['Job'] = None,
-        serializer: Any = None,
+            cls,
+            queue_key: str,
+            connection: Optional['Redis'] = None,
+            job_class: Optional['Job'] = None,
+            serializer: Any = None,
     ) -> 'Queue':
         """Returns a Queue instance, based on the naming conventions for naming
         the internal Redis keys.  Can be used to reverse-lookup Queues by their
@@ -119,18 +118,18 @@ class Queue:
         prefix = cls.redis_queue_namespace_prefix
         if not queue_key.startswith(prefix):
             raise ValueError('Not a valid RQ queue key: {0}'.format(queue_key))
-        name = queue_key[len(prefix) :]
+        name = queue_key[len(prefix):]
         return cls(name, connection=connection, job_class=job_class, serializer=serializer)
 
     def __init__(
-        self,
-        name: str = 'default',
-        default_timeout: Optional[int] = None,
-        connection: Optional['Redis'] = None,
-        is_async: bool = True,
-        job_class: Union[str, Type['Job'], None] = None,
-        serializer: Any = None,
-        **kwargs,
+            self,
+            name: str = 'default',
+            default_timeout: Optional[int] = None,
+            connection: Optional['Redis'] = None,
+            is_async: bool = True,
+            job_class: Union[str, Type['Job'], None] = None,
+            serializer: Any = None,
+            **kwargs,
     ):
         """Initializes a Queue object.
 
@@ -195,6 +194,12 @@ class Queue:
     def registry_cleaning_key(self):
         """Redis key used to indicate this queue has been cleaned."""
         return 'rq:clean_registries:%s' % self.name
+
+    @property
+    def scheduler_pid(self):
+        from rq.scheduler import RQScheduler
+        pid = self.connection.get(RQScheduler.get_locking_key(self.name))
+        return int(pid.decode()) if pid is not None else None
 
     def acquire_cleaning_lock(self) -> bool:
         """Returns a boolean indicating whether a lock to clean this queue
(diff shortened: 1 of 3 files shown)


lowercase00:
@selwin should be taking a look at this more closely, but I would have imagined the worker metadata should be managed exclusively by the worker, so this logic should probably be on the `worker` instead of on the `scheduler`.

The worker already has access to the scheduler information (`scheduler._process`) and knows whether it's working with or without a scheduler. In general, I would try to avoid having the worker metadata being changed by anything other than the worker itself.

The way this is setup we are having to pass the `worker_key` to the scheduler, so this is not very convenient, specially since we might be using a scheduler without a worker.

The change would actually be simpler without this, on the `if with_scheduler` block, we just set the scheduler `pid` and the `with_scheduler: 1` on Redis.

cunla:
It depends on the approach you want to take.

- I thought since an `RQScheduler` can only be instantiated by a worker, and it is more aware of its state (there is the catch exception in `run()` method), then it makes more sense to control its status here than have the worker check whether its scheduler is alive all over the place.
- Another option: Consider having a different redis-key for scheduler(s)? if you don't consider schedulers as "optional part of workers" then they can have `all()`, `get_by_key(..)`, etc. and all other methods like `Worker`, `Queue`, `Job`.

Different issue:
- When in `burst` mode - worker.scheduler should be None once done, but there is a test `test_work_burst` that checks it is not None, not sure if I fully understand the reasoning - burst mode means queue scheduled jobs once, not on-going.

lowercase00:
>  since an RQScheduler can only be instantiated by a worker

That's not really the case. I can get the RQScheduler normally and run it (that's actually what I do, since I prefer to have the scheduler be a separate service).

So when running a Worker, current behavior is: you instantiate the class, and then decide whether it's going to run with a scheduler in a separate process or no (`with_scheduler` at the `work(...)`method.

What I understand from your original issue is: "i want to be able to restore a worker, and take into consideration whether it was running with or without a scheduler". In that case, I think the solution would be simple: have a `with_scheduler` in the workers `__init__`, have a `_with_scheduler` private method to the worker, and have the same argument in the `work()` method, and now you (1) known whether the worker is running with or without a scheduler (2) can make that decision when running `work()` or when instantiating the `worker`.

In my personal opinion this is fine and should work well giving us extra flexibility.

--

Separate issues: we might open a new issue for that.

cunla:
> > since an RQScheduler can only be instantiated by a worker
> 
> That's not really the case. You can get the RQScheduler normally and run it (that's actually what I do, since I prefer to have the scheduler be a separate service). The `RQScheduler` was actually a completely separate project, and it's been slowly being ported to be a part of RQ, so it's on the to-do list to add the missing scheduler methods (see #1736)

That's a good point. I basically need to check what schedulers are running from django-admin - I thought to do it from the workers since I can query workers (using `Worker.all()`). But it makes more sense to have a separate key for RQSchedulers and be able to query them, etc.

lowercase00:
Got it. So that are two use cases here:
(1) Restore a worker with a scheduler (could be solved by having the `with_scheduler` key and `scheduler_pid` on the worker metadata
(2) Query for all schedulers, regardless of whether they are running through the worker or not.

The original `rq-scheduler` project does have namespaces and keys dedicated for the scheduler. So probably both things are necessary:
(a) improve the worker registry
(b) Port the rest of the scheduler code to this repo.

My suggestion is to use this PR to fix (a) and then track (b) through #1736

cunla:
> Got it. So that are two use cases here: 
> (1) Restore a worker with a scheduler (could be solved by having the `with_scheduler` key and `scheduler_pid` on the worker metadata 
> (2) Query for all schedulers, regardless of whether they are running through the worker or not.
> 
> The original `rq-scheduler` project does have namespaces and keys dedicated for the scheduler. So probably both things are necessary: (a) improve the worker registry (b) Port the rest of the scheduler code to this repo.

(a) requires deeper knowledge of how Worker works - mainly I can take it but on the first contribution to the repo.

(b) There is no need to migrate all the rq-scheduler code dealing with cron-jobs, etc. for use-case (2).  Just the registry part. ==> I can focus on that if acceptable.

lowercase00:
> (a) requires deeper knowledge of how Worker works - mainly I can take it but on the first contribution to the repo.

No worries, I can work on this. @selwin I'll add a `with_scheduler` and `scheduler_pid` keys to the worker metadata.

> (b) There is no need to migrate all the rq-scheduler code dealing with cron-jobs, etc. for use-case (2). Just the registry part.

I think the main idea is to totally deprecate the old `rq-scheduler` and concentrate everything in here, so I think that's desirable. Let's wait for @selwin word on those two topics.

selwin:
@cunla @lowercase00 I'd like the scheduler PID to be stored independently from the worker data. It should be stored in a separate key so that:
1. It's immediately accessible (by monitoring systems) without having to iterate through all workers and inspecting the `scheduler_pid` key within the worker hashes.
2. The logic of handling this key is purely kept within the `Scheduler` object

We can perhaps create a `rq:scheduler:<queue_name>` Redis hash containing information such as it's PID, when it's created etc.

lowercase00:
In this specific scenario where one is running the worker `with_scheduler=True`, I still think there are upsides to store this information in the worker data, specially considering the use case @cunla brought up where one wants to restore a worker with the same state. I don't really see any downside to this (having an extra key carrying information about the worker). At the end of the day this process is being managed by the worker.

 I do agree we want to port the scheduler registry from `RQScheduler` though, so perhaps this could be part of larger effort to port the scheduler to the main repo instead of an incremental change.

selwin:
> specially considering the use case @cunla brought up where one wants to restore a worker with the same state.

Not sure I see the use case for restoring a worker with the same state. In the [original issue](https://github.com/rq/rq/issues/1783) , I think we just need to know whether a worker is run with `--with-scheduler` and the scheduler PID info :).

> I do agree we want to port the scheduler registry from RQScheduler though, so perhaps this could be part of larger effort to port the scheduler to the main repo instead of an incremental change.

We actually already have a [ScheduledJobRegistry](https://github.com/rq/rq/blob/fdb14df18141b6bedc40d719904f56e64984bc37/rq/registry.py#L328) that keeps track of scheduled jobs in RQ. The `SchedulerRegistry` I mentioned just now would be a registry keeping track of all `Scheduler` instances, but upon thinking about this further I think we can skip this for now because it's unlikely that we'll be managing a large number of scheduler instances.

cunla:
Please let me know how you want to proceed with this.

selwin:
@cunla there's also another effort to show scheduler PID in django-rq. Take a look at this pull request: https://github.com/rq/django-rq/pull/570/files#diff-86374d856b018c779d498e105c46bd21bdd07450d63a28fa8fd440a9c42272cfR43

I think this is a quick win and is straight forward, let's bring this to RQ proper. We can start with having `queue.scheduler_pid` that returns the PID of the scheduler that works on RQ.

cunla:
> @cunla there's also another effort to show scheduler PID in django-rq. Take a look at this pull request: https://github.com/rq/django-rq/pull/570/files#diff-86374d856b018c779d498e105c46bd21bdd07450d63a28fa8fd440a9c42272cfR43
> 
> I think this is a quick win and is straight forward, let's bring this to RQ proper. We can start with having `queue.scheduler_pid` that returns the PID of the scheduler that works on RQ.

Done. Notice including it in `Queue` (vs. in `RQScheduler`) requires a local import to avoid circular dependency.

rq/queue.py
```diff
@@ -196,6 +195,12 @@ def registry_cleaning_key(self):
         """Redis key used to indicate this queue has been cleaned."""
         return 'rq:clean_registries:%s' % self.name
 
+    @property
+    def scheduler_pid(self):
```
selwin, line 199: Mind adding a return type annotation to this method?

cunla: sure

selwin:
@cunla mind fixing the tests so I can merge this in?

cunla:
> @cunla mind fixing the tests so I can merge this in?

I am not sure what's going on with the tests - They pass on my branch: https://github.com/cunla/rq/actions/runs/4154089123

lowercase00:
@cunla Im not sure id thats the case here, but I had issues with when not running slow tests locally (the default), maybe if you run with`RUN_SLOW_TESTS_TOO=1` it will run everything and you can try to replicate the CICD

cunla:
> @cunla Im not sure thats the case here, but I had issues with when not running slow tests locally (the default), maybe if you run with`RUN_SLOW_TESTS_TOO=1` it will run everything and you can try to replicate the CICD

The exact github workflow is passing on my branch (see link).

I added a commit to reformat some code to try running them again.

cunla:
Please trigger the run again. It seems to [work on my fork well](https://github.com/cunla/rq/actions/runs/4157101257) second time in a row.

selwin:
Thanks!

PR #1789, opened 2023-02-02
Bugfix: sets `exc_info` when working with `is_async=False`

### LINKED ISSUE #1633: `exc_info` not available when running jobs with `is_async=False`
I'm running jobs that may raise exceptions which would be forwarded to the user via `job.exc_info`. I wanted to test this by setting the queue to sync (aka `is_async=False`). Now some jobs intentionally fail, however the `exc_info` is not set. It stays on `None` while the job status is set to *failed*.

Is this intended behavior? If not I'll look into fixing this.

lowercase00:
Fix #1633. Currently `exc_info` is only being set by the worker on the `handle_job_failure` method. This adds the `exc_info` to the queue, when running a sync job.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -1,3 +1,4 @@
+import traceback
 import uuid
 import sys
 import warnings
@@ -855,6 +856,8 @@ class Queue:
             job.set_status(JobStatus.FAILED)
             if job.failure_callback:
                 job.failure_callback(job, self.connection, *sys.exc_info())
+            exc_info = sys.exc_info()
+            job._exc_info = ''.join(traceback.format_exception(*exc_info))
         else:
             if job.success_callback:
                 job.success_callback(job, self.connection, job.result)


selwin:
Mind slightly reworking this part to use `handle_job_failure()`? I feel like this would be the best way forward to ensure that we always have the same job success/failure handling.

lowercase00:
Sure, but in that case we would need a bigger refactoring - `handle_job_failure` is a worker method, and in this case we are not using the worker at all when `is_async=False`.

We would probably have to split some parts of the `handle_job_failure` by adding a `handle_failure` to the job itself, and then using this new method on the worker. The worker would then call `handle_failure`, and do what it needs to do with the failure (increment count, working time, etc).

selwin:
> We would probably have to split some parts of the handle_job_failure by adding a handle_failure to the job itself, and then using this new method on the worker. The worker would then call handle_failure, and do what it needs to do with the failure (increment count, working time, etc).

Yes, we'd need to separate the worker specific failure handling logic from the core job failure handling logic. I'm ok if you want to merge this PR in it's current state though, in any case a bug fix is a bug fix. Just pointing out the long term approach that we need to take to properly address this issue :).

Let me know whether you want to refactor this now or if I should just merge this PR in. Time not spent on not refactoring this part also means time spent on improving other parts of the library :)

lowercase00:
Totally agree and thanks for pointing that out! I still don't have a "birds eye" view as you do, slowly learning!

My 2 cents: I feel this logic is not that straight forward to change: the exception itself is being saved on the `add` method of the `FailedJobRegistry`, which is being called by the worker. But considering this is (1) a bug and (2) a two line change, I think we could solve the bug now, and then add a new task to optimize this and have it happen in a single place. I would be happy to work on improving this in the future though :)

selwin:
@lowercase00 please take a look at https://github.com/rq/rq/pull/1799

lowercase00:
Superseded by #1799

PR #1793, opened 2023-02-04
Fix - TypeError - accessing None when dequeued result is None (when timeout=None, e.g. in burst mode)

ronlut:
result might be None if burst=True for example

added a test (failing before the fix)

(This was introduced in #1752)

### DIFF
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -640,8 +640,7 @@ class TestWorker(RQTestCase):
         w = Worker([q])
 
         # Put it on the queue with a timeout value
-        res = w.dequeue_job_and_maintain_ttl(None)
-        assert res is None
+        assert self.assertIsNone(w.dequeue_job_and_maintain_ttl(None))
 
     def test_worker_sets_result_ttl(self):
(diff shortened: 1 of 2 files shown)


lowercase00:
Thanks for catching that @ronlut !

tests/test_worker.py
```diff
@@ -634,6 +634,15 @@ def test_timeouts(self):
...
         self.assertIn('JobTimeoutException', as_text(res.exc_info))
 
+    def test_dequeue_job_and_maintain_ttl_non_blocking(self):
+        """Not passing a timeout should return immediately with None as a result"""
+        q = Queue()
+        w = Worker([q])
+
+        # Put it on the queue with a timeout value
+        res = w.dequeue_job_and_maintain_ttl(None)
+        assert res is None
```
selwin, line 644: Mind changing this to `self.assertIsNone(w.dequeue_job_and_maintain_ttl(None))`?

selwin:
Mind resolving the conflict so I can merge this in?

selwin:
@ronlut can you check the tests?

selwin:
Thanks!

Issue #1642, opened 2022-04-04
Job decorator results_ttl default overrides worker's default value


veejar:
https://github.com/rq/rq/blob/master/rq/decorators.py#L18
Here "result_ttl=DEFAULT_RESULT_TTL" should be replaced with "result_ttl=None", because it overrides worker's default result_ttl, which is also DEFAULT_RESULT_TTL when not specified as argument for worker.

selwin:
Mind opening a PR for this?

PR #1794, opened 2023-02-04
Fix - Use worker TTL for timeout

ronlut:
Worker ttl was ignored.
BTW, do we have a better name for `timeout`? It was names `_get_timeout` before, but maybe something more specific can help with readability

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -227,7 +227,11 @@ class Worker:
         serializer=None,
     ):  # noqa
 
-        connection = self._set_connection(connection, default_worker_ttl)
+        self.default_result_ttl = default_result_ttl
+        self.default_worker_ttl = default_worker_ttl
+        self.job_monitoring_interval = job_monitoring_interval
+
+        connection = self._set_connection(connection)
         self.connection = connection
         self.redis_server_version = None
 
@@ -249,10 +253,6 @@ class Worker:
         self.validate_queues()
         self._ordered_queues = self.queues[:]
         self._exc_handlers: List[Callable] = []
-
-        self.default_result_ttl = default_result_ttl
-        self.default_worker_ttl = default_worker_ttl
-        self.job_monitoring_interval = job_monitoring_interval
 
         self._state: str = 'starting'
         self._is_horse: bool = False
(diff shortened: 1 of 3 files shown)


lowercase00:
@ronlut sorry, opened another PR without viewing this. I'll close the other one. We might want a test though (if you want to take the one for the other PR)

ronlut:
@lowercase00 took your test, thanks

rq/worker.py
```diff
@@ -227,7 +227,11 @@ def __init__(
         serializer=None,
     ):  # noqa
 
-        connection = self._set_connection(connection, default_worker_ttl)
+        self.default_result_ttl = default_result_ttl
+        self.default_worker_ttl = default_worker_ttl
```
selwin, line 231: Can we rename `self.default_worker_ttl` to `self.worker_ttl`? This is the actual worker TTL that will be used.

`self.default_result_ttl` can be kept unchanged because it's indeed a "default", each job can specify it's own TTL.

selwin: In version 2.0, we can also change the `default_worker_ttl` argument to `worker_ttl`

selwin:
> BTW, do we have a better name for timeout? It was names _get_timeout before, but maybe something more specific can help with readability

What about `dequeue_timeout`?

ronlut:
I split it to two properties, do you think connection_timeout should use `dequeue_timeout`? Are they dependant?

selwin:
I'm quite happy with this PR. Just want to make sure that @lowercase00 is ok since he worked on connection timeout.

lowercase00:
LGTM, I think that's a lot better than what it was.
@ronlut they are dependent, yes.

The reason I had made that function was because of the order of the attributes, and this fixes that, so it's a lot better, cleaner.

The connection timeout is a safe guard: since there were reports of workers not reading from the queue, and there were reports of the hanging BLOP when used with Redis Py, we added the `socket_timeout` to the connection, so if the BLOP hangs, the socket timeout would raise a `TimeoutError`.

With our current terminology if the timeout to dequeue a job doesn't work (`dequeue_timeout`) the socket would timeout (`connection_timeout`) - I think it all makes sense.

lowercase00:
Also maybe this fixes #1642 ? Didn't look the code, but seems related.

rq/worker.py
```diff
@@ -223,7 +223,11 @@ def __init__(
         serializer=None,
     ):  # noqa
 
-        connection = self._set_connection(connection, default_worker_ttl)
+        self.default_result_ttl = default_result_ttl
+        self.worker_ttl = default_worker_ttl
```
selwin, line 227: Sorry, one more nitpicky thing, `worker_ttl` makes it sound like the worker itself only lives for a certain amount of time (which it doesn't). So if we set 60 seconds as TTL, this makes it seems like the worker would only live for 60 seconds.

Can we rename this to `self.dequeue_ttl = default_worker_ttl`?

I have put renaming `Worker(default_worker_ttl)` in the list of things to change in version 2.0, since this won't be backwards compatible https://github.com/rq/rq/issues/1800

ronlut: `def heartbeat` uses this ttl as well, as well as `register_birth`. Are you sure about this?

selwin: Yes, this should be a simple find and replace situation so should be ok.

ronlut: I am not talking about the rename technically 😅 , that's not a problem - I did that already from `default_worker_ttl` to `worker_ttl`.
I am meaning semantically, calling it `dequeuet_ttl`, when `heartbeat` and `register_birth` using it is a bit confusing isn't it? Is it right for heartbeat and birth to use "dequeue_ttl" as their reference?

selwin: Ah I see, you have a point. I'll merge this in.

selwin:
Thanks!

PR #1795, opened 2023-02-04
Worker - max_idle_time feature

ronlut:
Use case: 
We are autoscaling rq workers in our k8s cluster by the length of the tasks queue, using Keda.
Because we are running long running tasks with rq (we are talking hours here), we use k8s `Job` object to scale up workers, so they won't be scaled down. 
We set `max_jobs` to 1 so each worker executes 1 job and scales itself down. The thing is, if a worker is booted but then it has no tasks in queue, it will stay up until another task enters the queue. We want to be able to set a maximum idle time for workers.
The only option is using `burst` mode, but because we have a minimum of X workers up all time, they will die and reborn indefinitely if there are no tasks in the queue.

(branch includes the fix in #1793, so ignore those changes until it's merged)


ronlut:
@selwin WDYT?

selwin:
The concept of having a worker wait for X amount of time before quitting sounds good. I'll take a look at this PR soon, thanks!

tests/test_worker.py
```diff
@@ -607,6 +607,20 @@ def test_deleted_jobs_arent_executed(self):
...
+        q = Queue()
+        w = Worker([q])
+        q.enqueue(say_hello, args=('Frank',))
+        result = w.dequeue_job_and_maintain_ttl(1)
+        assert result is not None
+
+        result = w.dequeue_job_and_maintain_ttl(None)
+        assert result is None
+
+        result = w.dequeue_job_and_maintain_ttl(1, max_idle_time=1)
```
selwin, line 621: Can we structure this test such that the `max_idle_time` logic is tested?

If we make the dequeue TTL 1 second and `max_idle_time` 3 seconds, then the worker will attempt to dequeue 3 times before exiting, right?

ronlut: right, clarified the test a bit, let me know if it makes sense

selwin:
This is good, could you please make sure that the heartbeat test passes? It's returning 3 heartbeats instead of 4 in this test case https://github.com/rq/rq/actions/runs/4102519117/jobs/7075579683

selwin:
@ronlut there's also conflicts that need to be fixed.

rq/worker.py
```diff
@@ -839,7 +844,7 @@ def stop_scheduler(self):
                 pass
             self.scheduler._process.join()
 
-    def dequeue_job_and_maintain_ttl(self, timeout: int) -> Tuple['Job', 'Queue']:
+    def dequeue_job_and_maintain_ttl(self, timeout: Optional[int], max_idle_time: Optional[int] = None) -> Tuple['Job', 'Queue']:
```
selwin, line 847: Shouldn't `max_idle_time` be factored into `timeout` calculation? I feel like there should be something like this:
```python
dequeue_timeout = min(max_idle_time, timeout)
self.queue_class.dequeue_any(.., timeout)
```

ronlut: I agree, I changed it a bit to really be closer to `max` in `max_idle_time` - wait less in case there is only a few more seconds left until we reach `max_idle_time`, LMKWYT

ronlut:
@selwin Are you aware with any problem with timings in the tests? I am not sure why it fails the test, it passes locally with the same redis versions. Before the latests changes I used a count of heartbeats and it failed as well, I guess because too much time passes between a dequeue to the following one

selwin:
@ronlut in general you'd need to be a little bit looser on the timing of heartbeats because the tests are run on CIs which may or may not be that responsive.

This means we need to design the tests such that they are a bit looser. For example we can make `max_idle_time=3` and expect 3 or 4 heartbeats to be sent and then set `max_idle_time=5` and expect 5 or 6 heartbeats.

selwin:
Thanks!

PR #1796, opened 2023-02-04
Fix worker TTL not being considered

lowercase00:
The default worker ttl set as a parameter was not being considered after we've concentrated the timeout
calculation into a single place.


lowercase00:
Marking as superseded by #1794

PR #1798, opened 2023-02-04
Debug mode log fixes

lowercase00:
Current worker on debug mode doesn't raises errors, but does have some formatting issues.


selwin:
Thanks!

Issue #1792, opened 2023-02-04
job.return_value() returns None when When `Async=False` but works for job.result


shirecoding:
when running in async=False

```python
job = logging_task.enqueue(...)
print(f"""

job.result: {job.result}  # --------------> Returns actual result
job.return_value(): {job.return_value()}  # ----------> None

""")
```

Is this a bug, this makes the code not work when switching between TEST and PRODUCTION environments because in tests we want to use async=False, but it does not work with return_value()

lowercase00:
This is related to #1789 
There seems to be a few things about the job that are being handled by the worker, so the when in running sync on there are a couple of things missing from the job indeed. Thanks for pointing this out

shirecoding:
Thanks for the reply, i'm actually using the on_success callback and saving the result into my own data structure as its more consistent at the moment

selwin:
Yeah, this is definitely a bug. `job.return_value()` should still return the result of the execution in sync mode. I'll make a PR for this.

PR #1799, opened 2023-02-05
Sync jobs should save job exceptions and results

selwin:
Fixes https://github.com/rq/rq/issues/1792


lowercase00:
@selwin this is a lot more robust than the previous solution, +1 to this I'm closing the other PR.
Do these method need to be public? Trying to think of a scenario that the user will need to `handle_x` directly, is there any?

Issue #734, opened 2016-08-03
Timeout does not stop multithreaded jobs


nirfuzz:
As posted here:
http://stackoverflow.com/questions/38741173/rq-timeout-does-not-kill-multi-threaded-jobs

It seems like timeout argument takes no effect if the job uses threads, for example:

```
from threading import Thread

class MyThing(object):
    def say_hello(self):
        while True:
            print "Hello World"

    def hello_task(self):
        t = Thread(target=self.say_hello)
        t.daemon = True # seems like it makes no difference
        t.start()
        t.join()
```

When running `say_hello` as a job with timeout argument passed, a timeout exception is eventually thrown and the job is moved to 'failed' queue successfully, where running `hello_task` (which runs `say_hello` as a thread) goes on forever.

wpromatt:
That's because the timeout is implemented by having the worker parent send `signal.SIGALRM` to the job. In python signals aren't forwarded to threads. You will need to assign a signal handler in the process that starts the thread so you can manually shutdown the thread when you receive the `SIGALRM` signal.

```
from threading import Thread, Event
import signals

class MyThing(object):
    def say_hello(self, stop_event):
        while not stop_event.is_set():
            print 'Hello World'

    def hello_task(self):
        stop_event = Event()
        self.setup_signals(stop_event)

        t = Thread(target=self.say_hello, args=(stop_event,))
        t.daemon = True
        t.start()
        t.join()

    def setup_signals(self, stop_event):
         # So you  don't just swallow the default handler set by the rq.
         default_handler = signals.getsignal(signal.SIGALRM)

         def stop_thread(signum, frame):
             stop_event.set()
             if callable(default_handler):
                 default_handler(signum, frame) 
           
           # This is the signal sent from the worker on timeout.
          signals.signal(signals.SIGALRM, stop_thread)
```

Somewhat related, an issue I have is that there is no way to listen for `SIGINT` or `SIGTERM` in the job process since they are swallowed by the parent the first time and then an immediate `SIGKILL`. This is frustrating because I have numerous jobs that are running external `subprocesses` which need to be killed if the job goes away. It might be nice to have a way for the running job to also be alerted of the warm-shutdown request.

megapctr:
Thank you! I've been fiddling with my workers over the last week and couldn't figure out why they would just freeze at random and eventually return to work after a long pause. Luckily, in my case I can ditch the ThreadPoolExecutor and just throw more workers at the queue 😄

[...]

Issue #404, opened 2014-08-21
a thread worker


lileixuan:
a thread worker in my project,for now it work well.

``` python
# -*- coding:utf-8 -*-
from rq.worker import Worker as rq_Worker
import signal
import logging
from rq.timeouts import BaseDeathPenalty, JobTimeoutException
from rq.worker import StopRequested, green, blue
from rq.exceptions import DequeueTimeout
from rq.job import Status
from rq.logutils import setup_loghandlers
from rq.version import VERSION
import threading
import thread
import ctypes
import time


slave_counter = 0
_threads = {}


def _async_raise(tid, exctype):
    res = ctypes.pythonapi.PyThreadState_SetAsyncExc(ctypes.c_long(tid), ctypes.py_object(exctype))
    if res == 0:
        raise ValueError('thread[%s] nonexistent thread id' % (tid,))
    elif res > 1:
        # if it returns a number greater than one, you're in trouble,
        # and you should call it again with exc=NULL to revert the effect
        ctypes.pythonapi.PyThreadState_SetAsyncExc(ctypes.c_long(tid), 0)
        raise SystemError("thread[%s] PyThreadState_SetAsyncExc failed" % (tid,))

def _pthread_cancel(ident):
    try:
        pthread = ctypes.cdll.LoadLibrary("libpthread.so.0")
        pthread.pthread_cancel(ctypes.c_ulong(ident))

        with threading._active_limbo_lock:
            if ident in threading._active:  # and re != 0:
                try:
                    del threading._active[ident]
                except:
                    pass
    except Exception as e:
        pass


class WorkerThread(threading.Thread):

    def __init__(self, job, **kwargs):
        """Set up thread in daemonic mode and start it immediatedly.
        """
        threading.Thread.__init__(self, **kwargs)
        self.setDaemon(1)
        self._exit = False
        self.job = job
        self.timeout = None

    def stop(self):
        self.force_exit()

    def exit(self):
        self._exit = True
        _async_raise(self.ident, SystemExit)

    def force_exit(self):
        _pthread_cancel(self.ident)


class ThreadDeathPenalty(BaseDeathPenalty):

    def __init__(self, timeout):
        self._timeout = timeout + time.time()
        self.current_thread = threading.current_thread()
        self.current_thread.timeout = timeout
        self._ident = self.current_thread.ident  #thread.get_ident()

    def setup_death_penalty(self):
#        self.thread_timeout = threading.Timer(self._timeout, self.force_exit)
[...]

Issue #45, opened 2012-03-28
Worker concurrency?


kennethreitz:
It'd be nice if there was a built in mechanism to use multiprocess, eventlet, or gevent to have rworker handle _n_ tasks concurrently.

I don't think it would complicate things.

joostdevries:
I see that this issue has been stale for a while but as someone who is just switching to RQ, I'd like to offer my thoughts:

We moved away from Celery because of what kind of beast it has become supporting different backends, brokers and serializers. If RQ wants to support concurrency, I'd pick **one** method (preferably one that doesn't add any extra dependencies) and delegate any other implementation to plugins/other libs. (rq-gevent) etc. 

I LOVE the fact that I can quickly glance over the entire RQ source and understand what's going on. With Celery + Kombu that was completely impossible.

selwin:
@joostdevries I'm completely in agreement with you. Ideally, I'd like to have a reference implementation using multiprocessing in the core library. The rest can live as separate, third party libraries implementing worker classes using different concurrency methods. I just haven't had the time/need to work on this feature yet.

Koed00:
I think having a standard implementation using multiprocessing is the way to go. You don't want to get caught up in too many dependencies. Personally, I do have a practical need for this. My projects run on Heroku and that means I have to dedicate a 512Mb quadcore dyno to a single worker process. I've been moving longer running tasks to RQ and these are blocking the high priority short tasks. This has led me to set up a second worker on a separate dyno for high priority tasks .A single dyno could easily manage all of these, and more, with multiprocessing.

arikfr:
@Koed00 as a workaround you can run multiple processes on a single Dyno (without defining multiple process types). Maybe these days there are more intelligent ways to do it, but you can run Honcho (to start multiple processes) as the process Heroku runs for you in this dyno.

Koed00:
@arikfr starting multiple workers with honcho is definitely a way to do this and I've tried it. It works, but it feels hacky and the teardown of the workers fails more often than not.

I've been trying a more direct approach with:

```
workers = []
for i in range(4):
    p = multiprocessing.Process(target=worker.work)
    workers.append(p)
    p.start()
```

and this seems to work fine too, except again for the teardown.

chenxin-csu:
I think its quite a simple way to use multiprocessing this way!but how do u mean by saying about the teardown issue?

Koed00:
[...]

PR #1803, opened 2023-02-07
Multithread Worker

### LINKED ISSUE #1804: Create a new `ThreadPoolWorker`
This adds to the list of concurrency models we may want to implement (along side a `WorkerPool`, `GeventWorker`, `AsyncWorker` and `MultiprocessWorker`), also related to #45. Some work being done on #1803.

We replicated all tests that `Worker` went through using the `ThreadpoolWorker`, that means the basic Worker test suite, dependencies tests, callbacks tests and registry tests. Most of the tests pass, an overview on the issues so far (I haven't spend time looking into them):

Failed tests:
- Timeout & Failed registry: a job that times out doesn't seem to be sent to the `FailedJobRegistry`
- Dependency's Race Condition: we don't seem to be able to get the `job.result` when job's being executed
- Main dependency test fails: jobs seems to be stuck at `queued` while we should be `finished`
- Dependencies are met at execution time: we can't seem to get the job result
 
Other todo's:
- Test the worker is not eager: doesn't take more jobs than the available idle threads (this should be working fine, but should be better tested)
- Save the worker class and the pool size on the worker registry
- Add the scheduler (not implemented)
- Any other thread-safety issues we want to test?

Any help on this is very much appreciated!

ccrvlh:
This is a (very) rough sketch of what a multithreaded worker may look like.

The way this is currently implemented is: the main worker process is responsible for dequeuing jobs and monitoring the pool, if the pool has a free slot, it dequeues a job, and submits it to the executor, if it doesn't, it waits. It uses the `TimerDeathPenalty` to ensure the timeout, and receives a `pool_size` argument (Python's default is CPU * 5, we could adopt something similar).

To monitor the thread pool, it uses a simple counter that is protected by a `Lock`, and decremented (the argument is `idle_pool_slot`) as soon as a job occupies a thread, and incremented (the free slots) as soon as the job finishes (through a `done_callback`) - the logic was built on top of an `idle_pool`, we could easily do the other way round as well and say `running_threads` and compare it to the `max_workers`.

I though this implementation was the simplest (no need to spin threads manually and kill one by one), and the one that required the least amount of customization on the main worker code (we like simple 😀). It would be great if I was wrong and there's an even simpler way!

The only reason the `work()` method was customized, was because of the waiting it does while checking for a free slot in the pool, other than that, is a copy/paste of the original. Same as the `_shutdown()` method: it's only there to actually shutdown the `executor`.

I copied the Worker tests and ran all of them using the `ThreadpoolWorker`. Currently, three of them fail: (1) timeout moves job to failed registry (2) race condition on dependencies and (3) result lifespan. I didn't spend time looking into why.

Concurrency has been a big topic for quite a while now. At the moment both the `SimpleWorker` and the `Worker` are very mature, but to be honest, it's going to take quite a while before a `ThreadpoolWorker` becomes as robust and mature as the existing ones (or any other new worker).

However, this topic has been hanging for close to 10 years now. I would favor a simplistic `beta` release approach, in which not all features are necessarily implemented (dependencies, custom handlers, etc) and tested. I think this would greatly help us understand the issues with this worker, room for improvements and ways to make it as solid as the other ones - we could use warnings and be explicit about its experimental nature.

This can be extremely useful for I/O bound workloads.
I would love comments, contributions, critics, and anything else that could help us move forward with adding native concurrency support to RQ - leaving as draft so we can discuss on top of it.

Related to #45 #404 #734 #1804
Fixes #1804


selwin:
> Concurrency has been a big topic for quite a while now. At the moment both the SimpleWorker and the Worker are very mature, but to be honest, it's going to take quite a while before a ThreadpoolWorker becomes as robust and mature as the existing ones (or any other new worker).

> However, this topic has been hanging for close to 10 years now. I would favor a simplistic beta release approach, in which not all features are necessarily implemented (dependencies, custom handlers, etc) and tested.

Agree, we'll definitely label this as beta when first introduced.

Another approach that I think is better, and builds on the robustness of existing worker implementations is the concept of having a `WorkerPool`. Which is a supervisor process that spawns workers and watches over them, spawning new ones whenever necessary (this is how gunicorn works if I'm not mistaken, so I'm comfortable with this model). I started implementing this a few months ago here: https://github.com/rq/rq/blob/watcher/rq/worker_pool.py

lowercase00:
> Another approach that I think is better, and builds on the robustness of existing worker implementations is the concept of having a WorkerPool. Which is a supervisor process that spawns workers and watches over them, spawning new ones whenever necessary

So this would be a bit more similar to what Celery does in which you have a single worker, and you just pass the `--concurrency` num and the `--pool` type? Would that mean rewriting the worker to be totally agnostic in terms of the concurrency model?

selwin:
> So this would be a bit more similar to what Celery does in which you have a single worker, and you just pass the --concurrency num and the --pool type? Would that mean rewriting the worker to be totally agnostic in terms of the concurrency model?

It's actually the inverse, so we don't need to rewrite any worker code (or very minimal, hopefully). The `WorkerPool` simply instantiate workers and only send management signals (e.g shutdown) to the workers.

Let's say we start the worker with this CLI command `rq workerpool -n 5`. Inside the worker pool we'd do something like this:
```python
for range(n):
    worker = Worker()
    worker.start()  # This would run in a separate process like the scheduler https://github.com/rq/rq/blob/master/rq/worker.py#L668
```

The good part about this concept is that we can keep the worker simple because it only has to deal with dequeueing and executing it's own jobs and keep track of its own results etc, which it has been doing for the past 10 years.

The `WorkerPool` is just a higher level of abstractions that makes it easier for us to start more workers.

lowercase00:
> It's actually the inverse, so we don't need to rewrite any worker code (or very minimal, hopefully). The WorkerPool simply instantiate workers and only send management signals (e.g shutdown) to the workers.

Got it, makes sense after reading the code more carefully.

> this implementation also leaves room for someone to implement their own multi-threaded worker. If the Worker is to be refactored to make this implementation easier

Doesn’t `WorkerPool` implementation, being a layer of abstraction on top of the worker _reduces_ the flexibility for custom classes, since it hides details that one would want to customize to make other concurrency models work? This is fine, and is sure helpful to run a simple pool of workers, but I don’t see this as being a reason. I guess this implementation would have to have very strong opinions on how to run multiple workers within it, so it seems that the path to a multithreaded worker (or any other concurrency model) doesn’t use a `WorkerPool`,  but rather a new `Worker`

One thing to keep in mind is the datastore, that’s something to pay attention to, specially in documentation, since this can get confusing:

- For new users, the "worker" might change slightly - now "worker" and "workerpool" are two concepts that overlap depending on how you look at it, the user can now start a `workerpool` (a level of abstraction on top of the worker), but without much control over the worker (child) itself, since the top level process that one started through the cli (what we have been calling `worker`) is just a pool controlling workers.
- The datastore would be showing multiple workers, but as these workers are hidden (they are not being controlled directly by the user) we might end up in a crossroad, where you are only controlling (spinning up, cancelling, shutting down, following the tasks) the parent.  It would definitely be possible to build APIs that would provide granular control of the child processes, but again this would mean more new code with a brand new approach.

That all being said, I think this effort is parallel (although related) to the concurrency implementation of other methods (namely `threads`, `gevent`, `asyncio` , and even `multiprocessing`)

- As this is running the full worker, more lightweight concurrency patterns probably wouldn't be well suited: it seems hard to justify running a full blown worker in it's own thread for example, while it could just be throwing jobs to a `Threadpool` with a lot less issues regarding to safety, race conditions and etc.
- This is even more relevant for `asyncio` and `gevent` - one wouldn't have 1000 workers (one for each `greenlet` for example). 

I guess whether the wrapper over the existing worker (`workerpool`) is preferable over new workers with different concurrency models is debatable. I would tend to prefer the later, but I understand if the roadmap pends the other way.


---

> the refactoring has to be done bit by bit, I generally stay away from wholesale changes because that may introduce bugs or instability.

This is a fairly different topic, but a very relevant one nonetheless. I understand your point here, although I might have a slightly different view. I guess my main concern is that the bit by bit favors backwards compatibility over anything else, which greatly reduces flexibility. As this is a +10y codebase, it’s very hard to make it `mypy` friendly for example, and there are choices that I think are worth reviewing - I discussed a few of them on the V2 discussion, but they are mainly related to organization, methods privacy/protection, and architectural decisions. The second relevant aspect is related to speed - the bit by bit approach tends to be a lot slower, changes are a lot more delicate, and slower may mean “not implemented”.

A polished and stable solution necessarily needs time, being exposed to battle and being tested and used in a lot of different scenarios by different people - having “simple” versions out there (beta) is super relevant to get this experience and expertise (think of `ThreadPoolWorker` that doesn’t support job dependencies just yet, or a `GeventWorker` that doesn’t have custom exception handling - just examples).

I tend to favor a path where 1.x is maintained while v2 is being developed in beta. Two good examples of this approach are `psycopg3` and `sqlalchemy2` that just went through (are going through) a significant change, and have been running on alpha/beta mode (gaining experience) for quite a while. This would allow for a lot more flexibility than continuously incrementing 1.x and keeping things backwards compatible, avoiding legacy code for compatibility reasons.

selwin:
> Doesn’t WorkerPool implementation, being a layer of abstraction on top of the worker reduces the flexibility for custom classes, since it hides details that one would want to customize to make other concurrency models work?

It shouldn't, because the worker pool can pass all configuration options to the underlying worker class. 

> That all being said, I think this effort is parallel (although related) to the concurrency implementation of other methods (namely threads, gevent, asyncio , and even multiprocessing)

I agree, I'm open to PRs that rearranges or reworks worker methods to make developing other kinds of concurrency implementation easier. This is something that I want to support. A few people maintained gevent powered workers.

>  I discussed a few of them on the V2 discussion, but they are mainly related to organization, methods privacy/protection, and architectural decisions.

I'm open to re-organizing the codebase to make it more structured, but it has to be done in PR chunks that I can actually review.

I also understand that there's a pattern of using the `_function_name()` to indicate private functions/methods, but honestly I don't care too deeply about this and my approach is to just give backwards compatibility guarantees to documented methods.

>  I guess my main concern is that the bit by bit favors backwards compatibility over anything else, which greatly reduces flexibility. As this is a +10y codebase, it’s very hard to make it mypy friendly for example, and there are choices that I think are worth reviewing - I discussed a few of them on the V2 discussion, but they are mainly related to organization, methods privacy/protection, and architectural decisions. 

I totally agree and realize that I tend to make decisions that prioritizes stability and slow evolution of things.

> The second relevant aspect is related to speed - the bit by bit approach tends to be a lot slower, changes are a lot more delicate, and slower may mean “not implemented”.

Yes, unfortunately this is an unfortunate side effect and I'm aware of this. But I'm in favor of rearranging methods/functions such that improvements can still be done by subclassing RQ classes. 

> I tend to favor a path where 1.x is maintained while v2 is being developed in beta. Two good examples of this approach are psycopg3 and sqlalchemy2 that just went through (are going through) a significant change, and have been running on alpha/beta mode (gaining experience) for quite a while. This would allow for a lot more flexibility than continuously incrementing 1.x and keeping things backwards compatible, avoiding legacy code for compatibility reasons.

This is a good idea but I currently don't have the time and bandwidth to support two different code bases :(.

lowercase00:
@selwin truly appreciate your frankness and openness (and patience :)) on this!
Sure understand your points, I think there's a lot of things that we could do without breaking stuff, so let's stick to that.
I'll keep working on some of the workers, try to fix a few bugs and will take a better look at the pool.

Implementation wise, i was thinking about this over the last couple of days, and I wonder what your thoughts are.
I guess there two main ways to go about concurrency: one is more transparent, and the other is more explicit, I guess there are pros/cons to it.

- Make 100% explicit, more verbose and less "automatic", which means always using the `--worker-class` argument and some general parameters, so `rq worker -w ' rq.worker.NativeConcurrentWorker' --pool-size 10`. The good thing about this is that keeps the API mostly the same, the downside is that is a more "manual" way of providing arguments
- Make it transparent (like Celery), and then we map the implementation to the class , so `rq worker --pool gevent pool-size 10`. The `WorkerPool` implementation is good example of the trade-offs: we could have a new command/concept called `WorkerPool`,  or we could have something like `rq worker --pool prefork --pool-size 5`

I guess the first option is easier to develop, since not many other argument/parms are needed and the architecture remains mostly the same. The second option is more natural/easier for the enduser.

What are your thoughts?

Also, what are the minimum set of features would you consider for a new worker? By limiting the feature set of a new worker (meaning: the threadworker doesn't work with scheduler/callbacks/dependencies) we could greatly reduce the complexity and increase speed of new implementations, which in turn should reduce bugs and unexpected behavior. Another option would be have it implemented, but "unsupported". Apart from the multithreaded worker I have been using a gevent worker for a while, and a couple of months ago I had started an `AsyncWorker` (I left Flask for FastAPI). What are your thoughts on the bare minimum we want to add theses?

lowercase00:
Some more tests with dependencies.
Looking into this a bit further, it seems the challenge is: when running in `burst` mode, the main thread exists before the `parent` job actually finishes executing, so the dependent job never get enqueued.

One way I can think of to fix this: before actually breaking the work loop on the main thread, we could have an array of  tuples with `tuple[job, futures]`, and then check if any job that is currently in a `future` object (eg being executed in a thread) has dependents. If they do, wait until they finish to enqueue the dependents. This would block the main thread, avoid it to `break` the work loop, and allow the `parent` job to enqueue dependents. 

This shouldn't happen when not in burst mode, since there's always going to be next loop cycle, so the `parent` will always have a chance to enqueue the dependent. TBH It sounds too complex, but it's the only idea I have at the moment to try to fix this. Any other ideas are more then welcome.

lowercase00:
>  before actually breaking the work loop on the main thread, we could have an array of tuples with tuple[job, futures], and then check if any job that is currently in a future object (eg being executed in a thread) has dependents

This seems to work - some more polishing needed. Tests are passing locally, but it seems something is wrong on the CI, will need to take a better look at this. I'll mark this as ready for review just so we can work on top of a feedback loop.

@selwin something to keep in mind: the current architecture for tests was made for a single worker, so to tests other workers basically all test suite has to be replicated, lots of room for trouble here. We might want to think of a more dynamic way of testing different worker classes in the future.

selwin:
I haven't had the time to look at this PR closely, for a PR this ambitious, I think we'll need time to iterate.

If you think there's a few chunks of logic that should be broken down into separate functions/methods, let me know. We can make a separate PR to get that function/method into master first so it's easier to maintain this PR. I'll try to take a look at this in a few days.

lowercase00:
@selwin , no worries. It's actually simpler that what it looks like, but obviously take your time and we'll iterate on it. The biggest method is the `work()` which is 90% the same as the original.

> If you think there's a few chunks of logic that should be broken down into separate functions/methods, let me know

TBH I'm not sure how to do this. I'm thinking of having a single class that tests all base workers' methods, and when initializing it pass in the different classes, and then a small class testing specific workers methods. We shouldn't need any PR before reviewing this though.

rq/worker.py
```diff
@@ -1378,3 +1388,199 @@ class RandomWorker(Worker):
 
     def reorder_queues(self, reference_queue):
         shuffle(self._ordered_queues)
+
+
+class ThreadPoolWorker(Worker):
+    death_penalty_class = TimerDeathPenalty
+
+    def __init__(self, *args, **kwargs):
+        self.threadpool_size = kwargs.pop('pool_size', 24)
```
selwin, line 1397: I was just thinking a bit more about this. I think we can bring in attributes like `threadpool_size` and `_idle_threads` used here into the main `Worker` class, so they can be used by other types of workers implementing their own concurrency methods.

The default worker class will simply have `1` for `threadpool_size`, `_idle_threads` can also be incremented and deducted as work is being done.

I'd like to find a better and more generic name for `threadpool_size` though.

This way your work can go into RQ's master branch bit by bit. And if this works and `WorkerPool` is no longer needed, I'm also quite ok with it.

lowercase00: I guess there are a couple of ways to go about this (thinking of multiple workers and future maintainability): (a) a "super" worker class, that does a lot and each worker implements very specific parts of different working models or (b) the minimum-denominator approach, where the base worker implements methods that all workers (regardless of working model) will use.

The current codebase uses strategy (a) - there are quite a few methods on the `Worker` that are not used by the `SimpleWorker` for example.

I would tend to favor strategy (b) - I think smaller classes are easier to reason about, simpler to maintain, and we "always" know that whatever is in the worker class is useful for any other worker, which is one of the main reasons why I think of the worker that forks a `ForkWorker` (we talked a bit about this on the V2 discussion).

No right or wrong here, so a choice of different approaches, both work, happy to adjust accordingly.
--

About the `WorkerPool`, I think that its complementary to a `ThreadWorker`. I think the main decision here is whether this will wrap the whole worker, or maybe just the `pop & work` part of things.

lowercase00:
Update: Even though tests are passing locally they fail in CI because I was missing the `RUN_SLOW_TESTS_TOO` env. It seems that there's an extra heartbeat happening sometime, will tackle this next.

Also opening a new PR to document this for future contributors.

lowercase00:
Update: The test `test_heartbeat_busy` still fails caused by a mismatch in the heartbeat count - we weren't expecting any heartbeats, but got one anyways. Still need to investigate this, couldn't find any obvious reason as to why we're getting the extra heartbeat. The other tests all seem to work work fine.

rq/worker.py
```diff
@@ -1378,3 +1389,213 @@ class RandomWorker(Worker):
...
+                    raise
+
+                except:  # noqa
+                    self.log.error('Worker %s: found an unhandled exception, quitting...', self.key, exc_info=True)
+                    break
+        finally:
+            self.register_death()
+            self.unsubscribe()
+
+        return bool(completed_jobs)
```
selwin, line 1520: @lowercase00 I'd like it so that you don't need to reimplement the whole `work()` method.

Do you mind opening a PR for a method called `worker.bootstrap()` that setups loghandlers to setting up scheduler? That way other workers can just implement the dequeue loop.

Depending on how this PR goes, I want to put in more hooks so that other workers can just implement already existing methods in the base class.

rq/worker.py
```diff
@@ -1378,3 +1389,213 @@ class RandomWorker(Worker):
...
+
+class ThreadPoolWorker(Worker):
+    death_penalty_class = TimerDeathPenalty  # type: ignore
+
+    def __init__(self, *args, **kwargs):
+        self.threadpool_size = kwargs.pop('pool_size', self.default_pool_size)
+        self.executor = ThreadPoolExecutor(max_workers=self.threadpool_size, thread_name_prefix="rq_workers_")
+        self._idle_threads = self.threadpool_size
+        self._lock = threading.Lock()
+        self._current_jobs: List[Tuple['Job', 'Future']] = []  # type: ignore
```
selwin, line 1402: `current_jobs` is also something that we should bring into RQ proper. Mind making a PR for this so we can merge this into master?

rq/worker.py
lowercase00: Done

rq/worker.py
lowercase00: Sure, but the current signature won't make sense. Do you want just the attribute and an empty array as a placeholder?

selwin: So `worker.set_current_job_id()` (and other methods saving job ID) would also save the job ID to `worker._current_jobs`.

The existing `worket.get_current_job()` method would simply take the first job from the array so it remains backward compatible.

lowercase00: Understood. This is not how this method is being used in this case. We may adapt and change the logic for the `setter` and `getter` of the current job on the current worker (there's the question whether this would be the `id` or the job instance). So regardless of whether this is implemented in the current worker or not, the current structure would still be necessary (and overlapping) for this worker (and it seems that for `GeventWorker` as well).

I understand the direction of having a more generic `Worker` class so that new implementations would be easier, but I'd like to propose a "step-by-step" approach where after implementing this (and others) workers and eventually reviewing the current worker mechanics, we may implement improvements - otherwise we are going to be concerned about (1) backwards compatibility (2) compatibility between workers (3) usefulness to other workers - it seems rather early to consider all of this for optimization sake.

selwin: But I think `worker.get_jobs()` needs to be a standard method, since this will be used by monitoring tools like django-rq - (they need to know what job(s) are being worked on by a worker).

lowercase00: For reutilizing components, not sure this makes much of a difference because the different types each concurrency model uses. Shouldn't make much difference to include that though, except the `getter` method for the current worker would be getting data from an internal object vs getting "fresh" data from the server.

Since those APIs don't exists, we might want to have the unified API done through the main `RQ` class (a proposal was shared [here](https://github.com/lowercase00/rq/blob/feature/app/rq/main.md)), this would avoid having to worry about the analytics on the `Workers` themselves and allow for more flexibility. To make new implementations predictable, we might want to write an interface to make sure new workers follow the right implementation. But all of these seem to be a different feature set.

Let me know how you want to proceed with this.

selwin: Hi, looks good, but there are a things that I'd like to change to make it more OOP like. For example I'd switch
```python
w = rq.get_worker("myworker")
# Kill a worker
rq.kill_worker("myworker")
```

To this:
```python
w = rq.get_worker("myworker")
# Kill a worker
w.kill()
```

So next what we can do is define the interface for each `Queue`, `Worker` and `Job`. Out of all these, the `Worker` interface is most relevant for this PR.

lowercase00:
This was the test that was failing.
I’m not entirely sure the logic behind it: it doesn’t expect any `heartbeat`, but when running `execute_job` we call `prepare_job_execution` which has a call to `heartbeat` in it. Tried waiting for all current jobs, having a single thread, and having a replicated `heartbeat` method on the thread pool, but couldn’t make it work - wondering if it should actually work.

```python
q = Queue()
w = ThreadPoolWorker([q], job_monitoring_interval=5)

for timeout, expected_heartbeats in [(2, 0), (7, 1), (12, 2)]:
    job = q.enqueue(long_running_job,
                    args=(timeout,),
                    job_timeout=30,
                    result_ttl=-1)
    with mock.patch.object(w, 'heartbeat', wraps=w.heartbeat) as mocked:
        w.execute_job(job, q)
        w.wait_all()
        self.assertEqual(mocked.call_count, expected_heartbeats)
    job = Job.fetch(job.id)
    self.assertEqual(job.get_status(), JobStatus.FINISHED)

```

rq/worker.py
```diff
@@ -1573,3 +1584,208 @@ class RandomWorker(Worker):
...
+                except SystemExit:
+                    # Cold shutdown detected
+                    raise
+
+                except:  # noqa
+                    self.log.error('Worker %s: found an unhandled exception, quitting...', self.key, exc_info=True)
+                    break
+        finally:
+            self.register_death()
+            self.unsubscribe()
```
selwin, line 1704: These two lines can go into `worker.teardown()`

rq/worker.py
```diff
@@ -1573,3 +1584,208 @@ class RandomWorker(Worker):
...
+                        if not burst:
+                            break
+                        has_pending_dependents = self.__check_pending_dependents()
+                        if has_pending_dependents:
+                            continue
+                        self.log.info("Worker %s: done, quitting", self.key)
+                        break
+
+                    job, queue = result
+                    self.reorder_queues(reference_queue=queue)
```
selwin, line 1677: `reorder_queues()` should also be moved into `dequeue_job_and_maintain_ttl()`, so it can adjust the order of queues before dequeueing.

rq/worker.py
lowercase00: We don't have this method now  - you mean creating a new method on the main worker?

rq/worker.py
lowercase00: @selwin what do you think of opening a brand new PR only for the worker refactoring?
the refactor of the current `Worker` methods being a conditioning factor for a new worker implementation, it seems that it will be more efficient to work on one thing at a time (refactoring it first), otherwise we will be constantly dealing with conflicts and rebasing.

rq/worker.py
lowercase00: Added on #1820

rq/worker.py
lowercase00: > reorder_queues() should also be moved into dequeue_job_and_maintain_ttl(), so it can adjust the order of queues before dequeueing.

`roundrobin` uses a `reference_queue` which comes from `dequeue_job_and_maintain_ttl`

rq/worker.py
lowercase00:
```python
w = rq.get_worker("myworker")

# Here, the worker has a `kill` method.
w.kill()
```

```python
rq = RQ(...)

# Here you're making all interactions available through a general `RQ` class
rq.kill_worker("myworker")
```

rq/worker.py
selwin: We should be able to do this:
```python
if not reference_queue:
    reference_queue = last_queue_in_the_list  # In case of round robin, the shuffling will return the first queue
```

lowercase00:
@selwin want to take another look at this ti ser what other changes you would like?

- Created bootstrap
- Moved reorder queues
- created teardown

selwin:
Yes, I'd like to take a more in depth look when I have more time. Thanks for getting this PR this far, expect more changes as we get this into shape ;)

This will be a very big step forward for RQ.

lowercase00:
Sure no worries, happy to contribute, will have more workers coming soon :)

rq/worker.py
```diff
@@ -1629,3 +1643,208 @@ class RandomWorker(Worker):
...
+    def __change_idle_counter(self, operation: int):
+        """Updates the idle threads counter using a lock to make it safe.
+
+        Args:
+            operation (int): +1 to increment, -1 to decrement.
+        """
+        with self._lock:
+            self._idle_threads += operation
+
+    def __wait_for_slot(self, wait_interval: float = 0.25):
```
selwin, line 1803: Is there some kind of push based mechanism instead of polling for free workers?

rq/worker.py
```diff
@@ -1629,3 +1643,208 @@ class RandomWorker(Worker):
...
+    def __init__(self, *args, **kwargs):
+        self.threadpool_size = kwargs.pop('pool_size', self.default_pool_size)
+        self.executor = ThreadPoolExecutor(max_workers=self.threadpool_size, thread_name_prefix="rq_workers_")
+        self._idle_threads = self.threadpool_size
+        self._lock = threading.Lock()
+        self._current_jobs: List[Tuple['Job', 'Future']] = []  # type: ignore
+        super(ThreadPoolWorker, self).__init__(*args, **kwargs)
+
+    @property
+    def is_pool_full(self):
```
selwin, line 1660: I think it would be better if we change this to `number_of_available_horses()` instead of a simple boolean.

rq/worker.py
```diff
@@ -1629,3 +1643,208 @@ class RandomWorker(Worker):
...
+            wait_interval (float, 0.25): How long to wait between each check. Default to 0.25 second.
+        """
+        while 1:
+            if not self.is_pool_full:
+                self.log.debug('Found idle thread, ready to work')
+                break
+            time.sleep(wait_interval)
+            continue
+
+    def __check_pending_dependents(self) -> bool:
```
selwin, line 1817: I think we need to restructure `work()` in such a way that this is not needed.

selwin:
I need to think further to further group some sections inside `work()` so that you don't need to rewrite so much of it.

On a more meta perspective, since we already have work horse terminology, I think we should stick to that terminology and call the threads `horses`.

rq/worker.py
```diff
@@ -1629,3 +1643,208 @@ class RandomWorker(Worker):
...
+
+class ThreadPoolWorker(Worker):
+    death_penalty_class = TimerDeathPenalty  # type: ignore
+
+    def __init__(self, *args, **kwargs):
+        self.threadpool_size = kwargs.pop('pool_size', self.default_pool_size)
+        self.executor = ThreadPoolExecutor(max_workers=self.threadpool_size, thread_name_prefix="rq_workers_")
+        self._idle_threads = self.threadpool_size
+        self._lock = threading.Lock()
+        self._current_jobs: List[Tuple['Job', 'Future']] = []  # type: ignore
```
selwin, line 1656: We should also bring `_current_jobs` to the main worker class.

rq/worker.py
lowercase00: Not that I could find, the `ThreadPoolExecutor` is a pretty thin wrapper - the only push based mechanics I saw was related to a single future with the `done_callback`.

rq/worker.py
lowercase00: The reason I went this path here is that in this case, things don't happen sequentially, so we need a global way of handling dependencies (thread Y can't work unless thread X has finished) - this is not a challenge on the `Worker` and on the `SimpleWorker`.

lowercase00:
A few comments that are relevant for multiple different review comments:

There are a three main reasons why I don’t think going with `horse` as a generic term for kind of worker might not be the best course of action.

The first, is that this is not obvious: is a thread a horse? How about a worker that runs each job on a separate process (using multiprocess for example)? And a Async worker running an event loop? And a gevent worker? This is historically used for forked processes (specifically using `fork()`) but as we introduce other concepts we would be trying to *fit* those concepts into the existing model, whereas the existing model doesn’t actually support new models.

Harder to maintain: not every worker will have the same concepts/structure as the current. The best example is the `SimpleWorker` which is arguably a lot simpler then the `ForkWorker`, and with the current implementation, what happens is we have 4-6 methods that are not used by the `SimpleWorker`, so we have to go around the code to understand whether a change in X is affecting worker Y (this just happened on that `Windows` thing, where something on the `horse` broke the `SimpleWorker`). Ideally we should be able to safely change a worker without affecting the others. (BTW: this is also one of the reasons why I think in the future it would make more sense to have the `SimpleWorker` the default `Worker` and the `ForkWorker` built on top of it: if the `Worker` class is shared, ideally it should be organized in a way that it only carries the minimum common denominator across all workers’ implementations).

By fitting things from other workers into the main worker, we would most likely cause bloat and less predictability on the current worker implementation - which makes the worker harder to maintain (less freedom to new worker implementations, by “demanding” them to follow the main worker class pattern/signature. Another place where this appears is the `_current_jobs` - the `ThreadWorker` will have to overwrite this anyways - it’s fine to have it on the `ForkWorker` but the signature is different, it means something else depending on the Worker.

Premature optimization: this is the very first implementation of a logically different worker. Fitting everything this new worker uses to the old one, can be seen as optimization at a time where we don’t actually know that much about the requirements for new workers (in my mind a `AsyncWorker` is crucial for those using FastAPI/Starlite for example, and `GeventWorker` is very convenient and lightweight for simple I/O bound tasks). Optimizing is difficult at the stage - by the time we find the `GeventWorker` needs a new concept to wait for dependencies for example, we would be changing the main worker code again - why don’t we just give more freedom for the new workers implementation, and by the time we implement 3-4 workers we can now look back and see that X and Y are attributes that follow a common pattern and we can simplify things by having them on the main `Worker` class?

Another good example is the dependency logic: for a multithread worker, it needs to check for dependencies outside a specific thread - to avoid a thread working on a job when it actually should be waiting for its dependency to finish (on a different thread). This is not a challenge on the `SimpleWorker`, `Worker`, so we would be working hard to change this logic on the main worker, while it doesn’t actually need that, and the logic on the new worker is simple enough and isolated enough in a way that’s simpler to maintain. 

At last the example of changing the `execute_job` logic. This is actually needed for the `ThreadWorker` because jobs are running within a thread, but the current implementation to the `Worker` doesn’t actually *need* a new logic, another case where by trying to fit everything into a super-Worker class we would increase complexity - we would still need to write the `execute_job` anyways for a new worker.

selwin:
> The first, is that this is not obvious: is a thread a horse? How about a worker that runs each job on a separate process (using multiprocess for example)? And a Async worker running an event loop? And a gevent worker? This is historically used for forked processes (specifically using fork()) but as we introduce other concepts we would be trying to fit those concepts into the existing model, whereas the existing model doesn’t actually support new models.

We need to have a term to describe the capability to have multiple processes/threads/worker inside a `Worker`, right? I'm open to suggestions on what we want to call this, maybe we can take inspirations from other similar libraries? So far we have been calling this feature multithreaded worker, but I think a `thread` (it has a very specific meaning in the Python language itself) is even more specific than a `horse`.

> Harder to maintain: not every worker will have the same concepts/structure as the current. The best example is the SimpleWorker which is arguably a lot simpler then the ForkWorker, and with the current implementation, what happens is we have 4-6 methods that are not used by the SimpleWorker, so we have to go around the code to understand whether a change in X is affecting worker Y (this just happened on that Windows thing, where something on the horse broke the SimpleWorker). Ideally we should be able to safely change a worker without affecting the others.

I understand the concern, but in my experience maintaining RQ (and other projects), breaking things into several methods blocks actually helps with the stability of other implementation that subclasses the main worker method. There have been several attempts at making [multi-threaded](https://github.com/zhangliyong/rq-gevent-worker) (for lack of a better term) workers which overrides the entire `work()` method and they never age well. They are never get updated to support the latest features like schedulers, pubsub or any of the new features/bugfixes that get released.

If we were to have multiple worker models in RQ's core library, we'll need to be able to maintain them :).

> BTW: this is also one of the reasons why I think in the future it would make more sense to have the SimpleWorker the default Worker and the ForkWorker built on top of it: if the Worker class is shared, ideally it should be organized in a way that it only carries the minimum common denominator across all workers’ implementations

I agree with you that the `SimpleWorker` model should be the default. If the `Simple` term bugs you, we can rename this to something else and update the docs to make this the default (this doesn't need to wait for v2.0).

> Premature optimization: this is the very first implementation of a logically different worker. Fitting everything this new worker uses to the old one, can be seen as optimization at a time where we don’t actually know that much about the requirements for new workers (in my mind a AsyncWorker is crucial for those using FastAPI/Starlite for example, and GeventWorker is very convenient and lightweight for simple I/O bound tasks). Optimizing is difficult at the stage - by the time we find the GeventWorker needs a new concept to wait for dependencies for example, we would be changing the main worker code again - why don’t we just give more freedom for the new workers implementation, and by the time we implement 3-4 workers we can now look back and see that X and Y are attributes that follow a common pattern and we can simplify things by having them on the main Worker class?

I share your concern, which is why I'm taking this really slowly, one step at a time. It would be good if we can have multiple worker models ready, but I think this will be a big effort so I'm doing what I can based on the implementation that we have.

At the very least, monitoring tools like django-rq (and even RQ's built in CLI) will need several predefined APIs to:
* See how many free/working horses is in a worker
* Show jobs currently worked on by a worker

> Another good example is the dependency logic: for a multithread worker, it needs to check for dependencies outside a specific thread - to avoid a thread working on a job when it actually should be waiting for its dependency to finish (on a different thread). 

This needs to change. A worker in `burst` model should only listen for jobs and work until there's nothing left on the queue, it shouldn't need other dependency logic.

lowercase00:
> We need to have a term to describe the capability to have multiple processes/threads/worker inside a `Worker`, right?

To be honest I don't really see any need for that right now. I do understand this may be relevant in the future (mostly for monitoring purposes), but for now I was just thinking of calling the processing unit of that specific worker as what it actually is, a thread. This is important anyways, so we can have any single job-processing unit as a `horse` or any other name then.

> I understand the concern, but in my experience maintaining RQ (and other projects), breaking things into several methods blocks actually helps with the stability of other implementation that subclasses the main worker method. 

Not sure I follow, what I mean was that if horse-related methods are not used by everyone, would make more sense to have the "base" worker (the one everyone inherits from) as thin as possible. In practice that would mean switching: the `SimpleWorker` would become `Worker`, and the current `Worker` would become `ForkWorker`. That was the minimum-denominator logic I mentioned. This would help Windows folks as well, since the default would then just work on Windows.

> There have been several attempts at making [multi-threaded](https://github.com/zhangliyong/rq-gevent-worker) (for lack of a better term) workers which overrides the entire `work()` method and they never age well.

To be honest I never quite understood why, I personally used a slightly updated version of  the `GeventWorker` in production quite a lot, and the logic is actually very similar to this `ThreadPoolWorker`. Hopefully this will be a PR soon. The main challenge was the import / patch logic. The most obvious way to fix would be to just use an `envvar`, but that's a pretty inconvenient solution, so I needed to dig deeper into other options. I can push a draft so we can look at it as well.

> They are never get updated to support the latest features like schedulers, pubsub or any of the new features/bugfixes that get released.

That's true, however, I also think it's fair that we accept that not necessarily all features will be implemented to all workers every time, specially during the first iterations - Celery is good benchmark here, on their docs they explicitly mention which `pool` is supported for each specific feature.

> At the very least, monitoring tools like django-rq (and even RQ's built in CLI) will need several predefined APIs
> * See how many free/working horses is in a worker
> * Show jobs currently worked on by a worker

That's a perfect approach, but arguably a really high standard for the very first iteration of the first Worker that implements concurrency to have to implement generic monitoring capabilities. But let's pursuit it, it's definitely a great ambition to have. 

> This needs to change. A worker in `burst` model should only listen for jobs and work until there's nothing left on the queue, it shouldn't need other dependency logic.

You mean the `__check_pending_dependents`? Sure let's look at this logic, that was a solution I found since multiple tests were failing - since the work was not sequential anymore, one thread would start before it's dependencies were actually finished.

Let me get back to your specific review comments so we can make some progress and iterate.

selwin:
> Not sure I follow, what I mean was that if horse-related methods are not used by everyone, would make more sense to have the "base" worker (the one everyone inherits from) as thin as possible. In practice that would mean switching: the SimpleWorker would become Worker, and the current Worker would become ForkWorker. That was the minimum-denominator logic I mentioned. This would help Windows folks as well, since the default would then just work on Windows.

This is a good idea, we can refactor things out to a `BaseWorker` class and make `Worker` and `SimpleWorker` subclass that.

> That's a perfect approach, but arguably a really high standard for the very first iteration of the first Worker that implements concurrency to have to implement generic monitoring capabilities. But let's pursuit it, it's definitely a great ambition to have.

Well, nothing prevents other people from writing their own `Worker` implementation outside of RQ, it has already been done before. But when we actually bring a different type of worker inside the core library itself, it would make sense that it conforms to some sort of common protocol, right?

That's what I'm trying to figure out now, bit by bit :).

selwin:
> You mean the __check_pending_dependents? Sure let's look at this logic, that was a solution I found since multiple tests were failing - since the work was not sequential anymore, one thread would start before it's dependencies were actually finished.

Yes, we should keep the logic simple:
1. `worker` waits on the queue (with X seconds of timeout) when there are processing units (for lack of a better term) available
2. After successful execution, processing unit enqueues other jobs (which will be picked up and distributed to other processing units)
3. If there's no other jobs within X seconds and if `burst=True`, the worker exits.

selwin:
I took a more thorough look on the implementation so far. My feedbacks:
1. `worker._current_jobs` is implemented nicely, we should bring this into RQ and should be made into public worker APIs.
2. I also like the `job_done()` callback approach, this should also be brought into other workers.
3. I think we can remove `_idle_threads` though, the number of idle threads can be computed by substracting the number of jobs from pool size

lowercase00:
> This is a good idea, we can refactor things out to a BaseWorker class and make Worker and SimpleWorker subclass that.

Great, we can start looking at this, in a new PR, leaving all common methods to the `Base` and having specific implementations. I think the biggest decision is what is the `Worker` class, since it's the default (I would still favor having the `Simple` as the default, but I think it would be possible on V2, since its a big behavioural change, even though not breaking. 

> it would make sense that it conforms to some sort of common protocol

Most definitely yes, my point here is that we could implement the protocol step by step, mainly because we are going to learn other things when implementing new workers. So my idea was to do that incrementally, instead of trying to figure the full protocol right now.

> 1. worker waits on the queue (with X seconds of timeout) when there are processing units (for lack of a better term) available
> 2. After successful execution, processing unit enqueues other jobs (which will be picked up and distributed to other processing units)
> 3. If there's no other jobs within X seconds and if burst=True, the worker exits.

I don't remember exactly why I chose that specific implementation. From what I can remember, with a concurrent worker, there was a scenario where two threads were working in parallel, and they actually shouldn't (because of the dependency). Let me take a look at this again, need to remember the dependency/queueing/deferring logic.


> worker._current_jobs is implemented nicely, we should bring this into RQ and should be made into public worker APIs.

Sure, I'll work on that on a separate PR then.

> I also like the job_done() callback approach, this should also be brought into other workers.

Ideas on this implementation? The way this works for the thread is native, since the callback is attached to the `Future` object, so for other workers we would have to do that "manually", maybe on a `handle_finished_job` or something?

> I think we can remove _idle_threads though, the number of idle threads can be computed by substracting the number of jobs from pool size

That's fine also, I don't see the downside of having the info pre calculated (thinking of monitoring), but not really that relevant, so we can remove that as well.

selwin:
> Most definitely yes, my point here is that we could implement the protocol step by step, mainly because we are going to learn other things when implementing new workers. So my idea was to do that incrementally, instead of trying to figure the full protocol right now.

Yes, I think we're aligned on this one. We're already doing this in a step by step fashion, as this PR continues to evolve :). We have three different worker models:
1. InProcess
2. Forking
3. Threaded

We'll try to sketch out the protocols based on these three models and evolve them as needed. The `_current_jobs` is a new one that we need because `ThreadedWorker` can work on multiple jobs in parallel.

> Ideas on this implementation? The way this works for the thread is native, since the callback is attached to the Future object, so for other workers we would have to do that "manually", maybe on a handle_finished_job or something?

Yes, I think `handle_job_completion()` is the next logical thing to implement. It may also absorb some of the common operations performed by `handle_job_failure()` and `handle_job_success()`.

Maybe it's signature can be `handle_job_completion(job, queue, started_job_registry, future)`? I'm still unsure about the `future` bit, but the idea is that it's an object that contains information about the processing unit. Maybe even a `Horse` class that has these attributes:
* `future`: for future based concurrency models
* `process_id`: for fork() and spawn() based models - these are also valid concurrency options

selwin:
As an aside, if we can land this PR, [ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor) has very similar APIs so we'll actually get `Process` based concurrency in addition to `Thread` based one 😀.

lowercase00:
> As an aside, if we can land this PR, [ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor) has very similar APIs so we'll actually get Process based concurrency in addition to Thread based one 😀.

Exactly, that was one of the reasons I went with the `Executor` instead of the `threading` module directly. The interface is the same. I had wrote a `MultiprocessWorker`, but it wasn't working because of the `if __name__ == '__main__'` thing and I didn't spend time figuring that up.

And actually the logic is very similar for a `GeventWorker` as well, or even if want to do `fork-based` concurrency. So it is a good framework for thinking of concurrency indeed.

selwin:
Just so that we don't work on the same things, I'm working on bringing `worker._current_jobs` to `Worker` class.

selwin:
> Just so that we don't work on the same things, I'm working on bringing worker._current_jobs to Worker class.

Unless you're already working on this, let me know. I'll work on something else.

danielbprice:
It would be really nice to have this.  We have been trying to use `rq` with Tensorflow, and Tensorflow's internal thread pools really seem to get screwed up by the forking that RQ wants to do, and subsequent calls into Tensorflow hang/deadlock.  There are some workarounds you can do to postpone TF's threadpool initialization, but the net result is that we're having to load our models post-fork, which is really a performance killer.  I'm willing to endure "it's Beta" if I could have a multithreaded worker.

docs/docs/workers.md
```diff
@@ -447,3 +447,74 @@ redis = Redis()
...
+This should be enough in most cases, but you can always customize the pool size according to your specific needs by using the `pool_size` param to the worker.
+
+```python
+w = ThreadPoolWorker([q], connection=q.connection, pool_size=12)
+```
+
+There are relevant limitations to the `ThreadPoolWorker`, here's a non extensive list:
+
+- Dependencies might not behave like expected
+- Job registries may not be reliable (specially with jobs that were timed out)
```
gyandeeps, line 518: Biggest thing i like about `rq` is its job reliability as it maintains different job registries correctly. Is it possible to have the same level of reliability with thread pool worker?
we process kafak queue and hand create rq task to process those events. Reliability is super important as we dont want to loose task without knowing about it.

ccrvlh: Thanks for the feedback, definitely something to take into account. I've been working on a different branch since there were other relevant changes to allow a more robust design, will work on this as well. I do imagine that there will be limitations though, specially on the first couple of versions. Hopefully nothing that will hurt reliability... we'll see.

gyandeeps: Appreciate it @ccrvlh . We are in the works to switch from celery to rq for the reliability factor.
btw, i pulled this branch down and tried the threadpool with task timing out but was able to get to consistently get a retry on it. Any thing specific you can point out where you have seen this issue? (just curious)

ccrvlh: It's been a while since I worked on this... I'm mostly working on the v2 branch now (which has this worker setup), but still a lot to do. I'll try to make some tests with it this weekend.

gyandeeps: if you can share the link of the branch i would appreciate it. Just want to see and learn. I tried searching on this repo but was not able to find it.

ccrvlh: Sure! It's on my fork only, you might be able to find it from this repo (or from my profile), I'm on mobile, but I'll post the link here when I get to the PC.

gyandeeps: @ccrvlh just curious on the timeline for this feature or in general V2 since we want to rollout rq in prod but multithread is critical to use. thanks

gyandeeps: also can we try to release this with 2.0 under experimental flag. Based on my tests things looked good.
cc @selwin

selwin: @gyandeeps like I mentioned, I'll only merge this in slowly, one bit at a time to make sure we don't mess anything up.

RQ version 2.0 also changed how jobs are enqueued so this branch is probably stale.

legout:
Hi guys, 

are there any updates here?

PR #1806, opened 2023-02-10
New dequeue strategy

### LINKED ISSUE #1805: Make dequeueing strategy work regardless of the worker
We current have `RoundRobin` and `Random` workers. The implementation is very straightforward, and actually just changes the `reorder_queues` method (which doesn't do anything on the default worker). As this is actually reordering the queue, we could have this work on any worker by just passing a parameter with the `dequeue_strategy` which is a string that can be `fifo` (the default) `random` or `roundrobin`. We would then remove the `RoundRobin` and `Random` workers.

What we can do today:
1. Add the `dequeue_strategy` feature to the worker's `__init__` and `work()` methods.
2. Add deprecation warnings to both the `RoundRobin` and `Random` worker

Work on #1806

lowercase00:
This implements a new parameter `dequeue_strategy` that should replace the `RoundRobinWorker` and `RandomWorker`.
It adds the parameter to (1) the `work()` method, (2) the `Worker` constructor and (3) to the CLI. It also adds deprecation warnings to the workers: I chose to do this through the CLI to avoid having to have a `__init__` method on the workers, and to avoid showing this warning only when the workers hits the `reorder_queue` method - that way we can have the warning as soon as possible.

Fixes #1805

### DIFF
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -27,7 +27,7 @@ from .exceptions import TimeoutFormatError
 logger = logging.getLogger(__name__)
 
 
-class DequeueStrategy(Enum):
+class DequeueStrategy(str, Enum):
     DEFAULT = "default"
     ROUNDROBIN = "roundrobin"
     RANDOM = "random"
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -12,6 +12,7 @@ import logging
 import numbers
 import sys
 import datetime as dt
+from enum import Enum
 from collections.abc import Iterable
 from typing import TYPE_CHECKING, Dict, List, Optional, Any, Callable, Tuple, Union
 
@@ -24,6 +25,12 @@ from redis.exceptions import ResponseError
 from .exceptions import TimeoutFormatError
 
 logger = logging.getLogger(__name__)
+
+
+class DequeueStrategy(str, Enum):
+    Order = "default"
+    RoundRobin = "roundrobin"
+    Random = "random"
 
 
 class _Colorizer:
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -27,10 +27,10 @@ from .exceptions import TimeoutFormatError
 logger = logging.getLogger(__name__)
 
 
-class DequeueStrategy(str, Enum):
-    Order = "order"
-    RoundRobin = "roundrobin"
-    Random = "random"
+class DequeueStrategy(Enum):
+    DEFAULT = "default"
+    ROUNDROBIN = "roundrobin"
+    RANDOM = "random"
 
 
 class _Colorizer:
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -103,7 +103,6 @@ def make_colorizer(color: str):
 
 
 class ColorizingStreamHandler(logging.StreamHandler):
-
     levels = {
         logging.WARNING: make_colorizer('darkyellow'),
         logging.ERROR: make_colorizer('darkred'),
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -12,7 +12,6 @@ import logging
 import numbers
 import sys
 import datetime as dt
-from enum import Enum
 from collections.abc import Iterable
 from typing import TYPE_CHECKING, Dict, List, Optional, Any, Callable, Tuple, Union
 
@@ -25,12 +24,6 @@ from redis.exceptions import ResponseError
 from .exceptions import TimeoutFormatError
 
 logger = logging.getLogger(__name__)
-
-
-class DequeueStrategy(str, Enum):
-    DEFAULT = "default"
-    ROUNDROBIN = "roundrobin"
-    RANDOM = "random"
 
 
 class _Colorizer:
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -28,7 +28,7 @@ logger = logging.getLogger(__name__)
 
 
 class DequeueStrategy(str, Enum):
-    Order = "default"
+    Order = "order"
     RoundRobin = "roundrobin"
     Random = "random"
 
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -97,7 +97,7 @@ def signal_name(signum):
 
 class DequeueStrategy(str, Enum):
     DEFAULT = "default"
-    ROUNDROBIN = "roundrobin"
+    ROUND_ROBIN = "round_robin"
     RANDOM = "random"
 
 
@@ -692,7 +692,7 @@ class Worker:
         if self._dequeue_strategy is None:
             self._dequeue_strategy = DequeueStrategy.DEFAULT
 
-        if self._dequeue_strategy not in ["default", "random", "roundrobin"]:
+        if self._dequeue_strategy not in ["default", "random", "round_robin"]:
             self.log.warning(
                 "Dequeue strategy %s is not allowed. Use one of `default`, `random` or `rounbrobin`. Using defalt ordering.",
                 self._dequeue_strategy,
@@ -700,7 +700,7 @@ class Worker:
             return
         if self._dequeue_strategy == DequeueStrategy.DEFAULT:
             return
-        if self._dequeue_strategy == DequeueStrategy.ROUNDROBIN:
+        if self._dequeue_strategy == DequeueStrategy.ROUND_ROBIN:
             pos = self._ordered_queues.index(reference_queue)
             self._ordered_queues = self._ordered_queues[pos + 1:] + self._ordered_queues[: pos + 1]
             return
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -693,7 +693,7 @@ class Worker:
 
         if self._dequeue_strategy not in ("default", "random", "round_robin"):
             raise ValueError(
-                f"Dequeue strategy {self._dequeue_strategy} is not allowed. Use one of `default`, `random` or `round_robin`. Using defalt ordering."
+                f"Dequeue strategy {self._dequeue_strategy} is not allowed. Use `default`, `random` or `round_robin`."
             )
         if self._dequeue_strategy == DequeueStrategy.DEFAULT:
             return
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -292,7 +292,7 @@ class Worker:
         self.scheduler: Optional[RQScheduler] = None
         self.pubsub = None
         self.pubsub_thread = None
-        self._dequeue_strategy = dequeue_strategy
+        self._dequeue_strategy: DequeueStrategy = dequeue_strategy
 
         self.disable_default_exception_handler = disable_default_exception_handler
 
@@ -681,7 +681,7 @@ class Worker:
             self.pubsub.unsubscribe()
             self.pubsub.close()
 
-    def reorder_queues(self, reference_queue: 'Queue'):
+    def reorder_queues(self, reference_queue: Optional['Queue'] = None):
         """Reorder the queues according to the strategy.
         As this can be defined both in the `Worker` initialization or in the `work` method,
         it doesn't take the strategy directly, but rather uses the private `_dequeue_strategy` attribute.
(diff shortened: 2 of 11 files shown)


rq/utils.py
```diff
@@ -26,6 +27,12 @@
 logger = logging.getLogger(__name__)
 
 
+class DequeueStrategy(str, Enum):
+    Order = "order"
```
selwin, line 31: Let's follow Python's enum convention of using all caps for the options. `DequeueStrategy.ORDER`.

Having said that, just wondering if you have an alternative name for `DequeueStrategy.ORDER`?

rq/utils.py
```diff
@@ -26,6 +27,12 @@
 logger = logging.getLogger(__name__)
 
 
+class DequeueStrategy(str, Enum):
```
selwin, line 30: Can we just use plain `Enum`? No need to subclass from string.

selwin: Also, can we move `DequeueStrategy` to `constants.py`?

rq/utils.py
lowercase00: > Let's follow Python's enum convention of using all caps for the options. DequeueStrategy.ORDER.
Sure.

> wondering if you have an alternative name for DequeueStrategy.ORDER
I didn't like that name also tbh, but out of ideas on what to use. Though if leaving it as `None`, but it wouldn't be as obvious. Maybe `SEQUENCE` or something?

rq/utils.py
lowercase00: > Can we just use plain Enum? No need to subclass from string.

The reason for subclassing is to avoid having to instantiate the `Enum`, and just use a string directly when using as an input - same for output, no need to use `.value` (I think `mypy` also complains about the raw `Enum`). I can make that change anyways.

> Also, can we move DequeueStrategy to constants.py?

Sure, agree it would be a better place. We don't have a `constants` yet though - if we were to add this, it would probably make sense to concentrate other constants as well for consistency. My suggestion would be to make this change in a new PR.

rq/utils.py
lowercase00: @selwin suggested `DEFAULT` on the new version, let me know if you have any other suggestions.

rq/utils.py
lowercase00: Plain Enum: ok. Let me know if we should adopt the `constants` now or on the next iteration.

lowercase00:
I guess something broke with the last changes, will move it back to draft to check that.

rq/utils.py
selwin: > Let me know if we should adopt the constants now or on the next iteration.

Actually you're right, we don't yet have `constants.py`, what we have is `defaults.py` which is not the right place to put this.

Maybe just move this into `worker.py`?

lowercase00: done

rq/worker.py
```diff
@@ -1569,7 +1603,7 @@ class RoundRobinWorker(Worker):
 
     def reorder_queues(self, reference_queue):
```
selwin, line 1604: Can we also move `reorder_queues` from line 828 into `dequeue_job_and_maintain_ttl()`

rq/worker.py
```diff
@@ -668,20 +681,39 @@ def unsubscribe(self):
...
+        """Reorder the queues according to the strategy.
+        As this can be defined both in the `Worker` initialization or in the `work` method,
+        it doesn't take the strategy directly, but rather uses the private `_dequeue_strategy` attribute.
 
         Args:
-            reference_queue (Union[Queue, str]): The queue
+            reference_queue (Union[Queue, str]): The queues to reorder
         """
-        pass
+        if self._dequeue_strategy is None:
```
selwin, line 692: I think we can skip this check because `_dequeue_strategy` is already `DequeueStrategy.DEFAULT` by default.

rq/worker.py
```diff
@@ -280,6 +292,7 @@ def __init__(
         self.scheduler: Optional[RQScheduler] = None
         self.pubsub = None
         self.pubsub_thread = None
+        self._dequeue_strategy = dequeue_strategy
```
selwin, line 295: Can we add type hint here?

rq/worker.py
lowercase00: Sure

rq/worker.py
```diff
@@ -668,20 +681,39 @@ def unsubscribe(self):
...
 
         Args:
-            reference_queue (Union[Queue, str]): The queue
+            reference_queue (Union[Queue, str]): The queues to reorder
         """
-        pass
+        if self._dequeue_strategy is None:
+            self._dequeue_strategy = DequeueStrategy.DEFAULT
+
+        if self._dequeue_strategy not in ["default", "random", "roundrobin"]:
```
selwin, line 695: I'm ok with not doing anything if `DequeueStrategy` is not round robin or random.

rq/worker.py
lowercase00: It is, but for any given reason `None` could be passed, I don't see the downside of going the safe path here and checking the status.

rq/worker.py
lowercase00: Sure

rq/worker.py
```diff
@@ -90,6 +95,12 @@ def signal_name(signum):
         return 'SIG_UNKNOWN'
 
 
+class DequeueStrategy(str, Enum):
+    DEFAULT = "default"
+    ROUNDROBIN = "roundrobin"
```
selwin, line 100: `ROUND_ROBIN` and `round_robin` :)

rq/worker.py
lowercase00: ok

rq/worker.py
lowercase00: ok

rq/worker.py
selwin: Ok, can we at least change the iterable from list to tuple? It performs better https://stackoverflow.com/a/22140115

rq/worker.py
lowercase00: Double checking, we don't do anything, there's just a warning here, since it means someone is using an argument that is not known, so it won't work as the user expected. I would favor keeping the log. Let me know if there are any further actions here.

rq/worker.py
lowercase00: I think the way the test work right now depends on when this is called, so when moving to the dequeue method, tests break. Since this is a refactor, and not related to the feature itself, my suggestion is to work on this on a new PR, we will have to chose how to handle the not having the `reference_queue` 

_(I think you mentioned doing something like `self._ordered_queues[-1:][0]` to get the last element, but this would possibly raise an `IndexError`, so we would have to change the signature to make for a safe flow._

rq/worker.py
lowercase00: Marking as resolved, let me know if you'd like to change this.

rq/worker.py
selwin: Ok, in this case no need to pass `dequeue_strategy` to `bootstrap()` just to be assigned to `self` here: https://github.com/rq/rq/pull/1806/files#diff-50984f276e8a6fea5f4ba4938afb6f4fbc544fd3751fbd16027dc0a326d6405eR734

Just assign `._dequeue_strategy` when `work` is called.

lowercase00: Hmm, not sure I follow the reasoning here - Isn't this use case exactly why the `bootstrap` method was created for - initial tasks that are generic enough that any worker would use?

rq/worker.py
```diff
@@ -668,20 +681,39 @@ def unsubscribe(self):
...
-            reference_queue (Union[Queue, str]): The queue
+            reference_queue (Union[Queue, str]): The queues to reorder
         """
-        pass
+        if self._dequeue_strategy is None:
+            self._dequeue_strategy = DequeueStrategy.DEFAULT
+
+        if self._dequeue_strategy not in ("default", "random", "round_robin"):
+            self.log.warning(
+                "Dequeue strategy %s is not allowed. Use one of `default`, `random` or `rounbrobin`. Using defalt ordering.",
```
selwin, line 697: `rounbrobin` should be `round_robin`

lowercase00: Ops, it's misspelled so the search didn't find it. Fixed.

rq/worker.py
```diff
@@ -668,20 +681,39 @@ def unsubscribe(self):
...
         Args:
-            reference_queue (Union[Queue, str]): The queue
+            reference_queue (Union[Queue, str]): The queues to reorder
         """
-        pass
+        if self._dequeue_strategy is None:
+            self._dequeue_strategy = DequeueStrategy.DEFAULT
+
+        if self._dequeue_strategy not in ("default", "random", "round_robin"):
+            self.log.warning(
```
selwin, line 696: Hey if we're checking this, I think we should just raise an exception. Mind adding this scenario to tests also?

lowercase00: Not sure this scenario is critical enough to interrupt execution, I think logging at `warning` level is enough in this case.

selwin: If there's a check here and the value is not expected, we should raise an exception :).

`Errors should never pass silently.` - Zen of Python ;)

rq/worker.py
```diff
@@ -233,6 +244,7 @@ def __init__(
         disable_default_exception_handler: bool = False,
         prepare_for_work: bool = True,
         serializer=None,
+        dequeue_strategy=DequeueStrategy.DEFAULT,
```
selwin, line 247: What do you think of moving this argument to `work()`? `__init__` is starting to have too many kwargs to my liking.

rq/worker.py
lowercase00: At this point we would be discussing whether a `warning` is "silent", but let's move forward on this - which kind of exception would you prefer here?

rq/worker.py
lowercase00: I don't particularly care about that many `kwargs`, since I feel it delivers a better DX with low marginal cost, but regardless of the `kwargs` count, I think it makes more sense to have this at `work()` indeed, a worker shouldn't really care about which strategy to use except when actually working

selwin: 👍

rq/worker.py
selwin: Just a regular `ValueError` would do.

rq/worker.py
lowercase00: While looking at this just noticed it was already on `work()`. The reason I had it on `__init__` was to make it flexible for the user to chose when/where to use. I'll delete anyway to reduce the number of  kwargs.

rq/worker.py
```diff
@@ -693,11 +692,9 @@ def reorder_queues(self, reference_queue: 'Queue'):
             self._dequeue_strategy = DequeueStrategy.DEFAULT
 
         if self._dequeue_strategy not in ("default", "random", "round_robin"):
-            self.log.warning(
-                "Dequeue strategy %s is not allowed. Use one of `default`, `random` or `round_robin`. Using defalt ordering.",
-                self._dequeue_strategy,
+            raise ValueError(
+                f"Dequeue strategy {self._dequeue_strategy} is not allowed. Use one of `default`, `random` or `round_robin`. Using defalt ordering."
```
selwin, line 696: Typo at `defalt`.
 
We can also make it a bit shorter:
```
dequeue_strategy {self._dequeue_strategy} is not allowed. Use `default`, `random` or `round_robin`.
```

rq/worker.py
```diff
@@ -699,6 +728,7 @@ def bootstrap(
         self.subscribe()
         self.set_state(WorkerStatus.STARTED)
         qnames = self.queue_names()
+        self._dequeue_strategy = dequeue_strategy
```
selwin, line 731: Sorry, one hopefully last review, `dequeue_strategy` doesn't need to be passed into `bootstrap()` as we're not doing anything with this here. We can simply assign `dequeue_strategy` to `self` inside `work()`.

lowercase00: No worries, we discussed a bit of this [here](https://github.com/rq/rq/pull/1806#discussion_r1111508285) and [here](https://github.com/rq/rq/pull/1806#discussion_r1111553095), but you didn't follow up. I'll make the change then.

selwin:
Thanks!

PR #1816, opened 2023-02-14
feat: splits the work loop into a bootstrap method

lowercase00:


rq/worker.py
```diff
@@ -711,34 +767,10 @@ def work(
...
+        self.bootstrap(logging_level, date_format, log_format)
         completed_jobs = 0
-        self.register_birth()
-        self.log.info("Worker %s: started, version %s", self.key, VERSION)
-        self.subscribe()
-        self.set_state(WorkerStatus.STARTED)
-        qnames = self.queue_names()
-        self.log.info('*** Listening on %s...', green(', '.join(qnames)))
-
         if with_scheduler:
```
selwin, line 772: I think this can also go into `worker.bootstrap()`.

lowercase00: You mean the `if with_scheduler`?
May I suggest to keep it separated? The starting the scheduler is a task (method) by itself, so we have clear separation of concerns. This is very cheap to leave (2 extra lines of code), and we allow for other workers to chose when/how/whether to implement the scheduler.

selwin: Ok, I'll merge this in first and see how it goes.

PR #1824, opened 2023-02-19
Add failure callback call to started job registry cleanup

ronlut:
At the moment, the failure callback is not called, which is problematic when you need to do something when the job fails.
This PR fixes it.

@selwin opened as a draft, open for discussion if you have a better idea


rq/exceptions.py
```diff
@@ -30,3 +30,7 @@ def __init__(self, msg, extra_info):
 
 class TimeoutFormatError(Exception):
     pass
+
+
+class JobExpiryError(Exception):
```
selwin, line 35: I'm debating whether `Job Expiry` is descriptive for this. This error happens because a job is left on the started job registry past it's supposed timeout, right? Any other ideas?

ronlut: Kind of, yeah. And if there are no retries left.
Maybe the failure callback should be called even if there are retries left? As happens on real job failure and no on registry cleanup?

Re:naming, I can call it I UnfinishedJobError or AbandonedJobError, I like the 2nd one more

selwin: Yeah, I'm leaning towards `JobTimeoutError` (the execution has timed out) or `AbandonedJobError`. What do you think? This error needs to be documented and explained.

ronlut: Renamed, added to docs and calling the failure_callback always - whether there is a retry or not

rq/registry.py
```diff
@@ -234,7 +251,7 @@ def cleanup(self, timestamp: Optional[float] = None):
 
                     else:
                         job.set_status(JobStatus.FAILED)
-                        job._exc_info = "Moved to FailedJobRegistry at %s" % datetime.now()
+                        job._exc_info = "Moved to FailedJobRegistry, due to job expiry, at %s" % datetime.now()
```
selwin, line 254: Can we update this message to specifically mention `AbandonedJobError`? Also please see https://github.com/rq/rq/pull/1840 to see how the docs can be improved.

luto:
Thanks for the extended handling of dropped jobs! If there is anything I can do to get this merged, please ping me :)

rq/registry.py
ronlut: Done, regarding docs, @luto WDYT?

rq/registry.py
```diff
@@ -202,6 +210,7 @@ class StartedJobRegistry(BaseRegistry):
     """
 
     key_template = 'rq:wip:{0}'
+    death_penalty_class = UnixSignalDeathPenalty
 
     def cleanup(self, timestamp: Optional[float] = None):
         """Remove expired jobs from registry and add them to FailedJobRegistry.
```
luto, line 216: Maybe change this to abandoned too? Then we've got the same naming everything 🤩

rq/registry.py
luto: Looks good to me! All messages and the docs now call those jobs "abandoned", which is fitting imho. The most important thing for me is that you can find the docs based on the message. Both talk about `AbandonedJobError`, so that works. Thanks! :)

ronlut: Thank you 🤝

rq/registry.py
ronlut: Done!

rq/registry.py
```diff
@@ -226,15 +235,26 @@ def cleanup(self, timestamp: Optional[float] = None):
                     except NoSuchJobError:
                         continue
 
+                    if job.failure_callback:
+                        try:
+                            with self.death_penalty_class(CALLBACK_TIMEOUT, JobTimeoutException, job_id=job.id):
+                                job.failure_callback(job, self.connection,
+                                                     AbandonedJobError, AbandonedJobError(), traceback.extract_stack())
+                        except:  # noqa
+                            logger.exception('Registry %s: error while executing failure callback', self.key)
```
selwin, line 244: Ideally I'd like to refactor this to `job.execute_failure_callback()` so we can have on method that handles execution of failure callbacks, currently it's also used here: https://github.com/rq/rq/pull/1824/files#diff-50984f276e8a6fea5f4ba4938afb6f4fbc544fd3751fbd16027dc0a326d6405eR1382 

Do you feel like refactoring this in this PR or leave it for later?

selwin:
Also, do you mind updating this PR with recent commits from master? It doesn't look like Github Actions is functioning properly, it's not running any test.

rq/registry.py
ronlut: Will do with pleasure, but a different PR will be better I think.
Started working on it - will open when done.

ronlut: @selwin see those 2 commits (I rebased on this branch so this is the diff)
https://github.com/rq/rq/pull/1851/files/2241493144c82727d5c56f7d56c7ff78642e0608..4bfae0bdc248632837d43166b145db519e6f81a4

selwin:
Thanks!

ronlut:
@selwin is a new release planned? Waiting for this change :)

selwin:
Possibly sometime in the two weeks or so. I'm polishing up https://github.com/rq/rq/pull/1874 , it's close to being done.

PR #1827, opened 2023-02-20
feat: connection max retries

ccrvlh:


rq/worker.py
```diff
@@ -235,7 +235,8 @@ def __init__(
         disable_default_exception_handler: bool = False,
         prepare_for_work: bool = True,
         serializer=None,
-        work_horse_killed_handler: Optional[Callable[[Job, int, int, resource.struct_rusage], None]] = None
+        work_horse_killed_handler: Optional[Callable[[Job, int, int, resource.struct_rusage], None]] = None,
+        connection_max_retries: int = -1
```
selwin, line 239: I think this should be made an argument for `work()` similar to `burst`.

PR #1835, opened 2023-02-23
docs: scheduler safe import

lowercase00:


docs/docs/scheduling.md
```diff
@@ -95,10 +95,11 @@ from rq import Worker, Queue
 from redis import Redis
 
 redis = Redis()
-
 queue = Queue(connection=redis)
-worker = Worker(queues=[queue], connection=redis)
-worker.work(with_scheduler=True)
+
+if __name__ == '__main__':
```
selwin, line 100: I think we shouldn't have this check here, since we're talking about running the worker programatically (this could be done via a Python shell.

This can be added in the next section, we can talk about adding this to write a Python script that can be run via CLI.

selwin:
Thanks!

PR #1837, opened 2023-02-24
Check dependencies when enqueue via Queue.enqueue_job()

### LINKED ISSUE #1404: Job is executed before "depends_on" job is finished when using Job.create and q.enqueue_job
Consider the following snippet below and two running workers:
```python
r = Redis(
    host="localhost",
    port=6379,
    db=0
)

q = Queue(connection=r)

id_1 = str(uuid4())

q.enqueue(func_1, job_id=id_1)
q.enqueue(func_2, depends_on=id_1)
```
In that snippet, everything works as expected; both jobs are enqueued ( one on each worker ) but func_2 does not run before func_1 is finished. 

Now consider the following snippet, still with two workers:
```python
r = Redis(
    host="localhost",
    port=6379,
    db=0
)

q = Queue(connection=r)

id_1 = str(uuid4())
id_2 = str(uuid4())

job_1 = Job.create(func_1, id=id_1, connection=r) 
job_2 = Job.create(func_2, id=id_2, depends_on=id_1, connection=r)
q.enqueue_job(job_1)
q.enqueue_job(job_2)
```
When I run that code both jobs are started immediately, despite setting the `depends_on` argument. 

<br>
<br>

#### Mini bonus question
Why is it necessary to set `connection=` on the jobs when I use `Job.create`? Doesn't the `Queue` instance know what Redis connection to use already? Code fails without me setting `connection` on all the jobs as well.

sim6:
Fixes #1404

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -660,12 +660,7 @@ class Queue:
             on_success=on_success,
             on_failure=on_failure,
         )
-
-        job = self.setup_dependencies(job, pipeline=pipeline)
-        # If we do not depend on an unfinished job, enqueue the job.
-        if job.get_status(refresh=False) != JobStatus.DEFERRED:
-            return self.enqueue_job(job, pipeline=pipeline, at_front=at_front)
-        return job
+        return self.enqueue_job(job, pipeline=pipeline, at_front=at_front)
 
     @staticmethod
     def prepare_data(
@@ -736,7 +731,7 @@ class Queue:
         """
         pipe = pipeline if pipeline is not None else self.connection.pipeline()
         jobs = [
-            self.enqueue_job(
+            self._enqueue_job(
                 self.create_job(
                     job_data.func,
                     args=job_data.args,
@@ -977,7 +972,26 @@ class Queue:
         return self.enqueue_at(datetime.now(timezone.utc) + time_delta, func, *args, **kwargs)
 
     def enqueue_job(self, job: 'Job', pipeline: Optional['Pipeline'] = None, at_front: bool = False) -> Job:
-        """Enqueues a job for delayed execution.
+        """Enqueues a job for delayed execution checking dependencies.
+
+        Args:
+            job (Job): The job to enqueue
+            pipeline (Optional[Pipeline], optional): The Redis pipeline to use. Defaults to None.
+            at_front (bool, optional): Whether should enqueue at the front of the queue. Defaults to False.
+
+        Returns:
+            Job: The enqued job
+        """
+        job.origin = self.name
+        job = self.setup_dependencies(job, pipeline=pipeline)
+        # If we do not depend on an unfinished job, enqueue the job.
+        if job.get_status(refresh=False) != JobStatus.DEFERRED:
+            return self._enqueue_job(job, pipeline=pipeline, at_front=at_front)
+        return job
+
+
+    def _enqueue_job(self, job: 'Job', pipeline: Optional['Pipeline'] = None, at_front: bool = False) -> Job:
+        """Enqueues a job for delayed execution without checking dependencies.
 
         If Queue is instantiated with is_async=False, job is executed immediately.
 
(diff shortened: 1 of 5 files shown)


rq/queue.py
```diff
@@ -663,11 +663,23 @@ def enqueue_call(
             on_success=on_success,
             on_failure=on_failure,
         )
+        return self._check_dependencies(job, pipeline=pipeline, at_front=at_front)
 
+    def _check_dependencies(self, job: 'Job', pipeline: Optional['Pipeline'] = None, at_front: bool = False) -> Job:
```
selwin, line 668: `_check_dependencies` shouldn't enqueue job to the queue. Besides, we already have `setup_dependencies`.

rq/queue.py
```diff
@@ -663,11 +663,23 @@ def enqueue_call(
...
+            job (Job): The job to enqueue
+            pipeline (Optional[Pipeline], optional): The Redis pipeline to use. Defaults to None.
+            at_front (bool, optional): Whether should enqueue at the front of the queue. Defaults to False.
+
+        Returns:
+            Job: The enqued job
+        """
         job = self.setup_dependencies(job, pipeline=pipeline)
         # If we do not depend on an unfinished job, enqueue the job.
         if job.get_status(refresh=False) != JobStatus.DEFERRED:
```
selwin, line 681: I think we can simply move these checks to `enqueue_job` and that would solve the issue (`enqueue_job` should also setup dependencies), right?

rq/queue.py
sim6: I changed the name from `_check_dependencies` to `_enqueue_checking_dependencies` to be more clear.

rq/queue.py
sim6: Do you mean the `enqueue_job` added in this pull request or the original `enqueue_job`?

sim6:
I improved new method name and rebased onto `master`.

sim6:
Rebased to `master` to solve conflicts.

rq/queue.py
selwin: I meant we should just add dependency checks to the original [enqueue_job()](https://github.com/rq/rq/blob/e92682c83aefda9475eba3dd9fe0ef42575d92b7/rq/queue.py#L982) method. So it looks something like this:
```python
def enqueue_job(self, job):
     job = self.setup_dependencies(job, pipeline=pipeline)
     # If we do not depend on an unfinished job, enqueue the job.
     if job.get_status(refresh=False) != JobStatus.DEFERRED:
        return self._enqueue_job(job, pipeline=pipeline, at_front=at_front)
    return job
```

sim6: I extracted the code from `enqueue_call()` to `_enqueue_checking_dependencies()` for use in `enqueue_call()` and `enqueue_job()`. Would you rather I duplicate the code?

sim6:
Rebased onto `master`.

rq/queue.py
selwin: Just move the logic in `_enqueue_checking_dependencies()` to `enqueue_job()`. `enqueue_call()` can then call `enqueue_job()`

sim6: I will do it. Thanks.

sim6: Done via https://github.com/rq/rq/compare/3f5e188fb93be91e9c7452d1a5724cdff9d5096a..4cf0db1ea9f2c93bbd2557882438efff3d2b77de

rq/queue.py
sim6: Deleted via https://github.com/rq/rq/compare/3f5e188fb93be91e9c7452d1a5724cdff9d5096a..4cf0db1ea9f2c93bbd2557882438efff3d2b77de

sim6:
Rebased onto `master`.

sim6:
Test docstring fixed.

selwin:
Thanks!

sim6:
:)

PR #1839, opened 2023-02-26
removed import error warning and changed function doc

amks1:
With reference to: https://github.com/rq/rq/discussions/1828

I decided to completely remove the logger line instead of changing it to `debug`, because in all cases during my observation, there was nothing happening that was worthy of a log. 

In case of a class method, the function moves through each component of the dotted path (raising and catching `ImportError`s as needed) till it found a module to import, which is the expected behavior.


selwin:
Thanks!

PR #1840, opened 2023-02-26
improve the docs on in-flight jobs of killed/lost workers

luto:
When a job gets stuck in WIP state, e.g. because the worker was killed or lost its network connection, the job eventually ends up in `FailedJobRegistry` with an `exc_info` value of `Moved to FailedJobRegistry at 2023-02-2(...)`. As far as I can see, this isn't documented anywhere at the moment. This just adds a small paragraph explaining the behavior.

The whole topic of zombie jobs and `StartedJobRegistry.cleanup` probably deserves a dedicated section in the docs. I am just not sure where exactly to put it?

### DIFF
--- a/docs/docs/results.md
+++ b/docs/docs/results.md
@@ -82,6 +82,8 @@ damage. Just sayin'.
 
 If the worker gets killed while a job is running, it will eventually end up in
 `FailedJobRegistry` because a cleanup task will raise an `AbandonedJobError`.
+Before 0.14 the behavor was the same, but the cleanup task raised a
+`Moved to FailedJobRegistry at` error message instead.
 
 ## Dealing with Job Timeouts
 
--- a/docs/docs/results.md
+++ b/docs/docs/results.md
@@ -78,10 +78,12 @@ chance to finish themselves.
 However, workers can be killed forcefully by `kill -9`, which will not give the
 workers a chance to finish the job gracefully or to put the job on the `failed`
 queue.  Therefore, killing a worker forcefully could potentially lead to
-damage.
+damage. Just sayin'.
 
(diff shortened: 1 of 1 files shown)


selwin:
Thanks, this really helps in clarifying this. @ronlut has a PR that changes how abandoned jobs are handled, as well as introducing a new exception specifically for this category of errors. @luto mind taking a look at https://github.com/rq/rq/pull/1824/ and see if this docs is still valid?

luto:
#1824 improves the situation greatly, because the error message becomes searchable. I'd still argue that the "Dealing with Interruptions" section should mention what happens to running jobs when a worker gets killed. Once the other PR is merged, I'll adapt this one to match the wording an de-duplicate the content. Thanks for the hint!

luto:
FYI: Adapted to use the same terminology as #1824. Feel free to squash this into one commit when merging.

selwin:
Mind putting an example of message that users will see in `job.exc_info` when this happens after https://github.com/rq/rq/pull/1824 is merged?

docs/docs/results.md
```diff
@@ -78,10 +78,10 @@ chance to finish themselves.
 However, workers can be killed forcefully by `kill -9`, which will not give the
 workers a chance to finish the job gracefully or to put the job on the `failed`
 queue.  Therefore, killing a worker forcefully could potentially lead to
-damage.
-
-Just sayin'.
+damage. Just sayin'.
 
+If the worker gets killed while a job is running, it will eventually end up in
```
selwin, line 83: We should indicate that this is new in version 0.14

selwin:
Thanks!

PR #1841, opened 2023-02-27
(Docs) Updated patterns - heroku

iggeehu:
Redis To Go no longer exists. The section is replaced with instructions for Heroku Data For Redis


selwin:
Thanks!

PR #1845, opened 2023-02-28
Fix TimerDeathPenalty not properly handling negative/infinite timeout

marqueurs404:
I believe the intended behavior when job timeout is negative is to have infinite timeout. However, `TimerDeathPenalty` immediately times out jobs instead. This PR fixes that behavior and adds a test for that case.


selwin:
Thanks!

PR #1850, opened 2023-03-02
Update arguments passed to the Sentinel Object when created from the settings

### LINKED ISSUE #1849: Add username support for Sentinel
When using Redis Sentinel, the username can't be specified from the configuration.

In the [`get_redis_from_config()`](https://github.com/rq/rq/blob/ed59b9248aeb87c566661d0f5ee2c2e934c41440/rq/cli/helpers.py#L32), the username should be recovered from the settings and sent to the `Sentinel` object:

```python
def get_redis_from_config(settings, connection_class=Redis):
    """Returns a StrictRedis instance from a dictionary of settings.
    To use redis sentinel, you must specify a dictionary in the configuration file.
    Example of a dictionary with keys without values:
    SENTINEL = {'INSTANCES':, 'SOCKET_TIMEOUT':, 'PASSWORD':,'DB':, 'MASTER_NAME':}
    """
    if settings.get('REDIS_URL') is not None:
        return connection_class.from_url(settings['REDIS_URL'])

    elif settings.get('SENTINEL') is not None:
        instances = settings['SENTINEL'].get('INSTANCES', [('localhost', 26379)])
        socket_timeout = settings['SENTINEL'].get('SOCKET_TIMEOUT', None)
        username = settings['SENTINEL'].get('USERNAME', None)
        password = settings['SENTINEL'].get('PASSWORD', None)
        db = settings['SENTINEL'].get('DB', 0)
        master_name = settings['SENTINEL'].get('MASTER_NAME', 'mymaster')
        ssl = settings['SENTINEL'].get('SSL', False)
        arguments = {'password': password, 'username': username, 'ssl': ssl}
        sn = Sentinel(
            instances, socket_timeout=socket_timeout, password=password, username=username, db=db, ssl=ssl, sentinel_kwargs=arguments
        )
        return sn.master_for(master_name)
```

I can make a PR to solve that.

joachimBurket:
resolves #1849 

- added `USERNAME` key
- added `CONNECTION_KWARGS` key to allow passing additionals arguments to the Redis connections
- updated the documentation

### DIFF
--- a/rq/cli/helpers.py
+++ b/rq/cli/helpers.py
@@ -33,21 +33,27 @@ def get_redis_from_config(settings, connection_class=Redis):
     """Returns a StrictRedis instance from a dictionary of settings.
     To use redis sentinel, you must specify a dictionary in the configuration file.
     Example of a dictionary with keys without values:
-    SENTINEL = {'INSTANCES':, 'SOCKET_TIMEOUT':, 'PASSWORD':,'DB':, 'MASTER_NAME':}
+    SENTINEL = {'INSTANCES':, 'SOCKET_TIMEOUT':, 'USERNAME':, 'PASSWORD':, 'DB':, 'MASTER_NAME':, 'SENTINEL_KWARGS':}
     """
     if settings.get('REDIS_URL') is not None:
         return connection_class.from_url(settings['REDIS_URL'])
 
     elif settings.get('SENTINEL') is not None:
         instances = settings['SENTINEL'].get('INSTANCES', [('localhost', 26379)])
-        socket_timeout = settings['SENTINEL'].get('SOCKET_TIMEOUT', None)
-        password = settings['SENTINEL'].get('PASSWORD', None)
-        db = settings['SENTINEL'].get('DB', 0)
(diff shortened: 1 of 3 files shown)


rq/cli/helpers.py
```diff
@@ -33,21 +33,27 @@ def get_redis_from_config(settings, connection_class=Redis):
...
+        
+        connection_kwargs = {
+            'db': settings['SENTINEL'].get('DB', 0),
+            'username': settings['SENTINEL'].get('USERNAME', None),
+            'password': settings['SENTINEL'].get('PASSWORD', None),
+            'socket_timeout': settings['SENTINEL'].get('SOCKET_TIMEOUT', None),
+            'ssl': settings['SENTINEL'].get('SSL', False),
+        }
+        connection_kwargs.update(settings['SENTINEL'].get('CONNECTION_KWARGS', {}))
+        sentinel_kwargs = settings['SENTINEL'].get('SENTINEL_KWARGS', {})
```
selwin, line 53: Thanks for this, can we test this part somehow so that codecov doesn't complain about this?

PR #1851, opened 2023-03-03
Consolidate job callbacks execution

ronlut:


ronlut:
@selwin This PR continues the [PR comment here](https://github.com/rq/rq/pull/1824#discussion_r1123919638)
I know tests are failing, let me know what you think about that direction before I continue and fix them.

rq/job.py
```diff
@@ -147,6 +151,8 @@ def create(
         description: Optional[str] = None,
         depends_on: Optional[JobDependencyType] = None,
         timeout: Optional[int] = None,
+        success_callback_timeout: Optional[int] = None,
```
selwin, line 154: I like the idea of having separate timeouts for success and failure callbacks.

Perhaps we can have a `Callback` class that can be used like this:
```python
queue.enqueue(say_hello, on_success=Callback(do_something, timeout=30))
```

rq/job.py
```diff
@@ -860,6 +872,8 @@ def restore(self, raw_data) -> Any:
             except Exception:
                 self._result = "Unserializable return value"
         self.timeout = parse_timeout(obj.get('timeout')) if obj.get('timeout') else None
+        self.success_callback_timeout = parse_timeout(obj.get('success_callback_timeout')) if obj.get('timeout') else CALLBACK_TIMEOUT
```
selwin, line 875: This is minor, but it would be better if we nest related parameters together so it's easier to reason about. At the moment it looks like this:
```
{
    "success_callback_name": "do_something",
    "success_callback_timeout": 60
}
```
If we nest variables:
```
{
    "success_callback": {
        "name": "do_something",
        "timeout": 60
    }
}
```

ronlut:
@selwin did you requested PR changes, they are both interface breaking so did them backwards-compatible
LMKWYT

rq/job.py
```diff
@@ -1507,3 +1566,16 @@ def __init__(self, max: int, interval: Union[int, List[int]] = 0):
 
         self.max = max
         self.intervals = intervals
+
+
+class Callback:
+    def __init__(self, func: Callable[..., Any], timeout: Optional[Any] = None):
```
selwin, line 1572: `timeout` should be an int or float.

rq/queue.py
```diff
@@ -84,7 +89,8 @@ def all(
 
         def to_queue(queue_key):
             return cls.from_queue_key(
-                as_text(queue_key), connection=connection, job_class=job_class, serializer=serializer
+                as_text(queue_key), connection=connection, job_class=job_class,
+                death_penalty_class=death_penalty_class, serializer=serializer
```
selwin, line 93: `death_penalty_class` should be added after `serializer`, otherwise it would break codes using positional arguments.

rq/job.py
ronlut: ditching support of passing a string? `parse_timeout` at the moment handles int,float and strings like `1m`

selwin: Sorry, totally forgot that we handle strings as timeout too

rq/job.py
```diff
@@ -864,11 +878,19 @@ def restore(self, raw_data) -> Any:
...
+            self._success_callback_name = success_callback['name']
+            self.success_callback_timeout = success_callback.get('timeout', CALLBACK_TIMEOUT)
 
-        if obj.get('failure_callback_name'):
+        if obj.get('failure_callback_name'):  # backward compatibility
             self._failure_callback_name = obj.get('failure_callback_name').decode()
+        elif 'failure_callback' in obj:
+            failure_callback = json.loads(obj.get('failure_callback').decode())
+            self._failure_callback_name = failure_callback['name']
+            self.failure_callback_timeout = failure_callback.get('timeout', CALLBACK_TIMEOUT)
```
selwin, line 893: It didn't occur to me that we'd need to encode this in JSON because Redis hashes doesn't support nesting. Had I known this, I wouldn't have suggested that we store callback data in a nested structure.

Do you want to revert this to how it was previously, or do you think it's worth the trouble to serialize and deserialize to JSON? I think this just adds unneeded complexity to job persistence.

rq/cli/helpers.py
```diff
@@ -325,6 +327,11 @@ def __init__(
         except (ImportError, AttributeError) as exc:
             raise click.BadParameter(str(exc), param_hint='--job-class')
 
+        try:
+            self.death_penalty_class = import_attribute(death_penalty_class)
+        except (ImportError, AttributeError) as exc:
+            raise click.BadParameter(str(exc), param_hint='--death-penalty-class')
```
selwin, line 333: Mind testing this check so codecov doesn't complain?

selwin:
Can we mention something about the new `Callback` class in the docs?

rq/job.py
```diff
@@ -1308,6 +1326,39 @@ def finished_job_registry(self):
...
+
+        Args:
+            death_penalty_class (Type[BaseDeathPenalty]): The penalty class to use for timeout
+            result (Any): The job's result.
+        """
+        if not self.success_callback:
+            return
+
+        logger.debug('Running success callbacks for %s', self.id)
+        self.heartbeat(utcnow(), self.success_callback_timeout)
```
selwin, line 1341: Mind moving `heartbeat` call out of this method? I think it's cleaner to have the worker explicitly call `heartbeat()` in the execution flow.

rq/job.py
```diff
@@ -1308,6 +1326,39 @@ def finished_job_registry(self):
...
+    def execute_failure_callback(self, death_penalty_class: Type[BaseDeathPenalty], *exc_info, heartbeat=False):
+        """Executes failure_callback with possible timeout
+        """
+        if not self.failure_callback:
+            return
+
+        logger.debug('Running failure callbacks for %s', self.id)
+        try:
+            if heartbeat:
+                self.heartbeat(utcnow(), self.failure_callback_timeout)
```
selwin, line 1354: Mind moving `heartbeat` out of this method so it's explicitly called by the worker?

ronlut: This requires the timeout to never be None because the worker now uses `job.X_timeout` when sending heartbeat, even if there is no callback defined.
I can add an `if` but if feels weird the worker should know that.

selwin:
Thanks!

PR #1852, opened 2023-03-05
bugfix: type annotations as str to allow Windows usage

### LINKED ISSUE #1848: #1729 breaks usage on Windows
#1729 introduced the following line:

`import resource`

So on windows a `ModuleNotFoundError` is raised. So even the recommended use of SimpleWorker on windows isn't possible anymore.

lowercase00:
Removes the `resource` import (that breaks usage for Windows), and wraps it to `try/except` block while type checking so that we still have the same effect on POSIX platforms.

Needs testing on Windows.

Fixes #1848

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -4,7 +4,6 @@ import logging
 import math
 import os
 import random
-import resource
 import signal
 import socket
 import sys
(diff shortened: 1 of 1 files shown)


selwin:
@lowercase00 is this PR still a draft or is it ready for review?

lowercase00:
It's still a draft, I don't have Windows easily accessible to test, so I sent this to the relevant issues, but seems no one on Windows tested yet.

selwin:
@viaregio mind testing this and see if it solves the problem for you?

viaregio:
Good morning from my part of the world. I installed with pip using the following command line

```bash
pip install git+https://github.com/lowercase00/rq.git@bugfix/windows
```

With this version I can startup a SimpleWorker again and it is processing jobs like before. So I think this fixes the issue.

PR #1853, opened 2023-03-05
feat: move `reorder_queues` to `dequeue_job`

lowercase00:
As discussed in #1803

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -830,7 +830,6 @@ class Worker:
                         break
 
                     job, queue = result
-                    self.reorder_queues(reference_queue=queue)
                     self.execute_job(job, queue)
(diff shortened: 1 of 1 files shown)


rq/worker.py
```diff
@@ -938,6 +937,9 @@ def dequeue_job_and_maintain_ttl(self, timeout: Optional[int], max_idle_time: Op
                 connection_wait_time = 1.0
 
         self.heartbeat()
+        if result:
+            _, queue = result
+            self.reorder_queues(reference_queue=queue)
```
selwin, line 942: Mind moving this to line 915?

lowercase00: Sure, much cleaner indeed.

PR #1857, opened 2023-03-06
fix: remove unsafe formating and standardize qoutes

kuwv:
Remove unsafe formatting and use builtin logging parameters instead.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -336,7 +336,7 @@ class Queue:
         else:
             end = length
         job_ids = [as_text(job_id) for job_id in self.connection.lrange(self.key, start, end)]
-        self.log.debug(f"Getting jobs for queue {green(self.name)}: {len(job_ids)} found.")
+        self.log.debug('Getting jobs for queue %s: %d found.', green(self.name), len(job_ids))
         return job_ids
 
     def get_jobs(self, offset: int = 0, length: int = -1) -> List['Job']:
@@ -455,7 +455,7 @@ class Queue:
             result = connection.lpush(self.key, job_id)
         else:
             result = connection.rpush(self.key, job_id)
-        self.log.debug(f"Pushed job {blue(job_id)} into {green(self.name)}, {result} job(s) are in queue.")
+        self.log.debug('Pushed job %s into %s, %s job(s) are in queue.', blue(job_id), green(self.name), result)
 
     def create_job(
             self,
--- a/rq/scheduler.py
+++ b/rq/scheduler.py
@@ -109,7 +109,7 @@ class RQScheduler:
         """Returns names of queue it successfully acquires lock on"""
         successful_locks = set()
         pid = os.getpid()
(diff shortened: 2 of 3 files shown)


kuwv:
@selwin Is this fine?

selwin:
Thanks for the PR! Mind checking the tests and make sure they pass?

selwin:
@kuwv there's a merge conflict that needs to be fixed.

kuwv:
@selwin Should be good to go now.

ronlut:
@kuwv can you explain the motivation for the formatting change? Mostly out of curiosity. I find the readability of f-strings much better.
Is the unsafeness you are talking about is what mentioned here https://bugs.python.org/issue46200 ?
I am not sure this should be "a standard", it's mostly relevant when logging user-supplied data, but when it's internal fields, ids and so on (like in `{job.id}`) I don't see the value. I agree a lot [with the comment here](https://bugs.python.org/issue46200#msg409499). 
Would love to hear your thoughts :)

kuwv:
@ronlut That's a fair question.

First, security:

Basically, we assume a `job.id` will be a string but that can just as easily be an object implementing `__str__` that dumps any available attrs within scope. Also, there are other attributes that are also being logged that can be used. Of course, this is only possible if a developer decides to use user supplied inputs to populate jobs. We could also say that there is no way of ensure a developer knows this.

IMO, exploits using log4shell demonstrated why we should be our "brother's keeper" in this regard.

https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/
https://www.geeksforgeeks.org/vulnerability-in-str-format-in-python/

Secondly, it's lazy:
https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html
https://medium.com/swlh/why-it-matters-how-you-log-in-python-1a1085851205

Lastly, this was an easy change because %90 of the codebase already used prepared statements for logging. This merge just does the "...should be one-- and preferably only one --obvious way to do it". Changing all the codebase to use f-strings would have been harder to do actually and potentially more unsafe.

lowercase00:
@kuwv that's interesting, I actually made a few of the changes that are using `fstrings` particularly because just as @ronlut I find readability a whole lot better - in my mind we would go the opposite direction and replace everything to `fstrings`. Security has been less of a concern in my mind, mostly because (1) it's not user supplied data and (2) this is mostly internal stuff. But anyways, better safe then sorry, and standard is relevant indeed.

> Basically, we assume a job.id will be a string but that can just as easily be an object implementing __str__ that dumps any available attrs within scope.

Just as a quick note on the side, going against Python's community long established practices, I particularly think that this shouldn't even be allowed. Actually I'm of the opinion that a lot of times the dynamic aspects of the language carry less value then strongly typed (annotated), predictable codebase (one of the reasons we've made quite a bit of effort to add type annotation to most of the codebase, with docstrings all around).

kuwv:
@lowercase00 yeah, totally use that if it were available. I am pretty happy about this though: https://github.com/RustPython/RustPython

PR #1859, opened 2023-03-07
Remove `use_connection`

lowercase00:
The deprecation warning for `use_connection` first appeared when the docs moved to this repo (~2019 if I'm not mistaken), this effectively removes the `use_connection` from the codebase. This is also the first step to completely eliminate `local` from the codebase: we are still missing deprecating `push/pop` connections, which are only used by the worker on the `perform_job` method, and the `with Connection` context manager, which already has a deprecation warning in place.

### DIFF
--- a/docs/docs/connections.md
+++ b/docs/docs/connections.md
@@ -3,61 +3,20 @@ title: "RQ: Connections"
 layout: docs
 ---
 
-Although RQ features the `use_connection()` command for convenience, it
-is deprecated, since it pollutes the global namespace.  Instead, prefer explicit
-connection management using the `with Connection(...):` context manager, or
-pass in Redis connection references to queues directly.
+### The connection parameter
 
-
-## Single Redis connection (easy)
-
-<div class="warning">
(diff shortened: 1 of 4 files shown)


selwin:
Thanks!

gabriels1234:
@selwin this breaks current django-rq.

selwin:
@gabriels1234 hmm ok, mind taking creating a patch for django-rq? I don't think I'll have the time to look at this in the next few days.

gabriels1234:
@selwin Will try, now focusing on handling #1871

gabriels1234:
@selwin
Does this mean that we should remove also the "get_current_job()"?
this is from django-rq:
image

selwin:
@gabriels1234 yes, we should remove all uses of `use_connection` and use explicit connections instead.

ccrvlh:
@gabriels1234 are you working on the patch? I can help if you need. I think it might be worth to have the version compatibility updated also

PR #1860, opened 2023-03-07
add deprecation warnings to connection functions

lowercase00:
Most of the functions on the `connections` module are used to interact with a Redis connection implicitly through the LocalStack: `push/pop/use/resolve_connection`.

The connection ideally should be handled explicitly and the `with Connection` already has a deprecation warning. This adds deprecation warnings to those functions. After these are deprecated, we would be able to delete the `local.py` entirely, and only handle connections explicitly.

### DIFF
--- a/rq/connections.py
+++ b/rq/connections.py
@@ -1,6 +1,7 @@
+import warnings
 from contextlib import contextmanager
 from typing import Optional
-import warnings
+
 from redis import Redis
 
 from .local import LocalStack, release_local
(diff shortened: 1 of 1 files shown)


lowercase00:
Failing tests are unexpected since only warnings were added. Tests are getting a lot more warnings indeed, since this is mostly used for tests. Let me know if you agree with this path @selwin so I know whether to invest time in this or not.

selwin:
> Tests are getting a lot more warnings indeed, since this is mostly used for tests.

I'm ok with this, just make sure that the tests pass :)

lowercase00:
Sure. I think this is because of the large number of warnings. Will look into it.

selwin:
> EDIT: if that is actually the case, perhaps it would make sense to already work on the tests and have the connection be an attribute of the RQTestCase, that way we would rely on the implicit management of push/pop around the tests.

👍 

I'll merge this PR in first.

PR #1864, opened 2023-03-07
build: implement pep621

kuwv:


pyproject.toml
```diff
@@ -1,4 +1,81 @@
...
+    'Programming Language :: Python :: 3.11',
+    'Topic :: Software Development :: Libraries :: Python Modules',
+    'Topic :: Internet',
+    'Topic :: Scientific/Engineering',
+    'Topic :: System :: Distributed Computing',
+    'Topic :: System :: Systems Administration',
+    'Topic :: System :: Monitoring',
+]
+dependencies = [
+    'redis>=3.0.0,<4',
```
selwin, line 49: We're compatible with `Redis >= 3`.

selwin: Redis `redis>=3.5.0` to be exact.

pyproject.toml
```diff
@@ -1,4 +1,81 @@
...
+    'click>=3.0,<4',
```
selwin, line 50: This also needs to be updated, `requirements.txt` specifies that we require `click>=5.0.0`.

pyproject.toml
```diff
@@ -1,4 +1,81 @@
...
+]
+
+[project.optional-dependencies]
+dev = [
+    'build',
+    'codecov>=2,<3',
+    'psutil>=5,<6',
+    'pytest>=7,<8',
+    'pytest-cov>=4,<5',
+    'sentry-sdk>=1,<2',
```
selwin, line 60: Please don't add unnecessary `<` to dependencies, I'd rather keep everything the same.

pyproject.toml
```diff
@@ -1,4 +1,81 @@
...
+[project.scripts]
+rq = 'rq.cli:main'
+# TODO: rqworker/rqinfo are kept for backward-compatibility remove in v2.0
+rqinfo = 'rq.cli:info'
+rqworker = 'rq.cli:worker'
+
+[project.urls]
+homepage = 'https://python-rq.org/'
+documentation = 'https://python-rq.org/docs/'
+repository = 'https://github.com/nvie/rq/'
```
selwin, line 72: This should be changed to `https://github.com/rq/rq/`

selwin:
Mind making the tests pass?

pyproject.toml
kuwv: done

pyproject.toml
kuwv: done

pyproject.toml
kuwv: fixed

pyproject.toml
kuwv: done

kuwv:
> Mind making the tests pass?

Well, still looking at it. The Ubuntu image is garbage: https://github.com/pypa/setuptools/issues/3269

kuwv:
@selwin setuptools doesn't support the `pyproject.toml` config for anything under v60. Python3.6 support was dropped in v59.

There are several options for declarative builds here:
1. Manage project metadata in `setup.cfg`.
2. Perform builds once and test on each version of Python (idiomatic builds)
3. Drop python3.6

selwin:
Sorry for the delayed response @kuwv . I'd like to keep Python 3.6 compatibility for a bit longer if possible.

What would the implications be if we choose to go with option 2?

kuwv:
No worries @selwin.

I agree about 3.6 support. There are probably a good many of systems that still use it.

As for impact, many projects build once so it won't be much. We'll need to keep setuptools pinned low enough to still support 3.6 while using a Python version high enough so that it can run setuptools properly.

I'd recommend python 3.11 for its speed.

selwin:
> I'd recommend python 3.11 for its speed. I'll double check tonight it those versions will fly.

👍

selwin:
@kuwv just a reminder to merge master into this branch. In the past few days I merged in a few PRs from @robhudson and @tchapi that improved our test and linting infrastructure.

selwin:
@kuwv is 3.6 causing all sort of build problems? If so we can consider dropping it, since it’s already end of life’d.

kuwv:
@selwin what is causing this: https://github.com/rq/rq/issues/1577

selwin:
The idea behind the `Dependencies` workflow is that RQ is continuously built with the latest versions of libraries such that when a dependency is updated and causes RQ to break, we'll find out in advance.

But in practice it created a lot of noise and I just commented it out. I don't think the noise is worth it because:
1. We're usually pretty quick with releasing updates whenever a major version of a dependency is released.
2. All dependencies we use are mature and as far as I know we don't use any undocumented methods.

kuwv:
@selwin reason I ask is that redis 6/7 are throwing those errors within the workflows.yml pipeline.

selwin:
@kuwv can we do this in a more step by step manner? For example:
1. Moving `setup()` function in `setup.py` to `pyproject.toml` first, without touching workflows.yml.
2. Eliminate `Makefile` (I don't think this is used anymore)

robhudson:
Since the project has a `.pre-commit-config.yaml` it might be worth adding [pyproject-fmt](https://github.com/tox-dev/pyproject-fmt).

selwin:
@robhudson PR welcome :)

kuwv:
Closing this since 3.6 has been deprecated.

PR #1874, opened 2023-04-02
Worker pool

selwin:
This is still a work in progress, still missing a CLI command to run a worker pool. I'm making opening this PR early to solicit feedback.

When finished, this is the kind of functionality I'm looking to add to RQ:
`rq pool high default low --num-workers=5`


ankush:
@selwin I've tried this out and looks very promising. Especially from memory efficiency POV because each worker is now forked from the same process. 


Is there any help required to get this to "stable" state?

---

Roughly I've seen around 60% of memory usage drop with 8 workers. (sum of pss measured using `smem`)

image


I also attempted hacky version of _auto-scaler_ of sorts which can be implemented on top of WorkerPool class here: https://github.com/ankush/rq_orchestrator/blob/develop/rq_orchestrator/frappe_worker_pool.py

selwin:
I labeled this beta out of abundance of caution so that a smaller number of users can try it out first.So far I haven’t heard of any issues with this feature. I’ll probably wait for a few more months and if there’s nothing major, I’ll take the beta label off.On Jun 25, 2023, at 7:09 PM, Ankush Menat ***@***.***> wrote:﻿
@selwin I've tried this out and looks very promising. Especially from memory efficiency POV because each worker is now forked from the same process.
Is there any help required to get this to "stable" state?

Roughly I've seen around 60% of memory usage drop with 8 workers. (sum of pss measured using smem)

I also attempted hacky version of auto-scaler of sorts which can be implemented on top of WorkerPool class here: https://github.com/ankush/rq_orchestrator/blob/develop/rq_orchestrator/frappe_worker_pool.py

jackkinsella:
I wanted to bring up an issue for worker pool that blocks its use with num_workers > 1 in Django applications that communicate with postgres via SSL (possibly any ORM that communicates with any DB over SSL may also be affected)

I also have a proposed solution. See this ossue, the linked-to articles (specifically about SSL MAC addresses changing during the fork call triggered by `Process` in `worker_pool.py`)

https://github.com/rq/django-rq/issues/650#issuecomment-1973353042

PR #1882, opened 2023-04-23
Add py.typed for PEP561 compliance

### LINKED ISSUE #1881: Typing information not exposed
The typing information added in #1698 is not available for projects that import the `rq` package. This is because RQ does not comply with [PEP 561](https://peps.python.org/pep-0561/) which requires a `py.typed` file in the root of the package.

For projects utilising MyPy this will result in the following error:
```
src/example.py:9: error: Skipping analyzing "rq": module is installed, but missing library stubs or py.typed marker  [import]
```

Further reading:
- https://peps.python.org/pep-0561/#packaging-type-information
- https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-library-stubs-or-py-typed-marker

Alt titles for searchability:
- `py.typed` missing
- RQ not PEP 561 compliant

nhairs:
This allows external projects that use type checkers themselves to read the typing information that RQ contains.

Without this file, compliant type-checkers will skip type checks for project code that calls RQ.

Fixes #1881

Note: this PR should be backwards compatible to pretty much any branch, so if there are more versions maintained than just master branch, applying the same patch to them would allow older versions to type check without upgrading to the latest version of RQ.

## Test Plan
1. Create test file `example.py`:
```lang=py
from redis import Redis
from rq import Queue

q = Queue(connection=Redis())
result = q.enqueue(sum, [1,2,3,4,5])
```
2. (optional: create virtual environment first) Install dependencies
    `pip install mypy redis types-redis rq`
3. Run MyPy `mypy example.py`
    Should observe MyPy `import` error.
4. Build new version of RQ `python3 setup.py bdist_wheel`
5. Install new version of RQ `pip install ./path/to/rq/dist/rq-1.13.0-py2.py3-none-any.whl`
6. Re-run MyPy (step 3)
    Observe no errors

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ setup(
                 'jobs, and processing them.',
     long_description=__doc__,
     packages=find_packages(exclude=['tests', 'tests.*']),
+    package_data = {"rq": ["py.typed"]},
     include_package_data=True,
     zip_safe=False,
     platforms='any',


selwin:
Thanks!

Issue #1879, opened 2023-04-17
[Scheduler] Connection pool class not used by scheduler to instantiate new connections


tchapi:
Hi

First of all thanks for the great work on this project 👍🏼 

While working with RQ in our code, we noticed that we could not make the RQ scheduler (`rq worker --with-scheduler`)  work correctly, as it would not use our connection the way we intended it to be used (_with some custom behavior like automatically reconnecting after secret rotation for instance_).

In the [scheduler code](https://github.com/rq/rq/blob/master/rq/scheduler.py#L57), you're using `self._connection_class = connection.__class__` which is bypassing any custom connection class we might want to use by explicitly using the _connection instance_ class name to recreate the connection (ie, `Redis`).

In the next line, the code is retrieving the `connection_class = connection.connection_pool.connection_class` (which is indeed our custom class), but does not use it to instantiate the connection. Is this intended?

aka. Is there any reason not to do this?

```python
self._connection_class = connection.connection_pool.connection_class
```

Happy to provide more details if needed
Thanks a lot for your answer!

selwin:
Please make a PR for this :) 

`self._connection_class = connection.connection_pool.connection_class`

tchapi:
Done in https://github.com/rq/rq/pull/1885 (with a slightly different approach)

tchapi:
Fixed in latest release

PR #1885, opened 2023-04-25
[Fix] Allow custom connection pool class for the scheduler

tchapi:
Following https://github.com/rq/rq/issues/1879, I'm proposing a change to allow passing a **custom connection class to the scheduler**;

To date, we can pass a custom connection to a worker with:

```python
my_custom_conn = redis.Redis(
    connection_pool=redis.ConnectionPool(
        connection_class=CustomRedisConnection,
        custom_arg="foo",
    )
)
worker = Worker(connection=my_custom_conn)
worker.work()
```

but doing so while enabling the scheduler will raise:
```python
worker.work(with_scheduler=True)
# raises:
# TypeError: Redis.__init__() got an unexpected 
# keyword argument 'custom_arg'
```

Passing extra arguments to the connection pool class is a standard feature in redis-py, as per the [documentation](https://redis.readthedocs.io/en/stable/connections.html#redis.connection.ConnectionPool).

I added 2 tests to verify that the custom connection class is indeed used, and that it does not interfere with the actual behavior (ie, using a standard `Redis` client).

_meta: I haven't found in the repo anything about the code standards / formatting / etc — feel free to point me to your contributing guidelines if needed_

Best regards,

### DIFF
--- a/rq/connections.py
+++ b/rq/connections.py
@@ -2,7 +2,7 @@ import warnings
 from contextlib import contextmanager
 from typing import Any, Optional, Tuple, Type
 
-from redis import Redis, SSLConnection, UnixDomainSocketConnection
+from redis import Connection as RedisConnection, Redis, SSLConnection, UnixDomainSocketConnection
 
 from .local import LocalStack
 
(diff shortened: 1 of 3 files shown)


tchapi:
> Do you mind moving this out to a helper function like what I did here? https://github.com/rq/rq/pull/1884

Not at all! Just for clarity, do you want me to use your branch as my base (and add my code in your "parse" function)?

selwin:
> Not at all! Just for clarity, do you want me to use your branch as my base (and add my code in your "parse" function)?

Whatever's easier for you :)

tchapi:
Done and based on `parse-redis-connection` - tell me if this is better for you

rq/connections.py
```diff
@@ -117,15 +117,15 @@ def resolve_connection(connection: Optional['Redis'] = None) -> 'Redis':
...
+def parse_connection(connection: Redis) -> Tuple[Type[Redis], Type[RedisConnection], dict]:
     connection_kwargs = connection.connection_pool.connection_kwargs.copy()
     # Redis does not accept parser_class argument which is sometimes present
     # on connection_pool kwargs, for example when hiredis is used
     connection_kwargs.pop('parser_class', None)
-    connection_class = connection.connection_pool.connection_class
-    if issubclass(connection_class, SSLConnection):
+    connection_pool_class = connection.connection_pool.connection_class
+    if issubclass(connection_pool_class, SSLConnection):
         connection_kwargs['ssl'] = True
```
tchapi, line 127: ℹ️ I just renamed the var for clarity

dams:
@selwin slightly off-topic question :) What's your release plan/Strategy? In other words, when can we expect the merged code to be available in a public release?

selwin:
@dams I usually try to make a release once every 3 months (roughly).

PR #1886, opened 2023-04-25
Update `push_job_id` debug string when using pipelines

robhudson:
Prior to this, the debug line would dump the pipeline `__repr__` into the string instead of the number of jobs on the queue.

Example:
```
DEBUG 2023-04-25 18:17:40,497 queue Pushed job c6027356-9ac4-4228-bae9-3e057543ca18 into default, Pipeline<ConnectionPool<Connection<host=redis,port=6379,db=2>>> job(s) are in queue.
```


selwin:
Thanks!

PR #1888, opened 2023-04-26
[Results] Allow unserializable return values

tchapi:
Hi 👋🏼 

Before RQ 1.12, RQ was actually allowing unserializable results in `job.save()` ([here](https://github.com/rq/rq/blob/master/rq/job.py#L977)):

```python
# note the try/except
try:
    obj['result'] = self.serializer.dumps(self._result)
except:  # noqa
    obj['result'] = "Unserializable return value"
```

RQ 1.12 introduced a breaking change in this regard, as we're now storing the result also in a proper `Result` object to keep multiple results ([here](https://github.com/rq/rq/blob/master/rq/results.py#L184)):

```python
# no try/except block here
serialized = self.serializer.dumps(self.return_value)
```

This PR proposes to be less strict and allow unserializable results (as before).
I'm also adding a test that tries to save a known non-picklable value (`tempfile.NamedTemporaryFile()`) to assess that it doesn't break

Happy to have feedback and adapt the code if needed
Thanks a lot!

### DIFF
--- a/rq/results.py
+++ b/rq/results.py
@@ -181,7 +181,11 @@ class Result:
         if self.exc_string is not None:
             data['exc_string'] = b64encode(zlib.compress(self.exc_string.encode())).decode()
 
-        serialized = self.serializer.dumps(self.return_value)
+        try:
+            serialized = self.serializer.dumps(self.return_value)
+        except:  # noqa
+            return data
+
         if self.return_value is not None:
             data['return_value'] = b64encode(serialized).decode()
 
--- a/rq/results.py
+++ b/rq/results.py
@@ -6,6 +6,7 @@ from datetime import datetime, timezone
 from enum import Enum
 from redis import Redis
 
+from .defaults import UNSERIALIZABLE_RETURN_VALUE_PAYLOAD
 from .utils import decode_redis_hash
 from .job import Job
 from .serializers import resolve_serializer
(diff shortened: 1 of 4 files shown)


rq/results.py
```diff
@@ -181,7 +181,11 @@ def serialize(self):
         if self.exc_string is not None:
             data['exc_string'] = b64encode(zlib.compress(self.exc_string.encode())).decode()
 
-        serialized = self.serializer.dumps(self.return_value)
+        try:
+            serialized = self.serializer.dumps(self.return_value)
+        except:  # noqa
+            return data
```
tchapi, line 187: ℹ️ I'm returning early so we don't b64encode a null value, which has no real sense here

selwin, line 187: Mind putting a simple `<Unserializable return value>` string as the `return_value`? This tells the user that the job actually returns something that we can't keep.

tchapi: Oh yes sure! I'll put the same string as in the `job._result` code

tchapi: Done in https://github.com/rq/rq/pull/1888/commits/5bf8239ea2a8378a6d0e344fcf80aaa07bcb37f3
I took the liberty of creating a default value so we don't repeat the same string twice in the code - I think it's cleaner this way

selwin:
Thanks!

PR #1894, opened 2023-05-03
[Hotfix] Fix SSL connection for scheduler

tchapi:
It seems that we were unnecessarily [changing the connection pool kwargs](https://github.com/tchapi/rq/blob/5bf8239ea2a8378a6d0e344fcf80aaa07bcb37f3/rq/connections.py#L126-L135) because we were trying to pass them to a Client instead of a ConnectionPool:

```python
if issubclass(connection_pool_class, SSLConnection):
    connection_kwargs['ssl'] = True
if issubclass(connection_pool_class, UnixDomainSocketConnection):
    # The connection keyword arguments are obtained from
    # `UnixDomainSocketConnection`, which expects `path`, but passed to
    # `redis.client.Redis`, which expects `unix_socket_path`, renaming
    # the key is necessary.
    # `path` is not left in the dictionary as that keyword argument is
    # not expected by `redis.client.Redis` and would raise an exception.
    connection_kwargs['unix_socket_path'] = connection_kwargs.pop('path')
```

Now that we're correctly using a ConnectionPool (since #1885), we don't need those changes anymore

### DIFF
--- a/rq/scheduler.py
+++ b/rq/scheduler.py
@@ -50,7 +50,7 @@ class RQScheduler:
         self._acquired_locks: Set[str] = set()
         self._scheduled_job_registries: List[ScheduledJobRegistry] = []
         self.lock_acquisition_time = None
-        self._connection_class, self._pool_class, self._connection_kwargs = parse_connection(connection)
+        self._connection_class, self._pool_class, self._pool_kwargs = parse_connection(connection)
         self.serializer = resolve_serializer(serializer)
 
         self._connection = None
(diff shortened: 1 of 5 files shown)


rq/scheduler.py
```diff
@@ -50,7 +50,7 @@ def __init__(
         self._acquired_locks: Set[str] = set()
         self._scheduled_job_registries: List[ScheduledJobRegistry] = []
         self.lock_acquisition_time = None
-        self._connection_class, self._pool_class, self._connection_kwargs = parse_connection(connection)
+        self._connection_class, self._pool_class, self._pool_kwargs = parse_connection(connection)
         self.serializer = resolve_serializer(serializer)
```
tchapi, line 54: Just renaming things for consistency

tchapi:
Tests pass BUT we're actually skipping the SSL tests in every run ([example](https://github.com/rq/rq/actions/runs/4870018245/jobs/8685247263?pr=1894)):

```
tests/test_scheduler.py ................s......                          [ 71%]
```

PR #1897, opened 2023-05-03
Add support for dependent jobs in enqueue_many

eswolinsky3241:


selwin:
This is a good feature to have. Let me know when this PR is ready :). It needs some tests at least.

selwin:
Honestly, this dependency checking part is one of the most complex part of RQ, especially because it needs to handle race conditions.

As much as possible I'd like to leave the current logic alone and come up with a new code path for dependency checking in `enqueue_many()`. We can even put this new bits of code in a new `dependency.py` module. I'm also ok with doing this in two parts to keep the code simpler:
1. Create all the jobs in deferred state
2. Check which jobs already have their dependency satisfied and enqueue them in a bulk fashion.

This way, both steps 1 and 2 can be unit tested.

eswolinsky3241:
@selwin Mind reviewing my last commit when you get a chance? Let me know if I'm on the right track and I'll work on the tests.

selwin:
Hey, this is a good attempt, I like this approach, it looks quite straight forward. Will review soon.

rq/job.py
```diff
@@ -716,7 +716,7 @@ def dependents_key(self):
 
     @property
     def dependencies_key(self):
-        return '{0}:{1}:dependencies'.format(self.redis_job_namespace_prefix, self.id)
+        return '{0}{1}:dependencies'.format(self.redis_job_namespace_prefix, self.id)
```
selwin, line 719: I'd rather not change this because this would break existing keys.

rq/queue.py
```diff
@@ -754,8 +758,10 @@ def enqueue_many(self, job_datas: List['EnqueueData'], pipeline: Optional['Pipel
             List[Job]: A list of enqueued jobs
         """
         pipe = pipeline if pipeline is not None else self.connection.pipeline()
-        jobs = [
-            self._enqueue_job(
+        if any([job_data.depends_on for job_data in job_datas]):
```
selwin, line 761: I think we should split the jobs into two different lists, with and without dependencies.

rq/job.py
eswolinsky3241: Reverted this change

rq/queue.py
eswolinsky3241: I tried this but it got pretty messy. It seems cleaner to just pass in all jobs to `Dependency.get_ready_jobs` and let that function return the jobs without dependencies, along with those whose dependencies have been met.

selwin: But this would mean if I enqueue 10 jobs and only one of them had dependencies, the 9 other jobs would be created in deferred state to be requeued later, right? I'd rather we do something like this:
```python
jobs_with_dependencies = [job_data for job_data in job_datas if job_data.depends_on]
if jobs_with_dependencies:
    ...
jobs_without_dependencies = [job_data for job_data in job_datas if not job_data.depends_on]
if jobs_without_dependencies:
    ....
```

eswolinsky3241: Makes sense, just pushed this change

rq/dependency.py
```diff
@@ -0,0 +1,25 @@
+from .job import Job, JobStatus
+from redis.client import Pipeline
+from typing import List
+from redis.exceptions import WatchError
+
+
+class Dependency:
+    @classmethod
+    def get_ready_jobs(cls, jobs: List['Job'], pipeline: Pipeline):
```
selwin, line 9: Can we rename `get_ready_jobs` to something more descriptive? I don't have better ideas at the moment, but `ready` is not a term that's already used in this library.

selwin:
In general the approach is sound, some tests are needed though :)

rq/dependency.py
eswolinsky3241: Gave it a shot but it's very wordy, let me know if that's alright: https://github.com/rq/rq/pull/1897/commits/fd211b9ae4b06e3cb3f0b56e09f078c263a2ba3c

eswolinsky3241:
@selwin Just committed some tests, let me know if these are sufficient

rq/dependency.py
```diff
@@ -0,0 +1,26 @@
+from .job import Job
+from redis.client import Pipeline
+from typing import List
+from redis.exceptions import WatchError
+
+
+class Dependency:
+    @classmethod
+    def get_jobs_with_dependencies_met(cls, jobs: List['Job'], pipeline: Pipeline):
```
selwin, line 9: I think this should be `get_jobs_with_met_dependencies`?

rq/dependency.py
```diff
@@ -0,0 +1,26 @@
...
+        pipe = pipeline if pipeline else cls.connection
```
selwin, line 10: `pipeline` is required so no need to fallback to `cls.connection`

rq/queue.py
```diff
@@ -754,8 +758,44 @@ def enqueue_many(self, job_datas: List['EnqueueData'], pipeline: Optional['Pipel
...
+                        failure_ttl=job_data.failure_ttl,
+                        description=job_data.description,
+                        depends_on=job_data.depends_on,
+                        job_id=job_data.job_id,
+                        meta=job_data.meta,
+                        status=JobStatus.QUEUED,
+                        timeout=job_data.timeout,
+                        retry=job_data.retry,
+                        on_success=job_data.on_success,
+                        on_failure=job_data.on_failure,
```
selwin, line 784: Since kwargs are mostly the same, mind putting these kwargs into a `kwargs` dict that can be passed on to both `create_job` calls in this method? All the kwargs should be the same except for `status`, right?

tests/test_dependencies.py
```diff
@@ -101,21 +98,150 @@ def test_job_dependency(self):
...
+        self.assertEqual(job2.get_status(), JobStatus.FINISHED)
+        q.empty()
+
+        # don't enqueue dependent job when parent fails
+        parent_job = q.enqueue(div_by_zero)
+        job1 = Queue.prepare_data(say_hello, depends_on=parent_job)
+        job2 = Queue.prepare_data(say_hello, depends_on=parent_job)
+
+        jobs = q.enqueue_many([job1, job2])
+        w.work(burst=True)
```
selwin, line 132: I think we don't need to retest whether jobs get enqueued if parent job succeeds or fails, this is already tested else where. For this PR, we simply need to test that jobs (with and without dependencies) are properly created when created with `enqueue_many`.

rq/dependency.py
eswolinsky3241: I renamed the function and relevant variables

rq/dependency.py
eswolinsky3241: Removed

rq/queue.py
eswolinsky3241: Let me know if my solution looks correct

tests/test_dependencies.py
eswolinsky3241: I removed the redundant tests, just left the one that checks the job statuses

selwin:
Thanks, mind fixing the conflict so I can merge this in?

eswolinsky3241:
Resolved, thanks!

rq/dependency.py
```diff
@@ -0,0 +1,25 @@
...
+    def get_jobs_with_met_dependencies(cls, jobs: List['Job'], pipeline: Pipeline):
+        jobs_with_met_dependencies = []
+        jobs_with_unmet_dependencies = []
+        for job in jobs:
+            while True:
+                try:
+                    pipeline.watch(*[Job.key_for(dependency_id) for dependency_id in job._dependency_ids])
+                    job.register_dependency(pipeline=pipeline)
+                    if job.dependencies_are_met(pipeline=pipeline):
+                        jobs_with_met_dependencies.append(job)
```
selwin, line 18: Hey sorry but it seems like this logic branch is not tested. To hit this branch, we need to call `enqueue_many()` with jobs whose dependencies are met.

Sorry for being fussy, hopefully this will be the last review.

rq/queue.py
```diff
@@ -26,6 +26,7 @@
 from .serializers import resolve_serializer
 from .types import FunctionReferenceType, JobDependencyType
 from .utils import as_text, backend_class, compact, get_version, import_attribute, parse_timeout, utcnow
+from .dependency import Dependency
```
selwin, line 29: Imports need to be sorted alphabetically.

eswolinsky3241: Resolved

rq/dependency.py
eswolinsky3241: Just pushed a change to the test. Really appreciate your help with this!

selwin:
Thanks!

PR #1905, opened 2023-05-08
callback func as string

### LINKED ISSUE #1891: Allow on_success and on_failure be referenced using string reference
RQ allows to refer to the job function using string reference, for example:

```python
q.enqueue(
    f="my_package.my_module.my_func"
)
```

I expected that `on_success` and `on_failure` can be used in the same way:

```python
q.enqueue(
    f="packageA.module.func",
    on_success="packageB.module.funcA",
    on_failure="packageB.module.funcB"
)
```

But I got this error: `on_success callback must be a function`. 

And then I ended up with placing these callback functions in the same module where `enqueue` code resides:

```python
q.enqueue(
    f="packageA.module.func",
    on_success=funcA,
    on_failure=funcB
)
```

But it feels that referring callbacks using string reference is a more right way because I want these functions to be placed inside worker module, and I do not have access to the worker module during enqueue call because it happens from separate project. That's the whole purpose why `f` can be string reference.

rishabh-ranjan:
fixes #1891


rishabh-ranjan:
Hi, sorry I'm new to this. What are the checks that are failing? And how to fix them?

selwin:
Thanks for the PR, can you add some test for this?

rq/job.py
```diff
@@ -1580,13 +1580,15 @@ def __init__(self, max: int, interval: Union[int, List[int]] = 0):
...
+        if not isinstance(func, str) and not inspect.isfunction(func) and not inspect.isbuiltin(func):
+            raise ValueError('Callback func must be a string or function')
 
         self.func = func
         self.timeout = parse_timeout(timeout) if timeout else CALLBACK_TIMEOUT
 
     @property
     def name(self) -> str:
+        if isinstance(self.func, str):
+            return self.func
```
selwin, line 1593: @rishabh-ranjan you need to add a test where `Callback` is passed in with a string.

rishabh-ranjan: please take a look at the new commit. may need some polish, but is hopefully most of the way there.

selwin:
@rishabh-ranjan this PR has merge conflicts. Can you also fix linting errors detailed here? https://github.com/rq/rq/actions/runs/5062692637/jobs/9088482071?pr=1905

tests/test_callbacks.py
```diff
@@ -53,6 +81,20 @@ def test_enqueue_with_failure_callback(self):
         job = Job.fetch(id=job.id, connection=self.testconn)
         self.assertEqual(job.failure_callback, print)
 
+    def test_enqueue_with_stopped_callback_as_string(self):
```
selwin, line 84: Hi, can you consolidate these three tests into one single test case? Using the regular `queue.enqueue()` is also sufficient.

selwin:
Additionally, we also need to make sure that the callbacks get executed properly. Actually I think it would be easier to amend existing tests for these callbacks and add a version with string as inputs.

rishabh-ranjan:
added a string version to all existing tests. maybe this is an overkill, but couldn't think of a simpler concise alternative.

selwin:
Adding string versions to existing tests is good. Do you mind deleting the new testcases added prior to this? (Sorry)

rishabh-ranjan:
done

selwin:
@rishabh-ranjan this looks good, if you can fix the merge conflict, I'll merge this in and release it as part of 1.15.

rishabh-ranjan:
@selwin can you take a look at the failed check? not sure if its due to my changes

selwin:
Github actions is flaky sometimes. All I had to do was rerun the failed jobs @rishabh-ranjan

PR #1906, opened 2023-05-08
Update cli.py to support custom loggers

juur:
Allows a config.py file (via rq worker --config) to support things like this for logfmt logging:

```
DICT_CONFIG = {
        "version": 1,
        "formatters": {"logfmt": {
            "()": "logfmter.Logfmter",
            "keys": ["level","when","pid"],
            "mapping": {"level":"levelname","when":"asctime","pid":"process"},
            "datefmt": "%Y-%m-%dT%H:%M:%S%z"
            }},
        "handlers": {"console": {"class": "logging.StreamHandler","formatter": "logfmt" }},
        "loggers": {
            "root": {"handlers":["console"], "level": "INFO"},
            "rq":   {"handlers":["console"], "level": "INFO", "propagate": False},
            }
        }
```


selwin:
Thanks for the PR. Mind updating the test a bit to add logging config? Ideally, this should also be documented.

juur:
I would love to, however I don't have the skill needed to update your complicated test suite to properly test the config file. I could just add it to tests/config_files/dummy.py but that doesn't really test it.

selwin:
> I would love to, however I don't have the skill needed to update your complicated test suite to properly test the config file. I could just add it to tests/config_files/dummy.py but that doesn't really test it.

That's actually all I need, the logic of parsing logging config itself is handled by `logging.config.dictConfig`. I just wanted to make sure that the config variable is properly used and doesn't cause a crash.

juur:
> > I would love to, however I don't have the skill needed to update your complicated test suite to properly test the config file. I could just add it to tests/config_files/dummy.py but that doesn't really test it.
> 
> That's actually all I need, the logic of parsing logging config itself is handled by `logging.config.dictConfig`. I just wanted to make sure that the config variable is properly used and doesn't cause a crash.

Hopefully I've managed to do that (and the documentation).

rq/cli/cli.py
```diff
@@ -227,6 +227,11 @@ def worker(
     sentry_debug = sentry_debug or settings.get('SENTRY_DEBUG')
     sentry_dsn = sentry_dsn or settings.get('SENTRY_DSN')
     name = name or settings.get('NAME')
+    dict_config = options.get("dict_config", settings.get('DICT_CONFIG'))
+    
+    if dict_config:
+        import logging.config
+        logging.config.dictConfig(dict_config)
```
selwin, line 234: Hmm this line is still marked as untested, which probably means this line is never triggered. Have you verified that this works properly? @juur

rq/cli/cli.py
```diff
@@ -227,6 +227,11 @@ def worker(
     sentry_debug = sentry_debug or settings.get('SENTRY_DEBUG')
     sentry_dsn = sentry_dsn or settings.get('SENTRY_DSN')
     name = name or settings.get('NAME')
+    dict_config = options.get("dict_config", settings.get('DICT_CONFIG'))
```
selwin, line 230: `dict_config` is never passed through `options`, right?

rq/cli/cli.py
```diff
@@ -227,6 +227,11 @@ def worker(
...
+    
+    if dict_config:
+        import logging.config
```
selwin, line 233: This import can be moved to module level.

rq/cli/cli.py
juur: I was trying to be consistent with the lines above.

rq/cli/cli.py
juur: Done

rq/cli/cli.py
selwin: But if `dict_config` is never passed through `options`, I don't think we need that. It's more important to be correct than to be consistent :)

juur:
I think this is OK now. I had forgotten to check my alternative test config into git ... Coverage report shows the dictConfig being processed and tests passing.

PR #1908, opened 2023-05-09
Stop calling deprecated function `resolve_connection`

### LINKED ISSUE #1907: Initialising a `Queue` object calls a deprecated function
It looks like the `__init__` function of `Queue` calls the function `resolve_connection` which raises a `DeprecationWarning`, is that right?

I'm assuming not.

tomhamiltonstubber:
Fix #1907

The function `resolve_connection` is now deprecated and doesn't need to be called from lots of places by default.

Now `resolve_connection` will only be called if `connection` has not been passed through already.

This should probably be done for `push_connection` also, but not in this PR.

### DIFF
--- a/rq/connections.py
+++ b/rq/connections.py
@@ -1,8 +1,8 @@
 import warnings
 from contextlib import contextmanager
-from typing import Any, Optional, Tuple, Type
+from typing import Optional, Tuple, Type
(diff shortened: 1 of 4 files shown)


selwin:
Thanks!

PR #1911, opened 2023-05-14
Reliable queue

selwin:


selwin:
@tchapi the SSL tests take about 20 minutes to finish. Is it possible to change this workflow so that it doesn't test across all supported Python and Redis server versions, but rather just across supported Redis versions?

tchapi:
> @tchapi the SSL tests take about 20 minutes to finish. Is it possible to change this workflow so that it doesn't test across all supported Python and Redis server versions, but rather just across supported Redis versions?

Of course! I'm AFK for a few days but I can submit a PR next week if that's ok for you

selwin:
> Of course! I'm AFK for a few days but I can submit a PR next week if that's ok for you

@tchapi no worries :), this is not a big deal, but it's sometimes annoying that I need to wait longer waiting for CI to pass before I can merge in a PR. Thanks!

PR #1914, opened 2023-05-16
Scheduler should release and heartbeat only acquired locks

xzander:
In current implementation there is possible such case:
1.  run 1st scheduler for 1 queue (for example queue-1), this scheduler acquires lock for that queue
2. run 2nd scheduler for 2 queues (for example queue-1 and queue-2), this scheduler acquires lock for queue-2 only, as queue-1 already acquired
3. stop 2nd scheduler and it will release both locks – for queue-1 and queue-2
4. start 3rd scheduler for queue-1, there is no lock, so it acquires the lock for queue-1 
5. now there are 2 schedulers (1 and 3) work for the same queue
6. it's possible that they call zrangebyscore simultaneously (`job_ids = registry.get_jobs_to_schedule(timestamp)`) and then put the same job in a queue twice via `queue._enqueue_job(job, pipeline=pipeline, at_front=bool(job.enqueue_at_front))`


selwin:
Thanks, this is a hard to track bug. Rerunning the SSL tests, will merge after it goes green.

tests/test_scheduler.py
```diff
@@ -203,6 +203,33 @@ def test_lock_acquisition_with_auto_start(self):
             self.assertEqual(mocked.call_count, 1)
             self.assertEqual(stopped_process.is_alive.call_count, 1)
 
+    def test_lock_release(self):
+        """Test lock acquisition"""
```
selwin, line 207: Mind updating this to `Test that scheduler.release_locks() only releases acquired locks`?

selwin:
Thanks!

PR #1915, opened 2023-05-16
Update linting configuration

robhudson:
This removes flake8 in favor of ruff, which also provides isort support, and updates all files to be black, isort, and ruff compliant. This also adds black and ruff checks to the tox and Github linting workflow.


.github/workflows/lint.yml
```diff
@@ -25,15 +25,13 @@ jobs:
...
-        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-        flake8 . --exit-zero --max-complexity=5
+        black --check --skip-string-normalization --line-length 120 rq tests
 
-    - name: Lint with black
+    - name: Lint with ruff
       run: |
-        black -S -l 120 rq/
+        # stop the build if there are Python syntax errors or undefined names
+        ruff check rq tests --show-source --exit-zero
```
robhudson, line 37: I should update the comment above this line, but I also think we want don't want `--exit-zero`. We want to see the failure if there are syntax or other errors. I'll update this after other feedback.

setup.py
```diff
@@ -3,7 +3,8 @@
 processing them.
 """
 import os
-from setuptools import setup, find_packages
+
+from setuptools import find_packages, setup
```
selwin, line 7: Can we exclude `setup.py` from `coverage.py`?

selwin:
Thanks for working on this, looks good 👍 . I can merge this in after your [next update](https://github.com/rq/rq/pull/1915#discussion_r1195619418).

setup.py
robhudson: Sure thing! I tweaked the `.coveragerc` slightly (removed some outdated paths and told it our source lives under `rq/`), then passed that wherever we are calling `pytest` with `--cov`.

selwin:
Thanks!

PR #1918, opened 2023-05-21
[Tests] Only run SSL tests in Docker

tchapi:
### Rationale

Following [this comment](https://github.com/rq/rq/pull/1911#issuecomment-1552758192):
- actually we only test Ubuntu's packaged Redis version (5.0.7 if I'm [not mistaken](https://packages.ubuntu.com/focal/database/redis))
- it takes a while because all tests are run for every tox env
- I'm proposing a modification so that only specific SSL tests are run on the `ssl-test` job, and only for the Python 3.10 env (_see tox config file_)
- We still only test against 5.0.7 but I think it's fine for now

### Result:

It's quicker now.

Screenshot 2023-05-21 at 22 06 15

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -21,11 +21,11 @@ jobs:
         file: tests/Dockerfile
         tags: rqtest-image:latest
         push: false
-    - name: Launch tox for all test scenarii
+    - name: Launch tox SSL env only (will only run SSL specific tests)
       uses: addnab/docker-run-action@v3
       with:
         image: rqtest-image:latest
-        run: stunnel & redis-server & RUN_SSL_TESTS=1 tox
+        run: stunnel & redis-server & RUN_SSL_TESTS=1 tox run -e ssl
 
   test:
     name: Python${{ matrix.python-version }}/Redis${{ matrix.redis-version }}/redis-py${{ matrix.redis-py-version }}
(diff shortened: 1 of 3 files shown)


selwin:
So just to clarify, this PR makes it such that it only runs SSL tests against Python 3.10 and the default Redis version installed on Ubuntu 20.04, right?

If we want to tests against a newer Python version, only [this line](https://github.com/rq/rq/pull/1918/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R51) needs to be changed, right?

tchapi:
> this PR makes it such that it only runs SSL tests against Python 3.10 and the default Redis version installed on Ubuntu 20.04, right?

Yes. Previous behaviour was "all tests for all python versions with default Ubuntu redis version"

> If we want to tests against a newer Python version, only [this line](https://github.com/rq/rq/pull/1918/files#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R51) needs to be changed, right?

Yes exactly. I don't know tox very much but I guess we could pass a list here somehow?

selwin:
> Yes exactly. I don't know tox very much but I guess we could pass a list here somehow?

No worries, I think testing against a single Python version is enough, just asking just in case we need to bump this in the future.

selwin:
Thanks!

tchapi:
> Thanks

Well, thank **you** for maintaining the library, and being reactive and helpful, it's a great pleasure

PR #1924, opened 2023-05-24
dont append exc_string to log message

bnomis:
https://docs.python.org/3/library/logging.html#logging.Logger.debug

self.log.error is called with arguments so % formatting operations are performed on the msg string. 

Before this PR exc_string was appended to the msg string and the new msg string if passed to the logging module. If exc_string contains, for example, %s it will be treated as a variable to be filled and the logging module will look for args to fill it. This will raise an exception if there are not enough args or the wrong args will get used.

This can be re-produced by raising an exception like this in a worker:

```python
raise Exception('desc %s' % value)
```

"raise Exception('desc %s' % value)" will be in exc_string and therefore part of the msg before this PR

The fix is not to append exc_string to msg but to include is in the msg string via %s processing and pass it as an arg

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -21,7 +21,7 @@ jobs:
     - uses: actions/checkout@v3
 
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v4.6.0
+      uses: actions/setup-python@v4.6.1
       with:
         python-version: ${{ matrix.python-version }}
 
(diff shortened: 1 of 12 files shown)


selwin:
@bnomis can you make sure that the tests pass?

bnomis:
> @bnomis can you make sure that the tests pass?

@selwin  Which tests? The GHA checks are failing. I can't re-run them and they don't provide any error message except "Error" I think there is an issue with the GHA scripts.

PR #1925, opened 2023-05-24
Update log string for better output when multiple queues

robhudson:
I added a 2nd queue to my worker today and saw the debug string as:

* `Starting BLPOP operation for queues rq:queue:defaultrq:queue:snitch with timeout of 405`
* `BLPOP Timeout, no jobs found on queues rq:queue:defaultrq:queue:snitch`

and wanted to fix this to be more readable.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -1274,7 +1274,7 @@ class Queue:
         if timeout is not None:  # blocking variant
             if timeout == 0:
                 raise ValueError('RQ does not support indefinite timeouts. Please pick a timeout value > 0')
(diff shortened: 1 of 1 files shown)


selwin:
Should we change `rq:queue:defaultrq:queue:snitch` to only the queue names? `BLPOP Timeout, no jobs found on queues default, snitch` would be even more readable.

robhudson:
Personally I like the full redis key names.

selwin:
Ok, I have no strong opinion on this. Merged :)

PR #1928, opened 2023-05-29
Add __all__ to please mypy

aumayr:
`mypy` complains with "implicit reexport disabled" when importing directly from the main `rq` package, eg.

```python
from rq import Worker
```

Without adding an `__all__` to `__init__.py`, mypy would expect it this way:

```python
from rq.worker import Worker
```

This PR does add the `__all__` to `__init__.py` to please mypy.


selwin:
Can you fix the lint error?

PR #1940, opened 2023-06-08
add lint target in Makefile

bnomis:
This PR add a "lint" target to the Makefile 
~~It also adds black and ruff to the dev-requirements~~
It does not add black and ruff to dev-requirements because ruff does not work with python 3.6 and adding those dependencies seems to cause the tests to fail for other python versions. I guess because they cause some version change of related dependencies which are specified without locked versions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,3 +19,7 @@ force_release: clean
 	git push --tags
 	python setup.py sdist bdist_wheel
 	twine upload dist/*
+
+lint:
+	@ black --check --skip-string-normalization --line-length 120 rq tests
+	@ ruff check --show-source rq tests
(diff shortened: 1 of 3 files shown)


selwin:
Actually I wonder if we can get rid of the `Makefile` altogether. I'm not even sure what it's used for.

bnomis:
For running commands locally rather than waiting for the GHA to run - giving a faster code/test loop. For example, "make test" is quite useful and I added "make lint" to run black/ruff.

selwin:
I was thinking of getting rid of this for a Python native version so we don't end up maintaining different ways to do the same things. For example, we can simply use `pytest` to run the tests. Is there a more pythonic way to run linters?

bnomis:
I suggest to keep the Makefile - it does its job well and it documents the commands to run to do a thing. For example, it makes it clear how to run all the tests in Docker and it does that with a simple "make test". I suggest not to spend time duplicating working functionality in a slightly different flavour.

selwin:
Thanks @bnomis !

PR #1945, opened 2023-06-12
Group jobs into batches and retrieve by batch name

eswolinsky3241:
This PR adds an option for the `enqueue_many` function return a `Batch` object, which holds a collection of jobs. The Batch acts as a vehicle to retrieve and manage a collection of jobs using a single Batch ID.


rq/batch.py
```diff
@@ -0,0 +1,85 @@
...
+            job.set_batch_id(self.id)
+            job.save(pipeline=self.connection)
+        self.renew_ttl()
+        self.save()
+
+    def fetch_jobs(self) -> list:
+        self.job_ids = [as_text(job) for job in self.connection.smembers(self.key + ":jobs")]
+        self.jobs = Job.fetch_many(self.job_ids, self.connection)
+
+    def renew_ttl(self, pipeline=None):
```
selwin, line 45: I don't think we should worry about Batch TTLs, they should always be persisted. We can simply delete the batch key once the last job in batch is deleted. What do you think?

rq/batch.py
```diff
@@ -0,0 +1,85 @@
...
+        self.ttl = ttl
+
+        self.jobs = []
+        if jobs:
+            self.add_jobs(jobs)
+
+    def add_jobs(self, jobs: List[Union['Job', str]], pipeline=None):
+        pipe = pipeline if pipeline else self.connection.pipeline
+        self.job_ids = [job.id for job in jobs]
+        self.connection.sadd(self.jobs_key, *self.job_ids)
```
selwin, line 33: Another consideration is what kind of data structure do we use to represent a batch in Redis? Options include:
* Stream - used to store job results
* SortedSet - used by job registries
* Set (current)

selwin:
Once we finalize what approach we want to take, we can also start merging this into master slowly, starting from the `Batch` class.

rq/batch.py
eswolinsky3241: Reason I went with the Batch TTL is I thought it might be expensive to determine whether a job is the "last" job in a large batch. If the job is expired in Redis, I assume we would need workers to run a maintenance task to check if any of the batch's jobs still exist. Lmk if I'm off here.

Also, for our use case, we'd ideally be able to see the status of all jobs in a batch as an atomic unit. If one user enqueues 100 jobs in a batch, and some expire before the whole batch has finished, it could mislead another user checking the status of the jobs. That's why the batch TTL that is inherited by the member jobs was appealing,

selwin: RQ runs maintenance tasks every 10 minutes or so. https://github.com/rq/rq/blob/10230ff0404073d892418c04879d8f04ad57ca9e/rq/worker.py#L309 . This is where we remove expired/deleted jobs from registries. So the mechanism for this already exists.

> it might be expensive to determine whether a job is the "last" job in a large batch.
I think we can around this by batching the deletions. For example if we have 1000 expired jobs in a batch of a million jobs, we can remove 1,000 jobs in a single call and simply check how many jobs are left in the queue before deleting it. 

I just checked Redis' set documentation and it just so happens that the way sets are implemented in Redis, when the last element is deleted from the set, the key will also be deleted. https://redis.io/commands/srem/

eswolinsky3241: What are your thoughts on having the batch TTL be optional and determine the job cleanup method? If batch_ttl is None, the jobs behave as normal with their own TTLs, and the maintenance task eventually deletes the jobs from the batch:jobs set when the job is deleted from the registry. If the user specifies a TTL, the `persist_jobs` and `renew_ttl` methods are used so all of the batches jobs expire at about the same time after the last job completes, which is useful if the user doesn't want to deal with partial batches.

selwin: > What are your thoughts on having the batch TTL be optional and determine the job cleanup method?

Actually, I think `Batch` objects themselves should not have TTLs. `Batch` should be just a light way of grouping jobs together. Each job within a batch can have their own TTLs. If needed, we can provide a shortcut to change TTLs of all jobs within a batch.

> If batch_ttl is None, the jobs behave as normal with their own TTLs, and the maintenance task eventually deletes the jobs from the batch:jobs set when the job is deleted from the registry.

Yes, I think this is the way to go.

Additionally, we'll need a key to store all available batches, similar to how all workers are stored in a single key https://github.com/rq/rq/blob/master/rq/worker.py#L245

selwin: > This is definitely something we'd like, is the current method of updating all job TTLs in a batch when a job completes acceptable?

Each job's TTL should be managed by the user, no implicit TTL updates should be done. What we need to manage are just the batch object themselves.

eswolinsky3241:
@selwin I removed all of the batch TTL logic. The worker maintenance task will now loop through all batches (stored in `rq:batches` and remove expired jobs from the individual batch keys. If the batch is emptied, it is removed from the batch list.

eswolinsky3241:
I'll fix the failing tests and add documentation, then mark this PR as ready for review

rq/batch.py
eswolinsky3241: I'm happy with the set. As you said, batch should just be a container for job IDs that gets cleaned out as those jobs expire.

rq/batch.py
```diff
@@ -0,0 +1,80 @@
...
+from redis.client import Pipeline
+
+from .exceptions import NoSuchBatchError
+from .job import Job
+from .utils import as_text
+
+logger = logging.getLogger("rq.job")
+
+
+class Batch:
```
selwin, line 15: We'll need a `Batch.all()` to return all batches in this Redis DB. The implementation should be similar to [Worker.all()](https://github.com/rq/rq/blob/master/rq/worker.py#L227).

docs/docs/index.md
```diff
@@ -134,6 +134,35 @@ with q.connection.pipeline() as pipe:
...
+
+```python
+batch = q.enqueue_many(
+  [
+    Queue.prepare_data(count_words_at_url, ('http://nvie.com',), job_id='my_job_id'),
+    Queue.prepare_data(count_words_at_url, ('http://nvie.com',), job_id='my_other_job_id'),
+  ],
+  batch=True,
+  batch_id='my_batch'
+)
```
selwin, line 149: I'm not quite sure about this API. It doesn't feel intuitive to me.

This would also make `enqueue_many()` return a list of jobs or a batch depending on its kwarg, which may be confusing.

docs/docs/index.md
```diff
@@ -134,6 +134,35 @@ with q.connection.pipeline() as pipe:
...
+```
+
```
selwin, line 151: Maybe this is a better API?
```python
batch = Batch.create(connection=redis)
batch.enqueue_many()
```

rq/job.py
```diff
@@ -1248,6 +1253,12 @@ def delete(
         if delete_dependents:
             self.delete_dependents(pipeline=pipeline)
 
+        if self.batch_id:
+            from .batch import Batch
+
+            batch = Batch.fetch(id=self.batch_id, connection=self.connection)
```
selwin, line 1259: This incurs an extra Redis call, let's make a `Batch.get_key(batch_id)` method so we can get the batch's Redis key without using `Batch.fetch()`.

docs/docs/index.md
selwin: This API raises a few questions, do we allow jobs from different queues to be put in the same batch?

rq/job.py
eswolinsky3241: Added get_key method.

docs/docs/index.md
eswolinsky3241: I think so. For example if a user wants to batch jobs enqueued to high, medium, and low queues, that seems ok to me.

eswolinsky3241: What if I adjust the APIs to have batching jobs be totally separate from enqueueing? Users would still enqueue jobs with Queue.enqueue_many, and then they would have the option to add the returned array of jobs to a Batch using a new `create` method, or add the jobs to an existing batch with the `add_jobs` method.

rq/batch.py
eswolinsky3241: Committed the all method

docs/docs/index.md
eswolinsky3241: I reverted the changes to `enqueue_many` and added a new `create` class method to Batch. Users can add jobs to a batch after they've been created or enqueued.

docs/docs/index.md
selwin: > What if I adjust the APIs to have batching jobs be totally separate from enqueueing?

I don't think this is a good idea because this means the enqueueing and batching of jobs wouldn't be atomic. This may lead to hard to debug race conditions in the future if we were to put more functionalities into `Batch`.

> Users would still enqueue jobs with Queue.enqueue_many.

This is something we can still do, both of these should still be possible:
```python
queue.enqueue(foo, batch=batch_object)
queue.enqueue(foo, batch='batch_name')

queue.enqueue_many(job_1, job_2, batch=batch)
```

rq/batch.py
```diff
@@ -0,0 +1,96 @@
...
+        pipe.sadd(self.key, *[job.id for job in jobs])
+        pipe.sadd(self.REDIS_BATCH_KEY, self.id)
+        self.jobs += jobs
+        for job in jobs:
+            job.set_batch_id(self.id, pipeline=pipe)
+            job.save(pipeline=pipe)
+        if pipeline is None:
+            pipe.execute()
+
+    def delete_expired_jobs(self, pipeline: Optional['Pipeline'] = None):
```
selwin, line 44: Can we rename this to `cleanup()` so it's consistent with other registry's cleanup method?

rq/batch.py
```diff
@@ -0,0 +1,96 @@
...
+        """Delete jobs from the batch's job registry that have been deleted or expired from Redis.
+        We assume while running this that alive jobs have all been fetched from Redis in fetch_jobs method"""
+
+        pipe = pipeline if pipeline else self.connection.pipeline()
+        job_ids = {as_text(job) for job in self.connection.smembers(self.key)}
+        expired_jobs = job_ids - set([job.id for job in self.jobs if job])  # Return jobs that can't be fetched
```
selwin, line 50: Using [exists](https://redis.io/commands/exists/) here will be much cheaper operation.

rq/batch.py
```diff
@@ -0,0 +1,96 @@
...
+            pipe.srem(self.key, job)
+        if pipeline is None:
+            pipe.execute()
+
+    def fetch_jobs(self) -> list:
+        """Retrieve list of job IDs from the batch key in Redis, then fetch and add those jobs to attribute"""
+        job_ids = [as_text(job) for job in self.connection.smembers(self.key)]
+        self.jobs = [job for job in Job.fetch_many(job_ids, self.connection) if job is not None]
+
+    def refresh(self, pipeline: Optional['Pipeline'] = None):
```
selwin, line 61: We shouldn't need to have `batch.refresh()` as we should never implicitly load jobs from Redis, this could be a very expensive operation if a batch contains a large number of jobs.

rq/batch.py
eswolinsky3241: Done

rq/batch.py
eswolinsky3241: Done

eswolinsky3241:
@selwin Could I please get a new review on these changes whenever you have time?

selwin:
Will try to review this week

selwin:
I’ll try to take a look soon. Sorry, my schedule recently has been packed.On Aug 4, 2023, at 7:46 PM, Ethan Wolinsky ***@***.***> wrote:﻿
@selwin Any chance I could get another review on this please?

rq/job.py
```diff
@@ -1248,6 +1254,11 @@ def delete(
         if delete_dependents:
             self.delete_dependents(pipeline=pipeline)
 
+        if self.batch_id:
+            from .batch import Batch
+
+            connection.delete(Batch.get_key(self.batch_id), self.id)  # Delete job from batch
```
selwin, line 1260: `connection.delete()` will delete the batch object from Redis. We should use `srem` instead. This command should also be pipelined.

selwin: Also, mind adding a test for this? Deleting a job should not automatically delete batch if there are other jobs in the batch.

selwin: To be safe, I think we should do `Batch().delete(job, pipeline=pipeline)`

rq/queue.py
```diff
@@ -856,6 +869,12 @@ def get_job_kwargs(job_data, initial_status):
             ]
             if pipeline is None:
                 pipe.execute()
+
+        all_jobs = jobs_without_dependencies + jobs_with_unmet_dependencies + jobs_with_met_dependencies
+
+        if batch is not None:
+            batch._add_jobs(all_jobs)
```
selwin, line 876: This call should be pipelined.

rq/batch.py
```diff
@@ -0,0 +1,89 @@
...
+                p.exists(Job.key_for(job))
+            results = p.execute()
+
+        for i, key_exists in enumerate(results):
+            if not key_exists:
+                expired_jobs.append(job_ids[i])
+
+        for job in expired_jobs:
+            job_ids.remove(job)
+            pipe.srem(self.key, job)
```
selwin, line 52: We can pass multiple keys into a single `srem` call, we should do that instead of repeating `srem` multiple times.

rq/batch.py
```diff
@@ -0,0 +1,89 @@
...
+    def cleanup(self, pipeline: Optional['Pipeline'] = None):
+        """Delete jobs from the batch's job registry that have been deleted or expired from Redis.
+        We assume while running this that alive jobs have all been fetched from Redis in fetch_jobs method"""
+
+        pipe = pipeline if pipeline else self.connection.pipeline()
+        job_ids = [as_text(job) for job in list(self.connection.smembers(self.key))]
+        expired_jobs = []
+        with self.connection.pipeline() as p:
+            for job in job_ids:
+                p.exists(Job.key_for(job))
```
selwin, line 43: `exists()` supports multiple keys so we should pass in multiple keys directly. https://redis.io/commands/exists/

rq/worker.py
```diff
@@ -711,6 +717,16 @@ def heartbeat(self, timeout: Optional[int] = None, pipeline: Optional['Pipeline'
         connection.hset(self.key, 'last_heartbeat', utcformat(utcnow()))
         self.log.debug('Sent heartbeat to prevent worker timeout. Next one should arrive in %s seconds.', timeout)
 
+    def clean_batch_registries(self):
```
selwin, line 720: `worker.py` is getting big, let's move this functionality into a `cleanup_batch_registry()` function in `batch.py`.

selwin:
Please resolve the conflicts also :)

rq/job.py
eswolinsky3241: Done. I believe I do have a test for this in [`test_batch.py`](https://github.com/eswolinsky3241/rq/blob/1b40dd302fa0b39bd286f521d0e00b18dfb7bd78/tests/test_batch.py#L66). Let me know if it needs to be improved



























@selwin, I may be confused, connection is a pipeline if one is passed in correct? I think I'm testing this in [`test_batch.py`](https://github.com/eswolinsky3241/rq/blob/1b40dd302fa0b39bd286f521d0e00b18dfb7bd78/tests/test_batch.py#L66)

rq/batch.py
eswolinsky3241: Done

rq/batch.py
eswolinsky3241: Done

rq/queue.py
eswolinsky3241: Done

rq/worker.py
eswolinsky3241: Done

eswolinsky3241:
@selwin resolved conflicts, thanks!

eswolinsky3241:
@selwin Would you mind reviewing my most recent changes please?

docs/docs/index.md
```diff
@@ -134,6 +134,38 @@ with q.connection.pipeline() as pipe:
 
 `Queue.prepare_data` accepts all arguments that `Queue.parse_args` does.
 
+### Batching jobs
```
selwin, line 137: I think the phrase `Batching jobs` is slightly misleading. We don't allow jobs to be "batched" [execution](https://en.wikipedia.org/wiki/Batch_processing) wise but just grouped, right? Should we change the implementation to use `Group` instead of `Batch`? I'm also open to other naming ideas that more accurately conveys the use case.

docs/docs/index.md
```diff
@@ -134,6 +134,38 @@ with q.connection.pipeline() as pipe:
...
+```python
+from rq.batch import Batch
+
+batch = Batch.create(id="my_batch", connection=redis_conn)
+jobs = q.enqueue_many(
+  [
+    Queue.prepare_data(count_words_at_url, ('http://nvie.com',), job_id='my_job_id'),
+    Queue.prepare_data(count_words_at_url, ('http://nvie.com',), job_id='my_other_job_id'),
+  ],
+  batch=batch  # You can also pass the batch ID as a string, i.e., batch="my_batch"
```
selwin, line 150: I think the interface we expose should be `batch.enqueue_many()`

rq/batch.py
```diff
@@ -0,0 +1,105 @@
...
+        for i, key_exists in enumerate(results):
+            if not key_exists:
+                expired_jobs.append(job_ids[i])
+
+        if expired_jobs:
+            job_ids = [job for job in job_ids if job not in expired_jobs]
+            pipe.srem(self.key, *expired_jobs)
+
+        if not job_ids:
+            self.delete()
```
selwin, line 54: I think batch should not implicitly delete itself. User must explicitly call `batch.delete()` to delete this object.

viperfx:
Would love to see this merged and supported officially. Is there support for cancelling all the jobs via the Batch object?

eswolinsky3241:
@viperfx I haven't had time to look at this in a while, will try to address @selwin's comments this week.

docs/docs/index.md
eswolinsky3241: @selwin I've moved the enqueueing logic to the batch class to match that interface.

docs/docs/index.md
eswolinsky3241: @selwin I've moved the enqueueing logic to the batch class to match that interface.

rq/batch.py
eswolinsky3241: @selwin I've removed that delete

docs/docs/index.md
eswolinsky3241: I like group. I've changed the name

eswolinsky3241:
@selwin Sorry that it's been a while, could I please get a new review on this?

selwin:
@eswolinsky3241 will try to review sometime in the next few days. Feel free to ping me again if I haven't done so in a week.

eswolinsky3241:
@selwin Checking in on this, do you think you'll have time to review this week?

rq/group.py
```diff
@@ -1,38 +1,40 @@
...
 
 
-class Batch:
-    """A Batch is a container for tracking multiple jobs with a single identifier."""
+class Group:
+    """A Group is a container for tracking multiple jobs with a single identifier."""
 
-    REDIS_BATCH_NAME_PREFIX = 'rq:batch:'
-    REDIS_BATCH_KEY = 'rq:batches'
+    REDIS_BATCH_NAME_PREFIX = 'rq:group:'
```
selwin, line 17: This should be `REDIS_GROUP_NAME_PREFIX` and `REDIS_GROUP_KEY`

rq/group.py
```diff
@@ -50,13 +52,27 @@ def cleanup(self, pipeline: Optional['Pipeline'] = None):
             job_ids = [job for job in job_ids if job not in expired_jobs]
             pipe.srem(self.key, *expired_jobs)
 
-        if not job_ids:
-            self.delete()
         if pipeline is None:
             pipe.execute()
 
+    def enqueue_many(
+        self, queue: Union[str, Queue], job_datas: List['EnqueueData'], pipeline: Optional['Pipeline'] = None
```
selwin, line 59: We can simplify this a bit by only accepting `queue: Queue`, what do you think?

docs/docs/index.md
```diff
@@ -134,6 +134,39 @@ with q.connection.pipeline() as pipe:
 
 `Queue.prepare_data` accepts all arguments that `Queue.parse_args` does.
 
+### Grouping jobs
+_New in version 1.16.0._
```
selwin, line 138: `_New in version 2.0_ `

docs/docs/index.md
```diff
@@ -134,6 +134,39 @@ with q.connection.pipeline() as pipe:
...
 
+### Grouping jobs
+_New in version 1.16.0._  
+Multiple jobs can be added to a Group to allow them to be tracked by a single ID:
+
+```python
+from rq import Queue
+from rq.group import Group
+
+group = Group.create(connection=redis_conn)
```
selwin, line 145: I wonder if we should allow (or force) the user to pass in a `name` argument.

rq/worker.py
```diff
@@ -677,6 +678,10 @@ def handle_job_failure(self, job: 'Job', queue: 'Queue', started_job_registry=No
             else:
                 enqueue_dependents = True
 
+            if job.group_id:
+                group = Group.fetch(job.group_id, self.connection)
+                group.cleanup(pipeline=pipeline)
```
selwin, line 683: Having `group.cleanup()` convenience method is nice, but we should also have a class method that accepts Redis connection and `group_id`. In this context, `Group.fetch()` incurs an extra Redis call which can be avoided.

rq/group.py
```diff
@@ -0,0 +1,121 @@
...
+        group_keys = [as_text(key) for key in connection.smembers(cls.REDIS_BATCH_KEY)]
+        return [Group.fetch(key, connection=connection) for key in group_keys]
+
+    @classmethod
+    def get_key(cls, id: str) -> str:
+        """Return the Redis key of the set containing a group's jobs"""
+        return cls.REDIS_BATCH_NAME_PREFIX + id
+
+    @classmethod
+    def clean_group_registries(cls, connection: 'Redis'):
```
selwin, line 113: I think we can rename this to `Group.clean_registries()`

eswolinsky3241: Done

docs/docs/index.md
eswolinsky3241: Done

rq/group.py
eswolinsky3241: Done, thank you

docs/docs/index.md
eswolinsky3241: create (and group constructor) does accept an optional ID, I did not make it required to stay consistent with the Job API.

rq/group.py
eswolinsky3241: Works for me, done

rq/worker.py
eswolinsky3241: Moved the logic to classmethod, let me know if that looks ok

eswolinsky3241:
@selwin Would you mind looking at my most recent changes when you have time?

rq/group.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def __init__(self, connection: Redis, id: str = None):
+        self.id = id if id else str(uuid4())
+        self.connection = connection
+        self.key = '{0}{1}'.format(self.REDIS_BATCH_NAME_PREFIX, self.id)
+
+    def __repr__(self):
+        return "Group(id={})".format(self.id)
+
+    def _add_jobs(self, jobs: List[Job], pipeline: Optional['Pipeline'] = None):
```
selwin, line 28: Since this is an internal method, let's make the `pipeline` argument required.

rq/group.py
```diff
@@ -0,0 +1,121 @@
...
+        if pipeline is None:
+            pipe.execute()
+
+    def cleanup(self, pipeline: Optional['Pipeline'] = None):
+        """Delete jobs from the group's job registry that have been deleted or expired from Redis.
+        We assume while running this that alive jobs have all been fetched from Redis in fetch_jobs method"""
+
+        pipe = pipeline if pipeline else self.connection.pipeline()
+        job_ids = [as_text(job) for job in list(self.connection.smembers(self.key))]
+        expired_jobs = []
```
selwin: Can we rename this to `expired_job_ids`?

rq/group.py
```diff
@@ -0,0 +1,121 @@
...
+            p.exists(*[Job.key_for(job) for job in job_ids])
+            results = p.execute()
+
+        for i, key_exists in enumerate(results):
+            if not key_exists:
+                expired_jobs.append(job_ids[i])
+
+        if expired_jobs:
+            job_ids = [job for job in job_ids if job not in expired_jobs]
+            pipe.srem(self.key, *expired_jobs)
```
selwin: `expired_jobs` should contain all job IDs whose keys are no longer present in Redis, correct? In this case, shouldn't we just directly delete all IDs in `expired_jobs`?

selwin: If this logic is faulty, would need a test to catch this.

rq/group.py
```diff
@@ -0,0 +1,121 @@
...
+    @classmethod
+    def get_key(cls, id: str) -> str:
+        """Return the Redis key of the set containing a group's jobs"""
+        return cls.REDIS_BATCH_NAME_PREFIX + id
+
+    @classmethod
+    def clean_group_registries(cls, connection: 'Redis'):
+        """Loop through groups and delete those that have been deleted.
+        If group still has jobs in its registry, delete those that have expired"""
+        groups = connection.smembers(Group.REDIS_BATCH_KEY)
```
selwin: `groups` -> `group_ids`

rq/group.py
```diff
@@ -0,0 +1,121 @@
...
+from .utils import as_text
+
+
+class Group:
+    """A Group is a container for tracking multiple jobs with a single identifier."""
+
+    REDIS_BATCH_NAME_PREFIX = 'rq:group:'
+    REDIS_BATCH_KEY = 'rq:groups'
+
+    def __init__(self, connection: Redis, id: str = None):
```
selwin, line 20: Can we change `id` to `name`? `id` is a reserved word in Python, plus using `name` would give it more symmetry with `Queue(name='foo')`

tests/test_group.py
```diff
@@ -0,0 +1,139 @@
...
+        enqueued_group.enqueue_many(q, [self.job_1_data, self.job_2_data])
+        fetched_group = Group.fetch(enqueued_group.id, self.testconn)
+        self.assertCountEqual(enqueued_group.get_jobs(), fetched_group.get_jobs())
+        assert len(fetched_group.get_jobs()) == 2
+        q.empty()
+
+    def test_add_jobs(self):
+        q = Queue(connection=self.testconn)
+        group = Group.create(connection=self.testconn)
+        group.enqueue_many(q, [self.job_1_data, self.job_2_data])
```
selwin, line 43: Just want to point out that this design means a single `Group` may contains jobs from multiple different queues. What are your thoughts?

tests/test_group.py
```diff
@@ -0,0 +1,139 @@
...
+        q.empty()
+
+    def test_expired_jobs_removed_from_group(self):
+        q = Queue(connection=self.testconn)
+        w = SimpleWorker([q], connection=q.connection)
+        short_lived_job = Queue.prepare_data(say_hello, result_ttl=1)
+        group = Group.create(connection=self.testconn)
+        group.enqueue_many(q, [short_lived_job, self.job_1_data])
+        w.work(burst=True, max_jobs=1)
+        sleep(3)
```
selwin, line 92: `sleep(2)` so we don't make the test run longer than it needs to be :)

tests/test_group.py
```diff
@@ -0,0 +1,139 @@
...
+        group.cleanup()
+        assert len(group.get_jobs()) == 1
```
selwin, line 94: Can we assert that it's `job_1_data` that remains in the group?

tests/test_group.py
```diff
@@ -0,0 +1,139 @@
...
+        q.empty()
+
+    def test_empty_group_removed_from_group_list(self):
+        q = Queue(connection=self.testconn)
+        w = SimpleWorker([q], connection=q.connection)
+        short_lived_job = Queue.prepare_data(say_hello, result_ttl=1)
+        group = Group.create(connection=self.testconn)
+        group.enqueue_many(q, [short_lived_job])
+        w.work(burst=True, max_jobs=1)
+        sleep(3)
```
selwin, line 104: `sleep(2)` :)

tests/test_group.py
```diff
@@ -0,0 +1,139 @@
...
+        w.run_maintenance_tasks()
+        redis_groups = {as_text(group) for group in self.testconn.smembers("rq:groups")}
+        assert group.id not in redis_groups
+
+    def test_fetch_expired_group_raises_error(self):
```
selwin, line 109: Mind marking all tests with `sleep()` with `@slow` decorator?

rq/group.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @classmethod
+    def create(cls, connection: Redis, id: Optional[str] = None):
+        return cls(id=id, connection=connection)
+
+    @classmethod
+    def fetch(cls, id: str, connection: Redis):
+        """Fetch an existing group from Redis"""
+        group = cls(id=id, connection=connection)
+        group.cleanup()
```
selwin, line 78: I think we shouldn't call `cleanup()` everytime a group is fetched but rather only when job IDs are read from it.

rq/group.py
eswolinsky3241: Done

rq/group.py
eswolinsky3241: Done

tests/test_group.py
eswolinsky3241: Done

tests/test_group.py
eswolinsky3241: Done

tests/test_group.py
eswolinsky3241: Done

tests/test_group.py
eswolinsky3241: Done

tests/test_group.py
eswolinsky3241: I think this is an option we'd like to have. We enqueue large groups of jobs to decrease load on an internal web API. Some of these jobs have to be on different queues because they need to be processed by different types of workers.

rq/group.py
eswolinsky3241: Done

rq/group.py
eswolinsky3241: Done

rq/group.py
eswolinsky3241: Not sure why I was constructing the job_ids variable, I think that was leftover from when I stored the jobs as an attribute on the group instance.

eswolinsky3241:
@selwin Could I please get a new review on this?

rq/group.py
```diff
@@ -17,34 +17,34 @@ class Group:
     REDIS_GROUP_NAME_PREFIX = 'rq:group:'
     REDIS_GROUP_KEY = 'rq:groups'
 
-    def __init__(self, connection: Redis, id: str = None):
-        self.id = id if id else str(uuid4())
+    def __init__(self, connection: Redis, name: str = None):
+        self.name = name if name else str(uuid4())
```
selwin, line 21: Nitpick, let's make this `uuid4().hex` so it's a bit shorter.

rq/group.py
```diff
@@ -68,35 +69,33 @@ def delete_job(self, job_id: str, pipeline: Optional['Pipeline'] = None):
...
-        group.cleanup()
-        if not connection.exists(Group.get_key(group.id)):
+        group = cls(name=name, connection=connection)
+        if not connection.exists(Group.get_key(group.name)):
             raise NoSuchGroupError
         return group
 
     @classmethod
-    def cleanup_group(cls, id: str, connection: Redis, pipeline: Optional['Pipeline'] = None):
+    def cleanup_group(cls, name: str, connection: Redis, pipeline: Optional['Pipeline'] = None):
```
selwin, line 84: Should we make this into an instance method? `group.cleanup()`

rq/group.py
```diff
@@ -119,6 +118,7 @@ def clean_registries(cls, connection: 'Redis'):
         groups = connection.smembers(Group.REDIS_GROUP_KEY)
```
selwin, line 118: Let's expose a method `Group.all(connection)`, to make it symmetric with `Worker.all()`

selwin:
Sorry for the late review, hopefully this will be the last revision. Thanks for the effort!

rq/group.py
eswolinsky3241: Done

rq/group.py
eswolinsky3241: Done

rq/group.py
eswolinsky3241: Done

eswolinsky3241:
@selwin Addressed all comments, thanks!

eswolinsky3241:
@selwin Could you please review again when you have time?

selwin:
Will do in a few days. Sorry for the delay 🙏🏻

eswolinsky3241:
Thanks!

rq/group.py
```diff
@@ -37,7 +37,17 @@ def cleanup(self, pipeline: Optional['Pipeline'] = None):
         """Delete jobs from the group's job registry that have been deleted or expired from Redis.
         We assume while running this that alive jobs have all been fetched from Redis in fetch_jobs method"""
         pipe = pipeline if pipeline else self.connection.pipeline()
-        Group.cleanup_group(self.name, self.connection, pipeline=pipe)
+        job_ids = [as_text(job) for job in list(self.connection.smembers(self.key))]
+        expired_job_ids = []
+        with self.connection.pipeline() as p:
```
selwin, line 42: This should use `pipe` from line 39.

rq/group.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def __repr__(self):
+        return "Group(id={})".format(self.name)
+
+    def _add_jobs(self, jobs: List[Job], pipeline: Pipeline):
+        """Add jobs to the group"""
+        pipe = pipeline if pipeline else self.connection.pipeline()
+        pipe.sadd(self.key, *[job.id for job in jobs])
+        pipe.sadd(self.REDIS_GROUP_KEY, self.name)
+        if pipeline is None:
```
selwin, line 33: `pipeline` is a required argument so this line is not needed.

rq/group.py
```diff
@@ -0,0 +1,126 @@
...
+    REDIS_GROUP_NAME_PREFIX = 'rq:group:'
+    REDIS_GROUP_KEY = 'rq:groups'
+
+    def __init__(self, connection: Redis, name: str = None):
+        self.name = name if name else str(uuid4().hex)
+        self.connection = connection
+        self.key = '{0}{1}'.format(self.REDIS_GROUP_NAME_PREFIX, self.name)
+
+    def __repr__(self):
+        return "Group(id={})".format(self.name)
```
selwin, line 26: Can we somehow test this to increase coverage?

rq/group.py
```diff
@@ -0,0 +1,126 @@
...
+        groups = Group.all(connection=connection)
+        with connection.pipeline() as p:
+            if groups:
+                # Remove expired jobs from groups
+                for group in groups:
+                    group.cleanup(pipeline=p)
+                p.execute()
+                # Remove empty groups from group registry
+                for group in groups:
+                    p.exists(group.key)
```
selwin, line 118: Should we move this logic into `group.cleanup()`?

tests/test_group.py
```diff
@@ -0,0 +1,140 @@
...
+        redis_groups = {as_text(group) for group in self.testconn.smembers("rq:groups")}
+        assert group.name not in redis_groups
+
+    @pytest.mark.slow
+    def test_fetch_expired_group_raises_error(self):
+        q = Queue(connection=self.testconn)
+        w = SimpleWorker([q], connection=q.connection)
+        short_lived_job = Queue.prepare_data(say_hello, result_ttl=1)
+        group = Group.create(connection=self.testconn)
+        group.enqueue_many(q, [short_lived_job])
```
selwin, line 121: You have put a lot of time and effort into this PR which I greatly apppreciate. This is just something extra (not required) but I wonder if we should also have an API from the queue side to make it a bit more consistent with the rest of RQ's APIs:
```python
queue.enqueue(func, group=group)
```

Honestly I have also forgotten whether the earlier iterations had this option.

selwin:
Friendly reminder to look at the last batch of comments @eswolinsky3241 :). Ideally we can get this into the next 2.x release.

eswolinsky3241:
@selwin thanks will try to address by the end of this week. Do you have an estimate on when you’ll make the 2.x release so I have an idea of a deadline?

selwin:
I’m hoping to release 2.x by end of March.On Feb 21, 2024, at 8:14 AM, Ethan Wolinsky ***@***.***> wrote:﻿
@selwin thanks will try to address by the end of this week. Do you have an estimate on when you’ll make the 2.x release so I have an idea of a deadline?

tests/test_group.py
eswolinsky3241: @selwin I believe we did have this at some point and agreed to move it to `batch`
https://github.com/rq/rq/pull/1945#discussion_r1328026775

rq/worker.py
```diff
@@ -1489,6 +1493,10 @@ def handle_job_success(self, job: 'Job', queue: 'Queue', started_job_registry: S
                     # started_job_registry.remove(job, pipeline=pipeline)
                     # self.execution.delete(pipeline)  # type: ignore
 
+                    if job.group_id:
+                        group = Group.fetch(job.group_id, self.connection)
+                        group.cleanup(pipeline=pipeline)
```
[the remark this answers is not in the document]
eswolinsky3241: Removed

rq/worker.py
```diff
@@ -677,6 +680,10 @@ def handle_job_failure(self, job: 'Job', queue: 'Queue', started_job_registry=No
             else:
                 enqueue_dependents = True
 
+            if job.group_id:
+                group = Group.fetch(job.group_id, self.connection)
+                group.cleanup(pipeline=pipeline)
```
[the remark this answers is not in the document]
eswolinsky3241: Removed

rq/group.py
eswolinsky3241: Fixed

rq/group.py
eswolinsky3241: Fixed

rq/group.py
eswolinsky3241: @selwin I think it makes more sense here since it's not related to any one group. Happy to move it into a separate classmethod if this one's getting to busy though.

eswolinsky3241:
@selwin Addressed all comments, thanks!

rq/group.py
```diff
@@ -0,0 +1,124 @@
...
+        self.name = name if name else str(uuid4().hex)
+        self.connection = connection
+        self.key = '{0}{1}'.format(self.REDIS_GROUP_NAME_PREFIX, self.name)
+
+    def __repr__(self):
+        return "Group(id={})".format(self.name)
+
+    def _add_jobs(self, jobs: List[Job], pipeline: Pipeline):
+        """Add jobs to the group"""
+        pipe = pipeline if pipeline else self.connection.pipeline()
```
selwin, line 30: This line is no longer needed since `pipeline` is no longer optional.

rq/group.py
```diff
@@ -0,0 +1,124 @@
...
+        return jobs
+
+    def get_jobs(self) -> list:
+        """Retrieve list of job IDs from the group key in Redis"""
+        self.cleanup()
+        job_ids = [as_text(job) for job in self.connection.smembers(self.key)]
+        return [job for job in Job.fetch_many(job_ids, self.connection) if job is not None]
+
+    def delete(self):
+        self.connection.delete(self.key)
```
selwin, line 72: Can we add a test for this?

selwin: Or if this is not used, remove this entirely.

rq/group.py
```diff
@@ -0,0 +1,124 @@
...
+        """Return the Redis key of the set containing a group's jobs"""
+        return cls.REDIS_GROUP_NAME_PREFIX + name
+
+    @classmethod
+    def clean_registries(cls, connection: 'Redis'):
+        """Loop through groups and delete those that have been deleted.
+        If group still has jobs in its registry, delete those that have expired"""
+        groups = Group.all(connection=connection)
+        with connection.pipeline() as p:
+            if groups:
```
selwin, line 109: Since `Group.all()` always returns a list, can we remove this `if` block?

selwin:
Hopefully last round of review. No logic or new features, just minor nitpicky details! Thanks!

rq/group.py
eswolinsky3241: Fixed

eswolinsky3241:
@selwin addressed all, thanks!

selwin:
Thanks!

eswolinsky3241:
@selwin Curious if you have any ETA on when the next major release will be? We'd like to start using this feature with the official package rather than building our own

selwin:
The next major release will be RQ 2.0, which ships with a major rewriting of how job executions are handled. I'll be away for a few weeks in June so I plan to release it in July-August timeframe so I can better support the 2.0 release.

As far as I know though, the master branch is relatively stable, would appreciate if you can help beta test it :)

eswolinsky3241:
No problem, will start testing it out!

PR #1947, opened 2023-06-14
Do not run dependent jobs when parent is canceled

### LINKED ISSUE #1902: `cancel_job` does not cancel job
Here's my test code:

```python3
@rq.job(RQ_HIGH)
def async_long_running(sleep_timer, *args, **kwargs):
	"""
	Made to run inside the async worker.
	"""
	print(f'Arguments received: {args}')
	print(f'Keyword arguments received: {kwargs}')
	time.sleep(sleep_timer)
	return args, kwargs

def wait_and_fetch_results(job):
	"""
	Waits for a job to finish and returns the result.
	"""
	while not job.is_finished:
		time.sleep(0.1)
		if job.is_stopped:
			raise Exception("Job was stopped while running.")
		elif job.is_canceled:
			raise Exception("Job was canceled before running.")
		elif job.is_failed:
			raise Exception("Job failed while running.")
	return job.result

class Test_RQ(unittest.TestCase):
	def test_queue_dependent_jobs_with_cancelation(self):
		TIMER = 10
		args3 = ('arg5', 'arg6')
		kwargs3 = {'kwarg5': 'kwarg5', 'kwarg6': 'kwarg6'}
		args4 = ('arg7', 'arg8')
		kwargs4 = {'kwarg7': 'kwarg7', 'kwarg8': 'kwarg8'}
		args5 = ('arg9', 'arg10')
		kwargs5 = {'kwarg9': 'kwarg9', 'kwarg10': 'kwarg10'}
		job3 = async_long_running.delay(TIMER, *args3, **kwargs3, is_async=RUN_ASYNC, job_id='run_third', queue=Q_WITH_2_WORKERS)
		job4 = async_long_running.delay(TIMER, *args4, **kwargs4, is_async=RUN_ASYNC, job_id='run_fourth', depends_on=['run_third'], queue=Q_WITH_2_WORKERS)
		job5 = async_long_running.delay(TIMER, *args5, **kwargs5, is_async=RUN_ASYNC, job_id='run_fifth', depends_on=['run_fourth'], queue=Q_WITH_2_WORKERS)

		# While job3 is running, we'll cancel job4 and check if job5 gets enqueued
                self.assertFalse(job4.is_started)
		rq.cancel_job(job4.id, enqueue_dependents=True, connection=redis_conn)
		## Check that job4 is canceled
		self.assertTrue(job4.is_canceled)
		## Wait for job3 to finish
		results3 = wait_and_fetch_results(job3)
		self.assertEqual(results3, (args3, kwargs3))
		## Wait for job5 to finish
		results5 = wait_and_fetch_results(job5)
		self.assertEqual(results5, (args5, kwargs5))
		with self.assertRaises(Exception):
			results4 = wait_and_fetch_results(job4)
			print(results4)
```

The last assertion fails. Even though `job4` has the `is_canceled` set to True, it still runs and the result is fetched. :-/

fredsod:
I ran into a problem when we canceled a running parent job and the dependent jobs were executed.

When rq cancel a running job the status is changed in the database but when rq tries to enqueue the dependent jobs it seems like it does not fetch the new value from the database.

I do not know that much about the internals of RQ, could be better ways to fix this.

Probably fixes #1902

### DIFF
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -4,6 +4,7 @@ import shutil
 import signal
 import subprocess
 import sys
+import threading
 import time
 import zlib
 from datetime import datetime, timedelta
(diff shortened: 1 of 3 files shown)


selwin:
Hey I think a better fix would be what's suggested by @parthvnp here: https://github.com/rq/rq/issues/1902#issuecomment-1543324530

Attaching a passing test like what @amks1 described here is also necessary https://github.com/rq/rq/issues/1902#issue-1696852254

fredsod:
Hi, I have tried the fix proposed in #1902 but I can't get it to work for my test cases. It could be that the fix it's needed if you try to enqueue a job to an already canceled job. But I think there are multiple problems when trying to cancel both running and queued jobs.

I have updated the PR with two more test cases that fail on the master branch, I also added a check if the dependent job is canceled when a finished job tries to enqueue its dependent jobs.

tests/test_worker.py
```diff
@@ -594,6 +595,91 @@ def test_deleted_jobs_arent_executed(self):
...
+        status_thread.start()
+
+        w = Worker([q])
+        w.work(
+            burst=True,
+        )
+        status_thread.join()
+
+        self.assertNotEqual(parent_job.result, None)
+        self.assertEqual(job.get_status(), "deferred")
```
selwin, line 626: Mind using status enums for comparison like so? `self.assertEqual(job.get_status(), JobStatus.DEFERRED)`

tests/test_worker.py
```diff
@@ -594,6 +595,91 @@ def test_deleted_jobs_arent_executed(self):
...
+        def cancel_parent_job(job):
+            while job.is_queued:
+                time.sleep(1)
+
+            job.cancel()
+            return
+
+        q = Queue(
+            "low",
+        )
```
selwin, line 611: This should be one line, and please format this patch with black.

selwin:
> Hi, I have tried the fix proposed in https://github.com/rq/rq/issues/1902 but I can't get it to work for my test cases. It could be that the fix it's needed if you try to enqueue a job to an already canceled job. But I think there are multiple problems when trying to cancel both running and queued jobs.

You're right, the bug you described happens because the parent job itself doesn't know it has been canceled in the background. This PR looks good, but I left some comments just to make sure the patch is stylistically consistent with the rest of the code base.

PR #1949, opened 2023-06-15
Stop calling deprecated Connection functions `push_connection`, `pop_connection`, `resolve_connection`, `parse_connection`

### LINKED ISSUE #1948: `perform_job` calls `push_connection` which is a deprecated function
in `worker.py`, the `perform_job` calls the function `push_connection` which raises a `DeprecationWarning`

PrenSJ2:
Fix #1948﻿

### DIFF
--- a/rq/connections.py
+++ b/rq/connections.py
@@ -38,11 +38,11 @@ def Connection(connection: Optional['Redis'] = None):  # noqa
     )
     if connection is None:
         connection = Redis()
-    connection or push_connection(connection)
+    connection or LocalStack().push(connection) or push_connection(connection)
     try:
         yield
     finally:
-        popped = pop_connection()
+        popped = LocalStack().pop() or pop_connection()
         assert (
             popped == connection
         ), 'Unexpected Redis connection was popped off the stack. Check your Redis connection setup.'
@@ -112,7 +112,7 @@ def resolve_connection(connection: Optional['Redis'] = None) -> 'Redis':
     if connection is not None:
         return connection
 
-    connection = get_current_connection()
+    connection = LocalStack().top() or get_current_connection()
     if connection is None:
         raise NoRedisConnectionException('Could not resolve a Redis connection')
     return connection
--- a/rq/contrib/legacy.py
+++ b/rq/contrib/legacy.py
@@ -1,6 +1,7 @@
 import logging
 
 from rq import Worker, get_current_connection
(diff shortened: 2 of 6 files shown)


PrenSJ2:
if
 ```
    if redis is None:
        redis = Redis()
    push_connection(redis)
    ```
    is equal to 
    ```
	Queue(connection=redis)
    ```
    :thinking:

PrenSJ2:
@ccrvlh I see you added the deprecation comments a while ago, I am trying to solve this issue however there is nothing clear in the docs or your PR to suggest how we should use connection explicitly instead of the Connection functions, any comments or suggestions would be greatly appreciated.

ccrvlh:
Hi @PrenSJ2.

In the PR #1860 there's a description about what was done:

> Most of the functions on the connections module are used to interact with a Redis connection implicitly through the LocalStack: push/pop/use/resolve_connection.
> 
> The connection ideally should be handled explicitly and the with Connection already has a deprecation warning. This adds deprecation warnings to those functions. After these are deprecated, we would be able to delete the local.py entirely, and only handle connections explicitly.

The docs also discuss ways of handling connections [here](https://python-rq.org/docs/connections/). When we worked on that PR, the connection `context_manager` was already deprecated, and explicit connection handling (passing the connection as an argument to the `Queue` and other objects) was already being suggested in the docs for a while.

It would probably make sense to also add a deprecation warning to the docs, and slowly refactor the internals to use the explicit connection handling pattern everywhere (mostly the `tests` if I remember correctly). 

> need to 🤔 think of how I can apply this to Connection's methods [push_connection, pop_connection, resolve_connection, parse_connection]

I don't remember the details and every case where the connection is being `pushed`/`popped` but ideally you won't do anything with those methods. If you're explicitly handling connection, they shouldn't be necessary anymore (those are all there for the "implicit" connection handling).

Unfortunately I haven't been having much time to work here (hopefully that will change in the future), but let me know if you have any doubts, I'll try to help as I can.

selwin:
I don't think we should change this behavior until version 2.0, where `use_connection` will be fully deprecated. Mind changing the PR so that we simply stop using `push_connection` [in our testcases](https://github.com/rq/rq/pull/1949/files#diff-001e61d97d9dee27d0c5aa1f23ba6c1972cd8a2e8320bac839656b0ab935b84dL80)?

PR #1950, opened 2023-06-19
Update CHANGES.md to show change in Redis library requirement

nick-merrill:
Version 1.15 added a new requirement for Redis 4.0.0 or greater. I didn't find out about this requirement until updating our dependencies. I think it would be helpful to add to the changelog.

Thank you for maintaining this wonderful library.


selwin:
Ah, I don't think I bumped the minimum redis-py version intentionally. 😅

I was thinking of supporting redis-py 3.5 longer because it works with Redis server < 5. If no one complains, I'll merge this in later. Requiring redis-py >= 4 also means we can stop supporting Python 3.6.

PR #1951, opened 2023-06-22
Add the Python dev packages for Python headers to tests/Dockerfile

robhudson:
The Docker tests were failing for me since psutil was trying to compile and looking for Python headers. With the python dev packages added these now work again.

Python 3.6 is skipped during the test runs presumably because the deadsnakes packages no longer build Python 3.6. So this could be removed later as part of an overall update to drop support. I'm leaving it as-is for now.

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8.3", "3.9", "3.10", "3.11"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
         redis-version: [3, 4, 5, 6, 7]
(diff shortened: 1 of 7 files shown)


selwin:
Let's drop Python 3.6 this release since it's been EOL more than a year ago.

robhudson:
Okay, I removed Python 3.6 from the testing workflows.

selwin:
Thanks!

robhudson:
> Let's drop Python 3.6 this release since it's been EOL more than a year ago.

And Python 3.7 security support [ended 2 days](https://endoflife.date/python) ago (Jun 27) 😱

selwin:
> And Python 3.7 security support [ended 2 days](https://endoflife.date/python) ago (Jun 27) 😱

Yeah, but I'd like to keep 3.7 compatibility for a bit more, as long as it's not high maintenance or there's a good reason to drop it (like packaging support in 3.6).

PR #1952, opened 2023-06-23
Store project metadata in pyproject.toml (PEP 621)

robhudson:
An alternative PR to #1864 using [hatch](https://hatch.pypa.io/latest/) and only focusing on removing `setup.py` and updating necessary parts in workflows etc because of this switch.

Happy to continue if there's any feedback or requested changes, but also happy if this is just an exercise and the other PR continues to move forward.


selwin:
I'd like to minimize build dependencies if possible. If we were to store project metadata in `pyproject.toml`, what alternatives do we have other than hatch?

This will be my first project moving to `pyproject.toml` so I don't know much about tooling in this area.

robhudson:
I found this article to be a good overview:
[The Basics of Python Packaging in Early 2023](https://drivendata.co/blog/python-packaging-2023)

The other [pull request](https://github.com/rq/rq/pull/1864) uses `flit-core`. I personally found `hatch` to have better documentation and is well received in the Python community, which is why I opened this PR as an alternative to compare against. Feel free to choose what you prefer or would work best for this project.

selwin:
> I found this article to be a good overview:
[The Basics of Python Packaging in Early 2023](https://drivendata.co/blog/python-packaging-2023)

Thanks, this is a useful read. Let's proceed with Hatch. I'll review this PR again a bit later.

.github/workflows/dependencies.yml
```diff
@@ -33,13 +33,12 @@ jobs:
...
         python -m pip install --upgrade pip
+        pip install hatch
         pip install redis==${{ matrix.redis-py-version }}
-        pip install -r requirements.txt -r dev-requirements.txt
-        pip install -e .
 
     - name: Test with pytest
       run: |
-        RUN_SLOW_TESTS_TOO=1 pytest --durations=5
+        RUN_SLOW_TESTS_TOO=1 hatch run test:pytest --durations=5
```
selwin, line 41: Any benefit to using `hatch run test:pytest` instead of calling `pytest` directly? If there's no big benefit I'd rather keep calling `pytest` directly, the less indirection the better.

pyproject.toml
```diff
@@ -1,6 +1,100 @@
...
+    "README.md",
+    "requirements.txt",
+    "tox.ini",
+]
+
+[tool.hatch.envs.test]
+dependencies = [
+  "black",
+  'coverage; python_version >= "3.7"',
+  'coverage==6.2; python_version == "3.6"',
```
selwin, line 82: We no longer need to support Python 3.6.

.github/workflows/dependencies.yml
robhudson: The benefit is that it allows us to define the Python dependencies in the pyproject.toml file in the `[tool.hatch.envs.test]` section and `hatch` creates the virtual environment with those dependencies installed automatically for us, including `rq` itself. This is why we can drop the `pip install` commands above.

pyproject.toml
robhudson: Okay. I can do another pass to make sure I remove this plus some others I see.

selwin:
Thanks!

ccrvlh:
Is there a `CONTRIBUTING` guide to handle Hatch in RQ? I just used `hatch run..` to run the tests, but tooling (eg. `mypy`) doesn't seem to be properly configured when just using `hatch run` (eg. VSCode MyPy extension doesn't find MyPy in the active intepreter, etc).

PR #1954, opened 2023-06-29
Fix bug with stopped callback in enqueue_many

eswolinsky3241:
stopped_callback was not being copied from the EnqueueData instance to the Job instance when using `enqueue_many`.


selwin:
Mind adding a quick test case for this @eswolinsky3241 ?

eswolinsky3241:
@selwin Added the test to test_callbacks

selwin:
Thanks!

PR #1964, opened 2023-08-12
Job with multiple executions

selwin:
This PR introduces an `Execution` object to represent a job's execution. Completion of this PR should allow us to properly track multiple job executions.


rq/job.py
```diff
@@ -55,6 +56,13 @@ class JobStatus(str, Enum):
     CANCELED = 'canceled'
 
 
+def parse_job_id(job_or_execution_id: str) -> str:
+    """Parse a string and returns job ID. This function supports both job ID and execution composite key."""
+    if ':' in job_or_execution_id:
```
dams, line 61: This is a problem: when creating a standard job, you can pass a job_id to it, so that instead of using the default uuid, you can use your own custom job id convention.
We use that, and we already use `:` and `|` as special separator in the job id. This code change breaks the tacit contract and now a job id can't contain some characters. This could break stuff in production for some users (it will for us...)
Can't the execution id be stored somewhere else ?

PR #1969, opened 2023-08-29
Allow CA be passed as a variable

elbuo8:
Currently SSL CA can only be passed in via a file in disk. `redis-py` allows passing the value as a string.


elbuo8:
Weird that linting fails on a fresh work 🤔

selwin:
Do you mind documenting this option?

elbuo8:
@selwin 👍🏽

elbuo8:
@selwin 👋🏽

selwin:
All tests need to pass before I can merge this in

elbuo8:
Could you re-trigger? The failures are not related to the change 🤔

selwin:
Merged, thanks!

PR #1970, opened 2023-08-31
Don't use `os.setsid()` after forking workhorse

craigds:
In 9adcd7e50c511ceba622ba20c826a0b701a917ea a change was made to make the workhorse process into a session leader. This appears to have been done in order to set the process group ID of the workhorse so that it can be killed easily along with its descendants when `Worker.kill_horse()` is called.

However, `setsid` is overkill for this purpose; it sets not only the process group ID but also the session ID. This can cause issues for user jobs which may rely on session IDs being less-than-unique for arbitrary reasons.

This change switches to `setpgrp`; this sets the process group ID so that the workhorse and its descendants can be killed *without* changing the session ID.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -425,7 +425,7 @@ class BaseWorker:
     @property
     def dequeue_timeout(self) -> int:
         return max(1, self.worker_ttl - 15)
-    
+
     @property
     def connection_timeout(self) -> int:
         return self.dequeue_timeout + 10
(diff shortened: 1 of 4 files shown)


selwin:
@craigds would this change negatively affect use cases outlined in the original PR? https://github.com/rq/rq/pull/1348

craigds:
no - `setpgrp` still sets the process group ID, which is all that is needed for killing a process and all of its descendents.

`setsid` not only sets the process group but also the session, which is a more powerful concept which isn't actually required for the purpose of killing related processes

selwin:
Thanks!

PR #1974, opened 2023-09-09
Allow worker version to be overriden

itzsimpl:
Currently `worker.version` is hardcoded to RQ version. This PR allows overriding the `worker.version` and for such cases keeps a reference to RQ version also as a separate variable `worker.rq_version`. By default both `worker.version` and `worker.rq_version` are set to the current RQ version so as to keep the current behaviour unmodified.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -125,6 +125,7 @@ class BaseWorker:
         self,
         queues,
         name: Optional[str] = None,
+        version: Optional[str] = VERSION,
         default_result_ttl=DEFAULT_RESULT_TTL,
         connection: Optional['Redis'] = None,
         exc_handler=None,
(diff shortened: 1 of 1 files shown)


selwin:
Hello, may I know what's the use case for this? I don't think making the version number user configurable is not such a good idea.

itzsimpl:
Hello Selwin. Assume one has a request receiving API, and multiple workers. The receiving API receives requests for jobs, these are pushed into RQ and the workers consume and execute jobs. Assume the receiving API is run in one container, the Redis database as a separate container, and each worker (multiple of them) is also run as a separate container (these can be all on a single server or distributed over multiple servers). 

Assume all workers perform the same job. Code changes to the workers code base (resolved bugs, optimisations, etc.) that are unrelated to the RQ version are important, and there are occasions when multiple worker versions will coexist (e.g. testing if a bug fix resolves the issue, ...). The worker version (e.g. the version of the job execution codebase, not the RQ version) is in these cases important. When multiple versions of workers coexist, the worker's version is also important from the job's perspective (e.g. the worker's - job execution codebase - version may affect the "quality" of the result/output).  

Making the worker version configurable, allows developers to pass the information about the specific worker version (e.g. version of the job execution codebase) that is running in an individual container to the receiving API (if they choose so), without resorting to worker class overrides. Note that the PR only relaxes the restriction; if the developers do not choose to override the version, the current behaviour is left unchanged, if they do, the RQ version is still available, but under a more descriptive variable `worker.rq_version`.

PR #1983, opened 2023-09-15
Add support for functools.partial in Callback class

jtfidje:
This commit adds support for `functools.partial` objects in the `Callback` class. The `__init__` method is updated to accept `functools.partial` objects in addition to strings, functions, and built-in functions. This extends the flexibility of the `Callback` class, allowing function arguments to be fixed in advance.


jtfidje:
I realize now that partials can't be serialized / deserialized as-is

Issue #1645, opened 2022-04-12
Support custom parameters to `on_success` and `on_failure` callbacks


zx8:
It would be nice if these functions accepted additional parameters so we could e.g. pass in reusable database clients/connections and avoid having to initialize them inline, when they are invoked..

One idea I had was to "smuggle" the database connection string within the `result` parameter of these functions, but it's a hacky way of achieving the desired result.

Thoughts?

selwin:
In general, I support the idea of passing custom parameters to `on_success` and `on_failure` callbacks, but it comes down to whether we can do it elegantly. If you have some ideas on how we can achieve this, let me know before you work on a PR.

interrogator:
I would really appreciate this, for the same reason as @zx8 , I would like access to a non-picklable database connection.

Could the `args` and `kwargs` that get passed to jobs be stored without pickling, and then made accessible as `*args` and `**kwargs` in the `on_success` and `on_failure` callbacks?

selwin:
I’d be happy tonaccept a PR implementing custom args.

interrogator:
I looked into the source, if I understand correctly, the callbacks are actually run by the worker? If so, I don't know if it's possible to pass non-picklable arguments to them from the non-worker...

For @zx8 , if it's a single DB connection that doesn't change from startup, I figured out one way to do it ... you can run a customised worker that runs code like this:

```python
    class MyJob(Job):
        def __init__(self, *args, **kwargs):
            super().__init__(*args)
            self._db_conn = some_expensive_db_connection()


    class MyWorker(Worker):
        job_class = MyJob

    with Connection():
        w = MyWorker(['default'])
        w.work()
```

Then during your callback you can use `get_current_job()._db_conn` to access the connection object.

Unfortunately for me, I need access to an object that gets mutated by the non-worker, which is where I'm hitting problems.

Manouchehri:
> I looked into the source, if I understand correctly, the callbacks are actually run by the worker? If so, I don't know if it's possible to pass non-picklable arguments to them from the non-worker...

The callbacks are executed by the rq worker, not the thread that creates the Queue/Job. This isn't super well documented (but is a bit obvious once you grab coffee and realize that your error statements are occurring in a different terminal.)

> Unfortunately for me, I need access to an object that gets mutated by the non-worker, which is where I'm hitting problems.

In my case, I'm trying to notify the non-Worker about the result of a Worker (without resorting to polling).

PR #1990, opened 2023-10-09
add: custom parameters

jazken:
Just started using this library and hope to have the feature to pass in custom parameters into callbacks.

#1645 Another issue highlighting this feature.
I am setting up an initial pull request to add params to callbacks.

The method of implementation is to add an optional dictionary params to the callback class, allowing any type of callback to use this to expand parameters for callbacks.


jpsnyder:
Call me naive, but couldn't you just use `functools.partial` to accomplish this?

selwin:
Can you add tests?

Issue #1602, opened 2021-12-08
Orphaned workhorse recreates redis key of the killed worker with TTL `-1`


kusaku:
When a worker terminates with `kill -9` (for example, `suprevisor` can terminate workers with `SIGKILL`), the workhorse will continue running.

Workhorse may or may finish or fail. In both cases it will execute some stats methods:

- `Worker.increment_successful_job_count()`
- `Worker.increment_failed_job_count()`
- `Worker.increment_total_working_time()`

These methods will execute redis `HINCRBY` and `HINCRBYFLOAT` on a non-existent key of killed worker (key of worker is deleted by TTL), for absent key redis will set the standard ttl `-1`and that key will become immortal.`

A check could be added to see if the worker still exists before updating the stats.

selwin:
Agreed. However, the checks shouldn't be done everytime the keys are incremented as this will incur one extra Redis operation for every job that finishes.

The check should only be performed when  `increment_successful_job_count()` or `increment_failed_job_count()` returns `1` (meaning it's possibly a new key).

Do you mind opening a PR for this?

PR #1993, opened 2023-10-19
Eliminating Orphan Workers

npgoossen:
Issue #1602 mentions that if a worker crashes with a work horse still running, the work horse process, when the job finishes will attempt to update stats. 
By default Redis creates keys with no expiration, resulting in a new worker key being created that makes it so you can't re-register a worker with the same key. The worker is also in an odd state, neither alive or dead. 

Mentioned in the issue is a suggesting to optimize this to avoid an extra Redis call by checking if the value was set to 1, but I'm not sure this will work as that will also catch the situation of the first job a worker runs, resulting in the number never being incremented.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -1076,33 +1076,44 @@ class BaseWorker:
 
     def increment_failed_job_count(self, pipeline: Optional['Pipeline'] = None):
         """Used to keep the worker stats up to date in Redis.
-        Increments the failed job count.
+        Increments the failed job count. Only if the worker still exists.
+        For long running tasks the worker might have been removed already,
+        and the work horse is still running.
 
         Args:
             pipeline (Optional[Pipeline], optional): A Redis Pipeline. Defaults to None.
         """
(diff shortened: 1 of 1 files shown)


selwin:
> Mentioned in the issue is a suggesting to optimize this to avoid an extra Redis call by checking if the value was set to 1, but I'm not sure this will work as that will also catch the situation of the first job a worker runs, resulting in the number never being incremented.

I'm not sure this will work because `if connection.exists(self.key):` will be pipelined. I have recently changed my way of thinking, I think we should simply set a TTL to the worker key after the increment calls are successful. This way, no worker keys would be orphaned forever.

Mind making this change?

npgoossen:
@selwin  - Thanks for the comment/review. Changed it to set the expiration when these values get set. Shouldn't have any adverse affects on normal operation as the heartbeat keeps increasing it.

selwin:
@npgoossen can you check the failing tests? They all need to pass before I can merge this in.

PR #1998, opened 2023-11-03
Prepare v2 docs

selwin:


ccrvlh:
@selwin a while ago I started working on v2 branch, and intentionally deviated from the main branch as to have more flexibility. My idea is for it to be a branch for experiments, so you can take a look and decide on what to add to the main project: https://github.com/ccrvlh/rq/tree/v2, take a look to see what you think - i tried to keep the changelog descriptive so it may be a starting point to see the changes.

selwin:
@ccrvlh thanks, I'll have a look sometime in the next few days. As per usual, if there's anything that you want to upstream to the main repo, please do so in a piece meal fashion so we can keep RQ stable.

ccrvlh:
Sure, the idea was just a place for experiments, I find it easier to draft ideas without much restrictions. In the same direction, if there's anything you like from the v2 branch we can cherry pick and work on isolated PRs in a way to make it backwards compatible.

PR #2003, opened 2023-11-08
For upstream/avoid useless cleanup

dams:
It's impacting performance, and it's redundant with the workers' maintenance task that already calls `cleanup`

### DIFF
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -88,7 +88,6 @@ class BaseRegistry:
         Returns:
             int: _description_
         """
-        self.cleanup()
         return self.connection.zcard(self.key)
 
     def add(self, job: 'Job', ttl=0, pipeline: Optional['Pipeline'] = None, xx: bool = False) -> int:
@@ -163,7 +162,6 @@ class BaseRegistry:
         Returns:
             _type_: _description_
         """
-        self.cleanup()
         return [as_text(job_id) for job_id in self.connection.zrange(self.key, start, end)]
 
     def get_queue(self):
@@ -385,12 +383,6 @@ class DeferredJobRegistry(BaseRegistry):
 
     key_template = 'rq:deferred:{0}'
 
-    def cleanup(self):
-        """This method is only here to prevent errors because this method is
-        automatically called by `count()` and `get_job_ids()` methods
-        implemented in BaseRegistry."""
-        pass
-
 
 class ScheduledJobRegistry(BaseRegistry):
     """
@@ -418,12 +410,6 @@ class ScheduledJobRegistry(BaseRegistry):
 
         timestamp = calendar.timegm(scheduled_datetime.utctimetuple())
         return self.connection.zadd(self.key, {job.id: timestamp})
-
-    def cleanup(self):
-        """This method is only here to prevent errors because this method is
-        automatically called by `count()` and `get_job_ids()` methods
-        implemented in BaseRegistry."""
-        pass
 
     def remove_jobs(self, timestamp: Optional[datetime] = None, pipeline: Optional['Pipeline'] = None):
         """Remove jobs whose timestamp is in the past from registry.
@@ -479,12 +465,6 @@ class CanceledJobRegistry(BaseRegistry):
 
     def get_expired_job_ids(self, timestamp: Optional[datetime] = None):
         raise NotImplementedError
-
-    def cleanup(self):
-        """This method is only here to prevent errors because this method is
-        automatically called by `count()` and `get_job_ids()` methods
-        implemented in BaseRegistry."""
-        pass
 
 
 def clean_registries(queue: 'Queue'):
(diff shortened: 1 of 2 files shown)


selwin:
`cleanup()` is called within count because sometimes this command is called without any workers running. As far as I know, `cleanup()` executes very quickly, even in registries with lots of jobs but I do understand it incurs a few extra Redis calls.

If you really need to disable this, I suggest adding a `cleanup=True`  argument, you can then pass in `False` so it doesn't call `cleanup()` needlessly.

dams:
> `cleanup()` is called within count because sometimes this command is called without any workers running.

Ah that's a good point.

> As far as I know, `cleanup()` executes very quickly, even in registries with lots of jobs but I do understand it incurs a few extra Redis calls.

Well, "executes very quickly" doesn't tell me a lot 🙂 
The code complexity of `cleanup` is in `O(2*log(N)+3*M)` with N the number of jobs in the registry and M the number of expired jobs.
We use RQ with up to millions of jobs, for instance in failed registries. Even if there is nothing to expire, hitting `O(2*log(N)` every time we call `count` is not possible, it has a real performance hit, both response time and CPU usage on Redis.
Having a `count` property that is *not* in `O(1)` is very counter intuitive in my opinion.

> 
> If you really need to disable this, I suggest adding a `cleanup=True` argument, you can then pass in `False` so it doesn't call `cleanup()` needlessly.

It's not a very usable solution, as `count` is used as a property, and it's also used when doing `len` so it's very difficult to prevent developers to trigger needless `cleanup` calls.

What if instead, `count` check if there is a worker consuming the queue, for instance using:

```
if len(worker_registration.get_keys(queue=self)) == 0 :
    self.cleanup()
```

Would you be happy with that?

selwin:
> It's not a very usable solution, as count is used as a property, and it's also used when doing len so it's very difficult to prevent developers to trigger needless cleanup calls.

Sorry, I forgot that it's a property. In that case do you mind opening a PR that adds a `get_job_count()` method that returns the number of jobs in registry without calling `cleanup()`?

> Having a count property that is not in O(1) is very counter intuitive in my opinion.
Perhaps, but it returns the correct number everytime and works for 99% users. I think it's ok to design something that works well for the vast majority of users. RQ works well even with millions of jobs in queue/registry, but they probably don't call `len()` as often as you do.

dams:
Thanks for your answer, I understand your position.
I think I'll provide a different PR to enable/disable the "cleanup when counting" behaviour at the queue level, but still cleanup when there is no worker associated.
Would that be acceptable?

selwin:
I think implementing `get_job_count()` would be simpler, let's do that first and see if that fixes your issue.

sophcass:
Somewhat related to this PR, this clean up also causes issues when calling [get_statistics()](https://github.com/rq/django-rq/blob/master/django_rq/utils.py#L44) using a custom serializer for the queue.

When `len()` is called [here](https://github.com/rq/django-rq/blob/master/django_rq/utils.py#L87), `count` is called. Count then calls `cleanup()` [here](https://github.com/rq/rq/blob/133d18dfb24d56a1f2781335584ed8015cac2dfb/rq/registry.py#L91), which sets the [incorrect serializer](https://github.com/rq/rq/blob/133d18dfb24d56a1f2781335584ed8015cac2dfb/rq/registry.py#L243), resulting in pickling errors.

One solution is to remove this cleanup in `count`, as is discussed in this PR. Or alternatively, we can initiate the registries with Queue instances in `get_statistics()` [here](https://github.com/rq/django-rq/blob/master/django_rq/utils.py#L82-L86), which ensures the custom serializer is used. 

I can open a PR for the second option if that is the preferred solution. Let me know!

anton-daneyko-ultramarin:
@dams I noticed the branch is called "For upstream..." I assume there's a downstream somewhere. How do you go by without having the change you need merged? Do you build a package internally or use `git submodule` to pull RQ into the project that uses it? I am exploring my options for the case the release is not going to be made in the near future.

dams:
> @dams I noticed the branch is called "For upstream..." I assume there's a downstream somewhere. How do you go by without having the change you need merged? Do you build a package internally or use `git submodule` to pull RQ into the project that uses it? I am exploring my options for the case the release is not going to be made in the near future.

We are using our [fork here](https://github.com/alan-eu/rq/tree/alan) which contains our fixes for high volume, high scale usage of RQ. We stopped rebasing on upstream as new features don't always take in account performance at large scale

anton-daneyko-ultramarin:
@dams
Thanks, for sharing.

I see you have "[automatically retry after an AbandonnedJobError](https://github.com/alan-eu/rq/commit/4f472785565b218f0daaa3631e0d786b1846a7ec)", which is something I wish to have as well and currently trying to do with `on_failure` callback, which crashes my FastAPI application because the callback is not [executed in the main thread](https://github.com/rq/rq/pull/2104#issuecomment-2397114656).

With your own fork, how bundle it with the app that uses it? Do you create a package and host it privately, so the app code can pull it or do you bundle your forked RQ source together with your app?

Issue #2006, opened 2023-11-14
Dependencies specified with Dependency(allow_failure=True) are never enqueued when a job is moved to FailedJobRegistry due to AbandonedJobError


Marishka17:
As stated in the documentation, since version 1.11.0 it is possible to enqueue jobs even if their dependency has failed.
It can be specified with *Dependency(allow_failure=True)*. However, that does not work in one case. 
For example, we have 2 jobs `X1` and `X2`. X2 depends on X1 and can be executed if X1 fails. The code is executed in the docker container. If in this case, the docker container is restarted when X1 is running, then X1 will be moved to `FailedJobRegistry` due to AbandonedJobError after the restart. However, there is no code that will start deferred jobs if needed in the `registry.StartedJobRegistry.cleanup`. Looks like this can be fixed by adding `queue.enqueue_dependents(job)` when handling expired jobs.

rq version: **1.15.1**

PR #2008, opened 2023-11-26
Enqueue dependents when job is abandoned and moved to FailedJobRegistry

Marishka17:
This PR fixes the issue https://github.com/rq/rq/issues/2006


rq/registry.py
```diff
@@ -270,6 +270,7 @@ def cleanup(self, timestamp: Optional[float] = None):
                         job.save(pipeline=pipeline, include_meta=False)
                         job.cleanup(ttl=-1, pipeline=pipeline)
                         failed_job_registry.add(job, job.failure_ttl)
+                        queue.enqueue_dependents(job)
```
selwin, line 273: You'll need to check whether the parent job has `allow_failure=True`. Please add a test case for this as well.

Marishka17: @selwin, I'm not sure why I need to check whether the parent job of this `job` has `allow_dependency_failures == True` or not.

Let's imagine a situation when there are 3 jobs in such sequence:
J1  <- J2 (`allow_dependency_failures=False`) <- J3 (`allow_dependency_failures=True`)

If `J1` is finished and `J2` is started, then after docker container stops, `J2` should be moved to `FailedJobRegistry`, but `J3` should be added to the queue (despite that `J2` as a parent job has `allow_dependency_failures==False`
If `J1` fails, then `J2` will not be started at all.
In addition, If a job has been started previously (because it is in the StartedJobRegistry), that means it has already passed the check on whether it can run or not, based on the parent job.

This PR does not fix the case when `J1` is started, the docker container stops, and `J1` is moved to the `FailedJobRegistry` while `J2` and `J3` remain in the `DeferredJobRegistry` because such a bug also occurs when the `StartedJobRegistry.cleanup` method is not called.

For instance:
```python
job1 = queue.enqueue_call(div_by_zero)
job2 = queue.enqueue_call(say_hello, depends_on=Dependency(jobs=job1, allow_failure=False))
job3 = queue.enqueue_call(say_hello, depends_on=Dependency(jobs=job2, allow_failure=True))
worker.work(burst=True)
job1.is_failed # True
job2.is_deferred # True
job3.is_deferred # True
```
I guess that in the example mentioned above, `J1` and `J2` should finally be moved to `FailedJobRegistry`, but `J3` should be added to `FinishedJobRegistry`.

selwin: You're right.

tests/test_registry.py
```diff
@@ -298,6 +298,47 @@ def test_get_queue(self):
         registry = StartedJobRegistry('foo', connection=self.connection, serializer=JSONSerializer)
         self.assertEqual(registry.get_queue(), Queue('foo', connection=self.connection, serializer=JSONSerializer))
 
+    def test_enqueue_dependents_when_parent_job_is_abandoned(self):
+        """Enqueuing parent job's dependencies after moving it to FailedJobRegistry due to AbandonedJobError."""
+        queue = Queue(connection=self.testconn)
+        worker = Worker([queue])
+        failed_job_registry = FailedJobRegistry(connection=self.testconn)
+        finished_job_registry = FinishedJobRegistry(connection=self.testconn)
```
selwin, line 306: Can we change all `self.testconn` to `self.connection`?

selwin: `self.testconn` will be deprecated in RQ 2.0.

PR #2010, opened 2023-12-03
Fix 'quiet' option for 'worker_pool' comand

Mindiell:
the _quiet_ option for **worker_pool** is not working correctly (as **worker** does).

### DIFF
--- a/rq/cli/cli.py
+++ b/rq/cli/cli.py
@@ -499,6 +499,7 @@ def worker_pool(
     else:
         job_class = Job
 
+    # if --verbose or --quiet, override --logging_level
     if verbose or quiet:
         logging_level = None
 
--- a/rq/cli/cli.py
+++ b/rq/cli/cli.py
@@ -499,6 +499,9 @@ def worker_pool(
     else:
         job_class = Job
 
+    if verbose or quiet:
+        logging_level = None
+
     pool = WorkerPool(
         queue_names,
         connection=cli_config.connection,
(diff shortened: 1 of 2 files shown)


selwin:
Mind adding a simple test case for this? Don't want the coverage to go down :)

Mindiell:
Hey, I would love to add some test, but it seems there is no for the already existing "quiet" option. And I'm clearly not an expert :/

rq/cli/cli.py
```diff
@@ -499,6 +499,9 @@ def worker_pool(
     else:
         job_class = Job
 
+    if verbose or quiet:
```
selwin, line 502: Why is `logging_level` set to `None` if `verbose` is passed?

selwin:
> If you explain to me how to write such a test, I can try !

You can use this example in [test_cli.py](https://github.com/rq/rq/blob/master/tests/test_cli.py#L836) that tests that job class argument is passed correctly.

rq/cli/cli.py
Mindiell: Line 297 : you do exactly the same. I jut copy-pasted this code from "worker" command.
As said, I'm not an expert but I can read code !

Mindiell:
> > If you explain to me how to write such a test, I can try !
> 
> You can use this example in [test_cli.py](https://github.com/rq/rq/blob/master/tests/test_cli.py#L836) that tests that job class argument is passed correctly.

I'll give it a try ! Thx !

rq/cli/cli.py
selwin: Ok, let's merge this in then :)

selwin:
Thanks!

Mindiell:
Thx for the help, and for the tool !

PR #2012, opened 2023-12-06
Add logging for job retries

duc00:
This PR improves observability of job workflows. 

It is quite painful to easily figure whether a job has been retried or is done retrying, after a failure. Currently not possible via looking at logs. The PR helps with that by logging when the job has been enqueued/scheduled as part of a retry, and with how many retries left.

Hope this helps!

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -1446,7 +1446,7 @@ class Job:
             with death_penalty_class(self.failure_callback_timeout, JobTimeoutException, job_id=self.id):
                 self.failure_callback(self, self.connection, *exc_info)
         except Exception:  # noqa
-            logger.exception(f'Job {self.id}: error while executing failure callback')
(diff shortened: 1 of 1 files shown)


kuwv:
@duc00 Don't use f-strings for logging. It is potentially a security vulnerability depending on the job content.

duc00:
@kuwv will update.

There are other occurrences of f-strings in this file, thus I just copied what seemed to be the pattern. Should I update those statements too?

Also how can f-strings not be safe in this case?

kuwv:
F-strings are a powerful templating tool that allows scope inspection and code execution.

https://bugs.python.org/issue46200#msg409499

selwin:
@duc00 please update the PR so I can merge this in :)

duc00:
@selwin @kuwv done! 👍

selwin:
Thanks!

Issue #2005, opened 2023-11-09
strange interactions between last_cleaned_at, maintenance_interval and rq:clean_registries lock key


dams:
Hi,

`run_maintenance_tasks` runs if `should_run_maintenance_tasks` returns `True`.
`should_run_maintenance_tasks` uses the `last_cleaned_at` attribute of the worker instance and compare it with `maintenance_interval`.
However if the worker just started, `last_cleaned_at` is None and `should_run_maintenance_tasks` returns True immediately. This could cause `should_run_maintenance_tasks` to return True too quickly if workers are short-lived
👉 Why not store `last_cleaned_at` as a redis key with the worker id as prefix?

the maintenance task runs the `clean_registries` method which takes a lock on queues. However the lock is implemented in a strange way, it's a simple redis key `rq:clean_registries:%s` with a TTL of 899, but it's never deleted. So the `clean_registries` cannot run more often than 899 sec, even if the `maintenance_interval` is smaller
👉 Why not delete the lock once the maintenance task has been done?

Thanks!

cc @tchapi

selwin:
> However if the worker just started, last_cleaned_at is None and should_run_maintenance_tasks returns True immediately. This could cause should_run_maintenance_tasks to return True too quickly if workers are short-lived
👉 Why not store last_cleaned_at as a redis key with the worker id as prefix?

I stored this as a local worker attribute so we can eliminate an extra Redis call everytime this is checked (if you have many short lived jobs, workers will perform an extra Redis call everytime they finish). But I think your suggestion is correct, we can also add a Redis key that's shared across all workers.

> the maintenance task runs the clean_registries method which takes a lock on queues. However the lock is implemented in a strange way, it's a simple redis key rq:clean_registries:%s with a TTL of 899, but it's never deleted. So the clean_registries cannot run more often than 899 sec, even if the maintenance_interval is smaller
👉 Why not delete the lock once the maintenance task has been done?

The lock should be deleted once maintenance task is finished. Mind opening a PR for this?

PR #2020, opened 2023-12-30
Fix maintenance task issues raised in #2005

dams:
- share 'last_cleaned_at' between workers and set it at maintenance task start time
- release maintenance lock when maintenance task is finished

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -401,12 +401,12 @@ class BaseWorker:
 
     @property
     def should_run_maintenance_tasks(self):
-        """Maintenance tasks should run on first startup or every 10 minutes."""
-        if self.last_cleaned_at is None:
-            return True
-        if (utcnow() - self.last_cleaned_at) > timedelta(seconds=self.maintenance_interval):
-            return True
-        return False
+        """
+        Maintenance tasks should run on first startup or every maintenance
+        interval (defaults to 10 minutes).
+        """
+        l_c_at = self.last_cleaned_at = self.last_cleaned_at or worker_registration.get_last_cleaned_at(self.connection)
+        return l_c_at is None or utcnow() - l_c_at > timedelta(seconds=self.maintenance_interval)
 
     def _set_connection(self, connection: Optional['Redis']) -> 'Redis':
         """Configures the Redis connection to have a socket timeout.
@@ -434,6 +434,8 @@ class BaseWorker:
 
     def clean_registries(self):
         """Runs maintenance jobs on each Queue's registries."""
+        self.last_cleaned_at = utcnow()
+        worker_registration.set_last_cleaned_at(self.last_cleaned_at, self.connection)
         for queue in self.queues:
             # If there are multiple workers running, we only want 1 worker
             # to run clean_registries().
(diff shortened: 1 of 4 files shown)


rq/worker_registration.py
```diff
@@ -9,13 +10,27 @@
...
+    """Returns the last time maintenance task was run."""
+    last_cleaned_at = connection.get(REDIS_WORKER_LAST_CLEANED_AT)
+    if last_cleaned_at:
+        return utcparse(as_text(last_cleaned_at))
+    return None
+
+
+def set_last_cleaned_at(last_cleaned_at: datetime, connection: 'Redis'):
+    """Sets the last time maintenance task was run."""
+    connection.set('last_cleaned_at', utcformat(last_cleaned_at))
```
selwin, line 31: Mind storing this as UTC timestamp instead of string?

rq/worker.py
```diff
@@ -401,12 +401,12 @@ def refresh(self):
...
-        if self.last_cleaned_at is None:
-            return True
-        if (utcnow() - self.last_cleaned_at) > timedelta(seconds=self.maintenance_interval):
-            return True
-        return False
+        """
+        Maintenance tasks should run on first startup or every maintenance
+        interval (defaults to 10 minutes).
+        """
+        l_c_at = self.last_cleaned_at = self.last_cleaned_at or worker_registration.get_last_cleaned_at(self.connection)
```
selwin, line 408: If we're storing the last cleaning time at `self.last_cleaned_at`, we don't need `l_c_at` variable.

selwin:
Actually I just realized something, the cleaning timestamp has to be done on a per queue basis and not RQ wide. It could be that we have different workers working on different queues.

rq/worker_registration.py
dams: How do I do that?

rq/worker.py
dams: l_c_at is there just for having a shorter line below

dams:
> Actually I just realized something, the cleaning timestamp has to be done on a per queue basis and not RQ wide. It could be that we have different workers working on different queues.

Unless I did it wrong, my changes store the cleaning timestamp **per worker** and the cleaning will be done on all the queues of this worker. It's not a RQ wide timestamps. So I think it'll work just fine

selwin:
> Unless I did it wrong, my changes store the cleaning timestamp per worker and the cleaning will be done on all the queues of this worker. It's not a RQ wide timestamps. So I think it'll work just fine

If I have 10 workers running `rq worker high` and 1 worker running `rq worker low`, all the workers will store their last maintenance timestamp at the same `rq:workers:last_cleaned_at` key. Chances are that the `low` queue will not be cleaned for a long time.

selwin:
> If I have 10 workers running rq worker high and 1 worker running rq worker low, all the workers will store their last maintenance timestamp at the same rq:workers:last_cleaned_at key. Chances are that the low queue will not be cleaned for a long time.

Please also write a test for this scenario.

rq/worker_registration.py
selwin: Use [current_timestamp()](https://github.com/rq/rq/blob/e2b600d9dfb3ed834cc766b368a0cb5d56ecfc9c/rq/utils.py#L221)

Issue #2022, opened 2024-01-10
Make registry cleaning lock expiry configurable


eswolinsky3241:
We run our workers on Kubernetes using a cluster autoscaler. This results in a decent number of abandoned jobs, and we use a custom exception handler to requeue jobs that get abandoned. Given the frequency of jobs being abandoned, we've set the `maintenance_interval` of our workers to one minute.
```
w = Worker(
    argv[1:],
    name=os.environ["HOSTNAME"],
    connection=client,
    exception_handlers=[abandoned_job_handler],
    maintenance_interval=60,
)
```
However on inspection, it looks like this interval is separate from how often the `clean_registries` method is called. That method is only run if a clean_registries lock is acquired, and that lock only expires every 15 minutes. Meaning our abandoned jobs are not being requeued as frequently as we'd like.

https://github.com/rq/rq/blob/a8209391fffe3bad195264e6ed22724e36ae1fb5/rq/queue.py#L251

Is it possible to make this a constructor argument on the Worker so that queues are cleaned more frequently?

selwin:
Yeah, so apparently there's a bug that the cleaning lock does not get deleted after maintenance is done. There's a PR addressing this issue, but it's not done yet.

If you want to submit a quick PR to delete the lock once maintenance tasks is done, it would be great.

eswolinsky3241:
@selwin PR submitted: https://github.com/rq/rq/pull/2024

PR #2024, opened 2024-01-18
Delete maintenance lock after registries cleaned

eswolinsky3241:


eswolinsky3241 (commented):
Addresses https://github.com/rq/rq/issues/2022

tests/test_worker.py
```diff
@@ -1063,11 +1063,11 @@ def test_clean_queue_registries(self):
...
         self.assertEqual(self.testconn.zcard(bar_registry.key), 0)
 
-        # worker.clean_registries() only runs once every 15 minutes
-        # If we add another key, calling clean_registries() should do nothing
+        # worker.clean_registries() releases the maintenance locks after cleaning registries.
+        # If we add another key and call clean_registries, the key should be deleted
         self.testconn.zadd(bar_registry.key, {'bar': 1})
         worker.clean_registries()
-        self.assertEqual(self.testconn.zcard(bar_registry.key), 1)
+        self.assertEqual(self.testconn.zcard(bar_registry.key), 0)
```
selwin, line 1070: Since the cleaning lock was added manually, shouldn't the manually added registry lock still be present? The worker shouldn't be able to acquire lock and thus, doesn't attempt to clean (and release the lock).

eswolinsky3241: @selwin it looks like the manually added lock's key is rq:wip:bar, not rq:clean_registries:bar.

eswolinsky3241: Also I apologize, is there a difference between the cleaning lock and the registry locK?

selwin: Good question, let me investigate further when I have the time.

selwin: So `rq:wip:bar` is the Redis key for `bar` queue's `StartedJobRegistry`. 

`self.assertEqual(self.testconn.zcard(bar_registry.key), 0)` means that the registry is empty, meaning `worker.clean_registries()` did indeed run.

Since the decision whether to clean registries is now handled by `should_run_maintenance_tasks`, I think we can delete the last test case.

Also, do you mind changing the assertions from `self.assertEqual(self.testconn.zcard(foo_registry.key), 0)` to `self.assertEqual(len(foo_registry), 0)`?

eswolinsky3241: @selwin Done

selwin:
Thanks!

PR #2027, opened 2024-01-22
Set workerpool with_scheduler defaults to True

chromium7:
#1874

### DIFF
--- a/tests/test_worker_pool.py
+++ b/tests/test_worker_pool.py
@@ -46,7 +46,7 @@ class TestWorkerPool(TestCase):
         sleep(0.5)
         _send_shutdown_command(worker_data.name, self.connection.connection_pool.connection_kwargs.copy(), delay=0)
         # 1 worker should be dead since we sent a shutdown command
-        sleep(0.2)
+        sleep(0.75)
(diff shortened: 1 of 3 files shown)


tests/test_worker_pool.py
```diff
@@ -43,9 +43,10 @@ def test_check_workers(self):
         self.assertEqual(len(pool.worker_dict.keys()), 2)
 
         worker_data = list(pool.worker_dict.values())[0]
+        sleep(0.5)
         _send_shutdown_command(worker_data.name, self.connection.connection_pool.connection_kwargs.copy(), delay=0)
         # 1 worker should be dead since we sent a shutdown command
-        sleep(0.2)
+        sleep(0.75)
```
chromium7, line 49: I increased the delay time because it takes a bit more time to start and stop the worker with a scheduler

ankush:
@chromium7 does this mean workerpool will always start with scheduler by default? :thinking: Based on diff it also doesn't seem like there's any way to disable this behaviour.

selwin:
@ankush in version 2.X, RQ's worker will also start with scheduler by default. I plan on introducing an option to disable the scheduler.

Issue #2038, opened 2024-02-09
Job.get_status() does not always return JobStatus Enum


indepndnt:
As it is, the return type for `get_status` is `Union[JobStatus, str, None]`. This results in an error if you expect an Enum and call `Job.get_status().value`.

I believe the return type annotations in line 325 and 332 should be `Optional[JobStatus]` and line 336 should be updated to `self._status = JobStatus(as_text(status)) if status else None`.

https://github.com/rq/rq/blob/2f5fecc1badb75b3e379b5d15c0b60b9c4931baa/rq/job.py#L325-L337

Also line 950 in `Job.restore` should probably be `self._status = JobStatus(as_text(obj.get('status'))) if obj.get('status') else None` for consistency and so that it doesn't set `Job._status` to a string.

https://github.com/rq/rq/blob/2f5fecc1badb75b3e379b5d15c0b60b9c4931baa/rq/job.py#L950

I also noticed that the docstring for the `as_text` function is inconsistent, it should probably say `value (str): The decoded string`.

https://github.com/rq/rq/blob/2f5fecc1badb75b3e379b5d15c0b60b9c4931baa/rq/utils.py#L41-L52

PR #2039, opened 2024-02-09
Update Job.get_status and Job.restore to consistently return JobStatus Enum

indepndnt:
I wrote up what I found in issue #2038, so I though I should create a PR with my suggested fixes.


rq/job.py
```diff
@@ -322,18 +322,18 @@ def get_position(self) -> Optional[int]:
             return q.get_job_position(self._id)
         return None
 
-    def get_status(self, refresh: bool = True) -> JobStatus:
+    def get_status(self, refresh: bool = True) -> Optional[JobStatus]:
```
selwin, line 325: I think `get_status()` should always returns something. If the job hash in Redis has no status, raise an exception.

indepndnt: Thank you for the note. I updated it to raise an exception if there's no status in Redis. I added a `status=JobStatus.QUEUED` to a few `Job.create` calls in other tests so that they wouldn't fail as a result. I don't think this changed the purpose or accuracy of those tests.

rq/job.py
```diff
@@ -327,12 +327,17 @@ def get_status(self, refresh: bool = True) -> JobStatus:
...
+            InvalidJobOperation: If refreshing and nothing is returned from the `HGET` operation.
+
         Returns:
             status (JobStatus): The Job Status
         """
         if refresh:
             status = self.connection.hget(self.key, 'status')
-            self._status = as_text(status) if status else None
+            if not status:
+                raise InvalidJobOperation(f"Failed to retrieve status for job: {self.id}")
```
selwin, line 339: Mind updating the test to test this code path so it can pass Codecov's test?

indepndnt: sure thing!

selwin:
Thanks!

PR #2041, opened 2024-02-14
Use latest interval for repeating

gnomeby:
Usually during repeating developers provide several intervals and each next is bigger than previous. 
It is more correct to use latest repeat interval than the first one in case when repeats are more than intervals.


gnomeby:
Closed to prevent backward incompatibility

Issue #2044, opened 2024-02-21
DeprecationWarning: datetime.datetime.utcnow()


superjcvd:
Hello

datetime.datetime.utcnow() will soon be deprecated and should be replaced by datetime.datetime.now(datetime.UTC)
(in utils.py lines 128 and 227)

I tried to create a PR but I'm not allowed to push my branch on this repo :)

Regards

dewalujjwal:
@superjcvd  this issue is resolved

riceyrice:
@dewalujjwal this fix was merged (to branch `master`) in March but is not in 1.16.2 which was released (from branch `v1`) in May? Is that intentional?

PR #2047, opened 2024-02-22
Replace deprecated utcnow() with timezone aware now()

Sybili:
Replace datetime.datetime.utcnow() which will be depreceated in the future with the timezone aware datetime.datetime.now(tz=datetime.timezone.utc).

This change is in regards to issue #https://github.com/rq/rq/issues/2044#issue-2146821884

### DIFF
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -125,7 +125,7 @@ def import_attribute(name: str) -> Callable[..., Any]:
 
 
 def utcnow():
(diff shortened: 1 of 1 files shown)


selwin:
Can you make the tests pass so I can merge this in?

selwin:
@Sybili reminder to update this when you have the time :)

selwin:
Already fixed in master.

PR #2055, opened 2024-03-10
 Replace deprecated utcnow() with timezone aware now() 

dewalujjwal:
Replace datetime.datetime.utcnow() which will be depreceated in the future with the timezone aware datetime.datetime.now(datetime.timezone.utc).

This change is in regards to issue https://github.com/rq/rq/issues/2044#issue-2146821884


dewalujjwal:
@selwin i have created the merge for the issue  review the code

selwin:
Thanks!

PR #2082, opened 2024-04-30
Fix fetch_many method type

hberntsen:
The type introduced in #1772 does not match the method description and behaviour. None is part of the returned list (line 628)


selwin:
Thanks!

Issue #2087, opened 2024-05-11
The @job decorator is not working as documented


naskio:
I am new to rq and I am to run the following code (extracted from the [docs#The @job decorator](https://python-rq.org/docs/#the-job-decorator))
```python
import redis
import rq
from rq.decorators import job
import time

r = redis.Redis(host='localhost', port=6379)
q = rq.Queue(connection=r)


@job('default', connection=r, timeout=5)
def add(x, y):
    return x + y


job = add.enqueue(3, 4)
time.sleep(1)
print(job.return_value())
```

Running this code is raising the following error:
```
AttributeError: 'function' object has no attribute 'enqueue'
```

rq version: 1.16.2
python version: 3.9

PR #2088, opened 2024-05-11
fix docs: use delay when using @job decorator

naskio:
fix for #2087


selwin:
I actually plan on deprecating `delay` method name, does `enqueue()` not work?

naskio:
Yes, it's not working on the mentioned version (delay is the one working now!)

selwin:
Thanks!

PR #2097, opened 2024-06-04
Make Retry type hint more exact

hongquan:
The current type hint of `Retry` makes MyPy check failed when user passes a tuple to `interval` parameter, though that usage is correct.


selwin:
Mind making the tests pass so I can merge this in? @hongquan

hongquan:
@selwin The failed test case is not related to this PR and it happens for older stack (Redis server v5 and redis-py v3.5). Could we just drop support for those stack?

PR #2104, opened 2024-07-05
Remove side effects in count and get_job_ids

mmacvicar:
Both count and get_job_ids methods in registries automatically call cleanup. It is an unexpected side effect that even triggers failure callbacks in the case of StartedJobRegistry. This have a couple of consequences:

- Logging or observability tooling end up having a real effect in the queue.
- Logging and observability scripts or tools using registries can fail due to undefined callbacks. This is the case with rq-exporter and rq-dashboard.
- Methods like a simple call to len(registry) or show_queues have obscure side effects. 

Based on the line blame I suspect there was a very good reason to put those cleanup calls there but I could not find any to keep them. @selwin may have more insights from 74a9982ecb4a01d8f60911b071c90cded806ce39 but this was 10 years ago. Registries are cleaned in the worker now as part of the maintainance tasks calling clean_registries, which makes me think this is no longer needed.

### DIFF
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -213,6 +213,7 @@ class BaseRegistry:
     def cleanup(self):
         pass
 
+
 class StartedJobRegistry(BaseRegistry):
     """
     Registry of currently executing jobs. Each queue maintains a
@@ -479,7 +480,6 @@ class CanceledJobRegistry(BaseRegistry):
 
     def get_expired_job_ids(self, timestamp: Optional[datetime] = None):
         raise NotImplementedError
-
 
 
 def clean_registries(queue: 'Queue', exception_handlers: list = None):
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -87,7 +87,6 @@ class BaseRegistry:
         Returns:
             int: _description_
         """
-        self.cleanup()
         return self.connection.zcard(self.key)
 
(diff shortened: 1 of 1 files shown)


selwin:
> Based on the line blame I suspect there was a very good reason to put those cleanup calls there but I could not find any to keep them. @selwin may have more insights from https://github.com/rq/rq/commit/74a9982ecb4a01d8f60911b071c90cded806ce39 but this was 10 years ago. Registries are cleaned in the worker now as part of the maintainance tasks calling clean_registries, which makes me think this is no longer needed.

When something goes wrong, it's possible that no workers are running. In this case, users may get inaccurate results if they didn't call `clean()` prior to getting the count of jobs in registries. By implicitly calling `cleanup()` before counting, we ensure that dashboards will always show the correct number when they are viewed.

In the case of dashboards and REPL, I think the performance vs accuracy tradeoff is worth it. But I do understand that the trade off is different in logging/observability tools which calls this method constantly.

Keeping these trade offs in mind, I think the better way forward is to:
1. Add a new `registry.get_job_count()` method that just returns the count of jobs without side effects.
2. Add a `cleanup=True` argument to `registry.get_job_ids()`

Logging/observability tools can use these if they simply want to get the number of jobs in registries without calling `clean()`.

selwin:
@anton-daneyko-ultramarin yes please! :)

anton-daneyko-ultramarin:
@selwin I've opened a PR, check out https://github.com/rq/rq/pull/2133

PR #2108, opened 2024-07-17
Register queue also for deferred job

sim6:
Without this change queues initialized with only deferred jobs are not
listed by `rq.Queue.all()`.


selwin:
Thanks!

sim6:
:)

PR #2111, opened 2024-08-02
Implement TTL for deferred jobs

hberntsen:
This builds upon #1685 and fixes the problem that jobs can stay forever in the deferred queue.


anton-daneyko-ultramarin:
@hberntsen Thanks for tackling this. This is an issue for me ATM.

Maybe you have an idea if it is possible to set the status of the child jobs to failed if the parent job failed?

rq/registry.py
```diff
@@ -398,11 +398,45 @@ class DeferredJobRegistry(BaseRegistry):
...
+
+        if job_ids:
+            failed_job_registry = FailedJobRegistry(self.name, self.connection, serializer=self.serializer)
+
+            with self.connection.pipeline() as pipeline:
+                for job_id in job_ids:
+                    try:
+                        job = self.job_class.fetch(job_id, connection=self.connection, serializer=self.serializer)
+                    except NoSuchJobError:
+                        continue
```
selwin, line 418: Mind adding a little test for this line so test coverage doesn't decrease?

rq/registry.py
```diff
@@ -398,11 +398,45 @@ class DeferredJobRegistry(BaseRegistry):
...
+
+                    job.set_status(JobStatus.FAILED)
```
selwin, line 420: Do you mine passing in pipeline in `set_status`?

hberntsen:
@anton-daneyko-ultramarin I'd need to look into that. It's probably something I'll also need myself

hberntsen:
@selwin Done! I've done some additional code cleanup as well

selwin:
Thanks!

Issue #2091, opened 2024-05-17
Use typing.Protocol to strictly enforce callback function signatures


Microwave-WYB:
The current implementation of callbacks is quite confusing. We must create a callback with the `Callback` class, while the class doesn't specify what keyword parameters should be included in a callback function. The only way to learn about the callback function protocol is to use the documentation.

To avoid using invalid callback functions in my project, I had to create protocols to enforce the format of the callback function.

This provides nice static-checking capability to my library. Invalid callback functions will be detected by the type checker and errors will be raised in the editor.

I would suggest implement something similar in the `rq` library.

```python
"""RQ task"""

from dataclasses import dataclass
from typing import Any, Callable, Protocol, runtime_checkable

import rq
from redis import Redis
from rq.job import Job


@runtime_checkable
class SuccessCallback(Protocol):
    """Protocol for success callbacks."""

    def __call__(self, job: Job, connection: Redis, *args: Any, **kwargs: Any) -> None: ...


@runtime_checkable
class FailureCallback(Protocol):
    """Protocol for failure callbacks."""

    def __call__(
        self,
        job: Job,
        connection: Redis,
        type: Any,  # pylint: disable=redefined-builtin
        value: Any,
        traceback: Any,
    ) -> None: ...


@runtime_checkable
class StoppedCallback(Protocol):
    """Protocol for stopped callbacks."""

    def __call__(self, job: Job, connection: Redis) -> None: ...


@dataclass
class Task:
    """Template class for a task to be submitted to an RQ queue."""

    queue: str | rq.Queue
    handler: Callable
    args: tuple | None = None
    kwargs: dict[str, Any] | None = None
    on_success: SuccessCallback | None = None
    on_failure: FailureCallback | None = None
    on_stopped: StoppedCallback | None = None

    def submit(self) -> Job:
        """Submit the task to the queue."""
        if isinstance(self.queue, str):
            self.queue = rq.Queue(self.queue)

        return self.queue.enqueue(
            self.handler,
            args=self.args,
            kwargs=self.kwargs,
            on_success=self.on_success,
            on_failure=self.on_failure,
            on_stopped=self.on_stopped,
        )
```

selwin:
Mind opening a PR for this? Since `Protocol` was introduced in Python 3.8, we'd need to drop  support for Python 3.7, which is also ok since it's already EOL.

Microwave-WYB:
[...]

PR #2113, opened 2024-08-09
Add callback protocols and runtime type checking for callback functions

Microwave-WYB:
This PR intends to implement the suggestion in #2091

I've made the following changes:

1. `success`, `failure`, and `stopped` callback types as a string enumeration class.
2. Protocol classes for each callback type.
3. `assert_type` method for the `Callback` class to assert the protocol of a callback function in runtime.
4. `InvalidCallbackException` exception class to raise when a callback function does not follow the protocol.

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -5,11 +5,24 @@ import logging
 import warnings
 import zlib
 from datetime import datetime, timedelta, timezone
-from enum import Enum
-from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Tuple, Type, Union
+from enum import Enum, StrEnum
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Optional,
+    Protocol,
+    Tuple,
+    Type,
+    Union,
+    runtime_checkable,
+)
 from uuid import uuid4
 
-from redis import WatchError
+from redis import Redis, WatchError
 
 from .defaults import CALLBACK_TIMEOUT, UNSERIALIZABLE_RETURN_VALUE_PAYLOAD
 from .timeouts import BaseDeathPenalty, JobTimeoutException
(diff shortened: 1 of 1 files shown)


Microwave-WYB:
I'm a bit confused how lint works. It seems like the code in the main branch also have lines longer than 80 in `job.py`

selwin:
Do you mind fixing the build errors? @Microwave-WYB

Microwave-WYB:
> Do you mind fixing the build errors? @Microwave-WYB 

I think the problem is the action runs python 3.7 tests first. As discussed in the issue, Protocol is introduced in 3.8.

Microwave-WYB:
> Do you mind fixing the build errors? @Microwave-WYB

I've fixed the python 3.8 compatibility issue using `typing.Type`. But all 3.7 tests will fail as discussed in #2091.

selwin:
Do you mind removing Python 3.7 from test matrix? Thanks!

Microwave-WYB:
I don't understand why the `Python3.8/Redis5/redis-py3.5.0` test failed :(

selwin:
@Microwave-WYB I simply reran the failed tests again :)

rq/job.py
```diff
@@ -108,7 +121,7 @@ def cancel_job(job_id: str, connection: 'Redis', serializer=None, enqueue_depend
     Job.fetch(job_id, connection=connection, serializer=serializer).cancel(enqueue_dependents=enqueue_dependents)
 
 
-def get_current_job(connection: Optional['Redis'] = None, job_class: Optional['Job'] = None) -> Optional['Job']:
+def get_current_job(connection: Optional["Redis"] = None, job_class: Optional["Job"] = None) -> Optional["Job"]:
```
selwin, line 124: One more thing, do you kind reverting the changes from single to double quotes? If you use black, please run it with —skip-string-normalization

Microwave-WYB: Oh didn't realize that. Sure we can recert the quotes

Microwave-WYB: Hmmm I may have messed up this. Let me try to fix it tomorrow to do things properly

Microwave-WYB:
Also we probably need tests for the new error

rq/job.py
Microwave-WYB: It seems like the original version mixes single and double quotes. What is a good way to properly revert only lines that changed the quotes? I've tried to add a temporary ruff linting config to fix the quotes, but some quotes in the file are double quotes.

selwin:
@Microwave-WYB it might be easier to open a new PR :)

rq/job.py
terencehonles: I did it for you by reverting everything and conditionally committing only the quotes revert, I also added a suggestion of using `typing_extensions` that I mentioned in the issue you initially created https://github.com/rq/rq/issues/2091

If you merge https://github.com/Microwave-WYB/rq/pull/1 it will apply those suggestions to this PR

rq/job.py
```diff
@@ -1679,8 +1695,77 @@ def __init__(self, max: int, interval: Union[int, Iterable[int]] = 0):
         self.intervals = intervals
 
 
+class CallbackType(str, Enum):
+    SUCCESS = 'success'
+    FAILURE = 'failure'
+    STOPPED = 'stopped'
+
+
+@runtime_checkable
```
terencehonles, line 1704: Should these really be runtime checkable? It seems like this should just be left for a type checking tool to check that it's the right type, if it's the wrong type at runtime you'll see an error when the job is invoked.

rq/job.py
```diff
@@ -1679,8 +1695,77 @@ def __init__(self, max: int, interval: Union[int, Iterable[int]] = 0):
...
+    CallbackType.FAILURE: FailureCallbackFunc,
+    CallbackType.STOPPED: StoppedCallbackFunc,
+}
+
+EXPECTED_SIGNATURE = {
+    CallbackType.SUCCESS: '(job: Job, connection: Redis, result: Any, *args: Any, **kwargs: Any) -> Any',
+    CallbackType.FAILURE: (
+        '(job: Job, connection: Redis, type: Type[BaseException], ' 'value: BaseException, traceback: Any) -> Any'
+    ),
+    CallbackType.STOPPED: '(job: Job, connection: Redis) -> Any',
```
terencehonles, line 1759: Instead of manually typing these and hoping they keep in sync you probably should use `inspect.signature`:

```suggestion
    callback_type: str(inspect.signature(protocol.__call__))
    for callback_type, protocol in CALLBACK_PROTOCOLS.items()
```

rq/job.py
```diff
@@ -170,7 +183,7 @@ def create(
         on_failure: Optional[Union['Callback', Callable[..., Any]]] = None,  # Callable is deprecated
         on_stopped: Optional[Union['Callback', Callable[..., Any]]] = None,  # Callable is deprecated
```
terencehonles, line 184: I can't comment on the first line, but you can change this to:

```suggestion
        on_failure: Optional[Union['Callback', FailureCallbackFunc]] = None,  # Callable is deprecated
        on_stopped: Optional[Union['Callback', StopCallbackFunc]] = None,  # Callable is deprecated
```

and you may want to make `Callback` generic so you can declare that as:

```suggestion
        on_failure: Optional[Union['Callback[FailureCallbackFunc]', FailureCallbackFunc]] = None,  # Callable is deprecated
        on_stopped: Optional[Union['Callback[StopCallbackFunc]', StopCallbackFunc]] = None,  # Callable is deprecated
```

rq/job.py
```diff
@@ -1679,8 +1695,77 @@ def __init__(self, max: int, interval: Union[int, Iterable[int]] = 0):
...
+    ),
+    CallbackType.STOPPED: '(job: Job, connection: Redis) -> Any',
+}
+
+
 class Callback:
-    def __init__(self, func: Union[str, Callable[..., Any]], timeout: Optional[Any] = None):
+    def __init__(
+        self,
+        func: Union[str, SuccessCallbackFunc, FailureCallbackFunc, StoppedCallbackFunc],
```
terencehonles, line 1766: if you make this generic this would look like:

```python
CallbackFn = TypeVar('CallbackFn', bound='Union[str, SuccessCallbackFunc, FailureCallbackFunc, StoppedCallbackFunc]')

class Callback(Generic[CallbackFn]):
    def __init__(self, func: CallbackFn, timeout: Optional[Any] = None) -> None:
```

and as mentioned above you could better type the `Job.create` for not only the legacy call-spec

terencehonles:
I was adding some type annotations for https://github.com/rq/django_rq and noticed that neither of these projects have their typing checked. I proposed https://github.com/rq/rq/pull/2123 and that probably should be merged in before this PR to make sure a type checker can reasonably use this type information.

selwin:
> I proposed https://github.com/rq/rq/pull/2123 and that probably should be merged in before this PR to make sure a type checker can reasonably use this type information.

@terencehonles I left some comments in the PR :)

Microwave-WYB:
I was not able to work on this during the last couple of weeks. The quotes linting is a mess. I guess I will just keep everything single quoted then. Hopefully I can fix everything by next week.

terencehonles:
> I was not able to work on this during the last couple of weeks. The quotes linting is a mess. I guess I will just keep everything single quoted then. Hopefully I can fix everything by next week.

@Microwave-WYB I fixed it for you in https://github.com/Microwave-WYB/rq/pull/1 if you merge that (it's targeted at your fork)

Microwave-WYB:
> > I was not able to work on this during the last couple of weeks. The quotes linting is a mess. I guess I will just keep everything single quoted then. Hopefully I can fix everything by next week.
> 
> @Microwave-WYB I fixed it for you in [Microwave-WYB#1](https://github.com/Microwave-WYB/rq/pull/1) if you merge that (it's targeted at your fork)

Oh sweet. Love it!

selwin:
@Microwave-WYB mind updating this PR?

Microwave-WYB:
Should we keep the `if TYPE_CHECKING` part?

terencehonles:
> Should we keep the `if TYPE_CHECKING` part?

That was just moved after all the imports, the conditional Protocol import was what was added. I'm not sure what conflicts need to be resolved but I'm seeing:

image

and it doesn't seem like the type checking tests have run :thinking:

Issue #2099, opened 2024-06-11
Discrepancies between documentation and code for `worker_ttl`


stv8:
Originally posted in the discord group here, but wanted to repost as an issue in case any one stumbles on this as well.

https://discord.com/channels/844816706231861248/844816707126296578/1248726228677034144

---

There appears to be a discrepancy between how the documentation describes `default_worker_ttl` and how it is actually used in code. The 

I stumbled on this discrepancy because I've been running into some difficult issues where jobs appear to be stuck, very similar to what is described in this open issue https://github.com/rq/rq/issues/758 and am hoping that adjusting the `socket_timeout` will fix the issue (ultimately it didn't).

The leading suggestion in the above GH issue is to configure the `socket_timeout` param. In the RQ docs it mentions that RQ automatically configures `socket_timeout` to be 10s more than the `default_worker_ttl`

https://python-rq.org/docs/connections/#timeout
> To avoid potential issues with hanging Redis commands, specifically the blocking BLPOP command, RQ automatically sets a socket_timeout value that is 10 seconds higher than the default_worker_ttl.
> 
> If you prefer to manually set the socket_timeout value, make sure that the value being set is higher than the default_worker_ttl (which is 420 by default).

However in the code it appears to be making the `socket_timeout` actually 5 seconds _shorter_ than the `default_worker_ttl`

`DEFAULT_WORKER_TTL = 420`

`_set_connection()` calls `self.connection_timeout`

https://github.com/rq/rq/blob/0e15b2a942b8d724e938b04c030161422942d1ea/rq/worker.py#L302

`timeout_config = {"socket_timeout": self.connection_timeout}`

https://github.com/rq/rq/blob/0e15b2a942b8d724e938b04c030161422942d1ea/rq/worker.py#L781-L782

which is 
```py
    def connection_timeout(self) -> int:
        return self.dequeue_timeout + 10
```
which calls 
```py
    def dequeue_timeout(self) -> int:
        return max(1, self.worker_ttl - 15)
```
Therefore, `connection_timeout = 420 - 15 + 10` => 415 seconds

While the documentation mentions it needs to be 10 seconds higher than the `default_worker_timeout`, the code ensures that it's 10 seconds higher than the **dequeue_timeout** which is an important distinction.

It appears that this code change is intentional according to this PR: https://github.com/rq/rq/pull/1794

[...]

PR #2114, opened 2024-08-09
TTL documentation clarifications

stv8:
Resolves: https://github.com/rq/rq/issues/2099

- clarify usage of the various timeouts and their values in the docs
- rename `default_worker_ttl` -> `worker_ttl` argument int he worker class
- sets a deprecation warning for the `default_workert_ttl` argument

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -1,5 +1,5 @@
 baseurl: /
-exclude: design
+exclude: [design]
 permalink: pretty
 
 navigation:
--- a/rq/cli/cli.py
+++ b/rq/cli/cli.py
@@ -267,7 +267,8 @@ def worker(
             queues,
             name=name,
             connection=cli_config.connection,
-            default_worker_ttl=worker_ttl,
+            default_worker_ttl=worker_ttl,  # TODO remove this arg in 2.0
+            worker_ttl=worker_ttl,
             default_result_ttl=results_ttl,
             maintenance_interval=maintenance_interval,
             job_monitoring_interval=job_monitoring_interval,
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -152,11 +152,7 @@ class BaseWorker:
         work_horse_killed_handler: Optional[Callable[[Job, int, int, 'struct_rusage'], None]] = None,
     ):  # noqa
 
-        warnings.warn(
-            "default_worker_ttl is deprecated. Use worker_ttl instead.",
-            DeprecationWarning,
-            stacklevel=2
-        )
+        warnings.warn("default_worker_ttl is deprecated. Use worker_ttl instead.", DeprecationWarning, stacklevel=2)
         self.default_result_ttl = default_result_ttl
         self.worker_ttl = worker_ttl
         self.job_monitoring_interval = job_monitoring_interval
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -139,7 +139,8 @@ class BaseWorker:
         connection: Optional['Redis'] = None,
         exc_handler=None,
         exception_handlers=None,
-        default_worker_ttl=DEFAULT_WORKER_TTL,
+        default_worker_ttl=DEFAULT_WORKER_TTL,  # TODO remove this arg in 2.0
+        worker_ttl=DEFAULT_WORKER_TTL,
         maintenance_interval: int = DEFAULT_MAINTENANCE_TASK_INTERVAL,
         job_class: Optional[Type['Job']] = None,
         queue_class: Optional[Type['Queue']] = None,
@@ -150,8 +151,14 @@ class BaseWorker:
         serializer=None,
         work_horse_killed_handler: Optional[Callable[[Job, int, int, 'struct_rusage'], None]] = None,
     ):  # noqa
+
+        warnings.warn(
+            "default_worker_ttl is deprecated. Use worker_ttl instead.",
+            DeprecationWarning,
+            stacklevel=2
+        )
         self.default_result_ttl = default_result_ttl
-        self.worker_ttl = default_worker_ttl
+        self.worker_ttl = worker_ttl
         self.job_monitoring_interval = job_monitoring_interval
         self.maintenance_interval = maintenance_interval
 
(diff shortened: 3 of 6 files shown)


docs/_config.yml
```diff
@@ -1,5 +1,5 @@
 baseurl: /
-exclude: design
+exclude: [design]
```
stv8, line 2: I had to make this change in order for `jekyll` to run correctly locally. I installed the latest Ruby (`3.3.4`) and latest Jekyll (`4.3.3`) on my machine and attempting to run Jekyll caused this error: 

```
Configuration file: /Users/sean/repos/rq/docs/_config.yml
jekyll 4.3.3 | Error:  'exclude' should be set as an array, but was: "design".
/Users/sean/.gem/ruby/3.3.4/gems/jekyll-4.3.3/lib/jekyll/configuration.rb:291:in `block in check_include_exclude': 'exclude' should be set as an array, but was: "design". (Jekyll::Errors::InvalidConfigurationError)
```

rq/worker.py
```diff
@@ -150,8 +151,14 @@ def __init__(
...
     ):  # noqa
+
+        warnings.warn(
+            "default_worker_ttl is deprecated. Use worker_ttl instead.",
+            DeprecationWarning,
+            stacklevel=2
+        )
         self.default_result_ttl = default_result_ttl
-        self.worker_ttl = default_worker_ttl
+        self.worker_ttl = worker_ttl
```
stv8, line 161: Let me know if this satisfies what you referred to in the issue

> Add a deprecation warning to default_worker_ttl and change it to worker_ttl

rq/cli/cli.py
```diff
@@ -267,7 +267,8 @@ def worker(
             queues,
             name=name,
             connection=cli_config.connection,
-            default_worker_ttl=worker_ttl,
+            default_worker_ttl=worker_ttl,  # TODO remove this arg in 2.0
+            worker_ttl=worker_ttl,
```
stv8, line 271: Not sure if adding a `TODO` is part of the code style of this, feel free to remove it. I just wanted to highlight the change.

rq/cli/cli.py
```diff
@@ -267,7 +267,8 @@ def worker(
             queues,
             name=name,
             connection=cli_config.connection,
-            default_worker_ttl=worker_ttl,
+            default_worker_ttl=worker_ttl,  # TODO remove this arg in 2.0
```
selwin, line 270: Mind raising a `DeprecationWarning` if `default_worker_ttl` is passed?

stv8: One problem is that the user can supply their own `worker_class`, so until there is a breaking 2.0 change which removes that arg I don't think we reliably tell if `default_worker_ttl` was passed in.

However If the built in worker class is used this is what the log looks like:

```
venv ❯ python -W default -m rq.cli worker default
/Users/sean/repos/rq/rq/cli/cli.py:266: DeprecationWarning: default_worker_ttl is deprecated. Use worker_ttl instead.
  worker = cli_config.worker_class(
10:05:59 Worker rq:worker:ebde86e466874bc0b60e6222354243ec started with PID 51139, version 1.16.2
```

Notice it is pointing to line 266 which is where we instantiate the class

`cli/cli.py:266: DeprecationWarning: default_worker_ttl is deprecated. Use worker_ttl instead.`

So we indicate to the user that this argument is being deprecated at the usage anyways. This is the reason why I added `stacklevel=2` in the deprecation warning as it logs where the usage occurs rather than the line where the deprecation warning is.

Let me know if that makes sense or if you would like some other adjustments.

rq/cli/cli.py
selwin: This may be removed in version 3.0. I plan on releasing 2.0 sometime in the next few weeks so this would not give people enough time to adapt their code.

rq/worker.py
```diff
@@ -139,7 +139,8 @@ def __init__(
         connection: Optional['Redis'] = None,
         exc_handler=None,
         exception_handlers=None,
-        default_worker_ttl=DEFAULT_WORKER_TTL,
+        default_worker_ttl=DEFAULT_WORKER_TTL,  # TODO remove this arg in 2.0
+        worker_ttl=DEFAULT_WORKER_TTL,
```
selwin, line 143: `worker_ttl` should be optional. If `worker_ttl` is `None` and `default_worker_ttl` is supplied, `default_worker_ttl` should still be used in line 157.

stv8: I'm quite confused about what you would like the end result to be here. 

The original ask in the linked ticket was:

> Add a deprecation warning to default_worker_ttl and change it to worker_ttl, I'd like this to be released as part of RQ v2.0

I changed `default_worker_ttl` -> `worker_ttl`  like you asked.

> worker_ttl should be optional.

It is currently optional since it has a default value of `DEFAULT_WORKER_TTL`

>  If worker_ttl is None and default_worker_ttl is supplied, default_worker_ttl should still be used in line 157.

Why have both `worker_ttl` and `default_worker_ttl` parameters? In the linked ticket it was suggested to rename it.

Also the renaming aligns it with the other parameters, i.e. they don't have `default` in the name.

For example:

`maintenance_interval: int = DEFAULT_MAINTENANCE_TASK_INTERVAL,` and `job_monitoring_interval=DEFAULT_JOB_MONITORING_INTERVAL,`

---
 
Let me know if I'm misunderstanding something but I'm having a hard time understanding what else there is to be done here.

Thanks

selwin: It should be something like this:
```python
default_worker_ttl=DEFAULT_WORKER_TTL,  # TODO remove this arg in 2.0
worker_ttl=None,
```

```python
if worker_ttl:
    worker_ttl = worker_ttl
elif default_worker_ttl:
    worker_ttl = default_worker_ttl
else:
    worker_ttl = DEFAULT_WORKER_TTL
```

I'd like it so that people using `Worker(default_worker_ttl=180)` to still work while it's being deprecated.

selwin:
Fixed in https://github.com/rq/rq/pull/2126

stv8:
@selwin Thanks for the guidance and for wrapping this up 😃 

Sorry for the delays on my end, we just had our second child recently so time for extra curricular coding has been sparse :)

Cheers

selwin:
No worries @stv8 . Congrats on the birth of your second child 😀

Issue #1641, opened 2022-04-02
Feature Request: Log job running time and worker id


veejar:
It's will be very useful to see in log job running time to monitor long jobs and improve them.
Also often multiple workers run for the same queue, so it will be useful to see worker ID for job in log.

selwin:
PR for this is welcome :)

PR #2118, opened 2024-08-15
Log job success

jackkinsella:
This closes https://github.com/rq/rq/issues/1641

## Motivation

When dealing with any queue system at scale, you need to identify slow jobs and understand issues related to potential concurrency problems between workers.

In addition, this logging gives system administrators an overview of what happened in the past when debugging difficult problems.

## Why INFO level logging?

Because logging _what_ jobs got completed _when_ is as fundamental to a job system as an access log is to a web server. It helps the programmers monitor _their_ code (whereas debug level in the `rq` library gives insight into the `rq` internals and would be too much info for a typical persisted log)

## Demonstration

Here's an example of the output

```
[INFO] Successfully completed snakes.tasks.populate_projects_iso_matches() job in 0:00:00.066600s on worker 11b1ade34709410484c854c1066708d2  [worker:1402]
```

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -1400,10 +1400,7 @@ class Worker(BaseWorker):
 
                     time_taken = job.ended_at - job.started_at
                     self.log.info(
-                        'Successfully completed %s job in %ss on worker %s',
-                        job.description,
-                        time_taken,
-                        self.name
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

PR #2120, opened 2024-08-26
Ensure the Redis socket timeout is long enough for blocking operations

terencehonles:
This change checks for Redis connection socket timeouts that are too short for operations such as BLPOP, and adjusts them to at least be the expected timeout.


selwin:
Thanks!

Issue #1296, opened 2020-07-08
Add "username" to redis.Redis initialization


KMouratidis:
Since Redis 6.0 we can use cool ACL stuff and `redis.Redis(username="kostas")` is already supported in the latest version or redis-py. 

An extra `'username': settings.get('REDIS_USERNAME', None),` [in here](https://github.com/rq/rq/blob/2775185374d829d8c946116d9d732a9ddb1b0611/rq/cli/helpers.py#L51) should do it. I'd write it myself and submit a PR but my soul belongs to EvilCorp :cry:

selwin:
RQ only takes a `redis` instance as `connection` argument. You should already be able to do it.

KMouratidis:
On the client side, yes that's easy, but what about the workers? I might be missing something but it seems you would need to sublass `redis.Redis` to get it's auth to work. Not hard, but thought it made more sense to be able to take that from the config file just like the other keyword arguments.

PR #2121, opened 2024-08-28
add: redis username to the settings

mertcanaliciaaron:
Hello,

I have encounter the problem that we don't have username parameter, only password exists. The issue briefly raised in #1296 but I guess not solved.

I have implemented the username parameter to the settings. And fixed some tests.

Workaround: REDIS_URL parameter works fine even though we don't have username parameter.


mertcanaliciaaron:
@selwin

selwin:
Could you make sure all tests pass so I can merge this in?

PR #2123, opened 2024-09-03
test type annotations with mypy

terencehonles:


rq/job.py
```diff
@@ -144,13 +150,66 @@ def requeue_job(job_id: str, connection: 'Redis', serializer=None) -> 'Job':
 class Job:
     """A Job is just a convenient datastructure to pass around job (meta) data."""
 
+    _dependency: Optional['Job']
     redis_job_namespace_prefix = 'rq:job:'
 
+    def __init__(self, id: Optional[str] = None, connection: Optional['Redis'] = None, serializer=None):
```
terencehonles, line 156: move `__init__` up so that the type information is used elsewhere in the class, alternatively the type information can be placed like on https://github.com/rq/rq/pull/2123/files#diff-5ff734c15373fbcdb39d32faf0555265d8eaae657c1a6174e5e9b0dedd2815f7R153

rq/local.py
```diff
@@ -14,13 +14,7 @@
 try:
     from greenlet import getcurrent as get_ident
 except ImportError:
-    try:
-        from threading import get_ident
-    except ImportError:
-        try:
-            from _thread import get_ident
-        except ImportError:
-            from dummy_thread import get_ident
```
terencehonles, line 23: This would only be needed for unsupported Python versions

rq/local.py
```diff
@@ -313,44 +307,125 @@ def __setitem__(self, key, value):
...
+        return len(self._get_current_object())
+
+    def __getitem__(self, i):
+        return self._get_current_object()[i]
+
+    def __iter__(self):
+        return iter(self._get_current_object())
+
+    def __contains__(self, obj):
+        obj in self._get_current_object()
```
terencehonles, line 353:
```suggestion
        return obj in self._get_current_object()
```

rq/worker.py
```diff
@@ -1499,6 +1499,8 @@ def handle_job_success(self, job: 'Job', queue: 'Queue', started_job_registry: S
 
                     pipeline.execute()
 
+                    assert job.started_at
+                    assert job.ended_at
```
selwin, line 1503: I don't personally like adding assertions just to satisfy type checkers because it's causing us to run extra code during run time. Can we just use type ignore?

selwin: If there's some version of `type ignore` that lets us bypass just the `None` type warning, it would be even better.

terencehonles: Assertions aren't exactly expensive so of any runtime code I think they're probably fine, but I understand your perspective. `# type: ignore[union-attr]` is a bit of a sledgehammer and may silence multiple errors where it's possible one of them is actually an error, conversely using `# type: ignore` does not update mypy's perspective of the code and you need to use `# type: ignore` on every access rather than having mypy narrow its understanding of what possible values the variable can be.

The "other version" is to use `cast` which is also runtime code (also a function call) and in order for mypy to narrow its understanding of what the variable is you'd need to reassign the value. This works for variables, but not for attributes like this (or at least I don't think it does). Regardless a reassign + cast is more expensive than an assertion.

I would argue the reason to use assertions would be to make sure that type definitions really do match the runtime data model. If you don't run anything at runtime then you just are relying on luck, and that you didn't lie to the typechecker. Adding assertions allows tests (runtime) to check if you're actually typing something too strictly and what you believed was an assertion is incorrect.

selwin:
Thanks!

PR #2129, opened 2024-10-06
Added `desc` argument to registry.get_job_ids()

selwin:


rq/registry.py
```diff
@@ -153,7 +153,7 @@ def get_expired_job_ids(self, timestamp: Optional[float] = None):
         expired_jobs = self.connection.zrangebyscore(self.key, 0, score)
         return [as_text(job_id) for job_id in expired_jobs]
 
-    def get_job_ids(self, start: int = 0, end: int = -1):
+    def get_job_ids(self, start: int = 0, end: int = -1, desc: bool = False):
         """Returns list of all job ids.
 
         Args:
```
anton-daneyko-ultramarin, line 159: Does it make sense to add new argument description here?

Issue #1836, opened 2023-02-24
Workers fail to accept job commands after brief redis connection loss


buroa:
I'm not sure why, but I think it has to do with the pubsub channel. The only way to "fix" the issue presently is to basically restart the worker.

lowercase00:
How brief is the connection loss?

buroa:
Roughly less than a minute & this is how I setup a redis connection on the worker:

```RETRY = Retry(ExponentialBackoff(10, 1), 12)
RETRY_ERRORS = [TimeoutError, ConnectionError, ConnectionRefusedError]

def get_redis_conn():
    redis_url = os.getenv("REDIS_URL", "redis://localhost:6379")
    return redis.from_url(
        redis_url,
        retry=RETRY,
        retry_on_error=RETRY_ERRORS,
        socket_timeout=(60 * 2) + 5,
        health_check_interval=15,
    )
```

lowercase00:
Ok, I was asking because for longer connection losses it seems there's a bug where the worker key is expired but the Worker keeps the heartbeats, so I was wondering if that was because of that. But considering the connection loss is brief, this is most likely related to the `pubsub` subscription.

Did you have the chance to check whether the worker still has the subscription to the channel (after it reestablishes the connection)?

buroa:
Looks like none are listening. I just replicated the issue in a local development environment and same thing.

You can replicate by:

1. create a rq worker
2. shut down redis for 15 seconds and then start it
3. create a long running job + attempt to stop it: it won't stop

```127.0.0.1:6379> pubsub channels
(empty array)
```

lowercase00:
Thanks for that @buroa, wonder if the `redis-py` exception `ConnectionLost` is silent because `pubsub` runs on a separate thread. Even in this case, quickly going through `redis-py` it seems there's a hook that should be reconnecting the channels in case of lost connections:

```python
...
    def on_connect(self, connection):
        "Re-subscribe to any channels and patterns previously subscribed to"
```

This is registered here:

```python
...
# register a callback that re-subscribes to any channels we
# were listening to when we were disconnected
self.connection.register_connect_callback(self.on_connect)
```

Anyways, will take a look at this, thanks for the report.

buroa:
Thanks @lowercase00, let me know if you need help. I'm debugging on my end too.

NefixEstrada:
Hello,

We've encountered the same bug. We've been investigating why it happens and why there's no reconnection, and I think we've found the reason:

[...]

Issue #2070, opened 2024-04-08
no way to catch the redis.exceptions.ConnectionError on pubsub.get_message


alor:
when running the worker (from cli or with a custom class) the `subscribe()` method does not pass any `exception_handler` to `self.pubsub.run_in_thread(sleep_time=0.2, daemon=True)` so if a disconnection from Redis happens while waiting for messages in the pubsub channel, this code (inside redis python driver) is triggered:

```
            try:
                pubsub.get_message(ignore_subscribe_messages=True, timeout=sleep_time)
            except BaseException as e:
                if self.exception_handler is None:
                    raise
                self.exception_handler(e, pubsub, self)
```

and since there is no `exception_handler` the exception is raised.

the issue is that this code runs in a separate thread `PubSubWorkerThread(threading.Thread)` so there is no way to `rescue` from it in the worker class.  

the program works correctly and reconnects after Redis connection is restored, but it's annoying to receive reports from `bugsnag` (setup as global catch all exceptions) for an unhandled exception that can occur sometimes but cannot be "silenced" since it's perfectly fine to get it.

selwin:
I think the solution would be to configure bugsnag to ignore this exception. In addition, we can add logging statements around this area so users are aware of this. Mind opening a PR?

selwin:
My suggestion was simply to add logging statements so we are aware that this is happening.

alor:
something like this, will be ok?  if this is what you mean, I'll create a PR with this changes

```
   def subscribe(self):
        """Subscribe to this worker's channel"""
        def handle_redis_exception(e, pubsub, thread):
            self.log.warn('Worker %s: Redis exception: %s', self.key, str(e))

        self.log.info('Subscribing to channel %s', self.pubsub_channel_name)
        self.pubsub = self.connection.pubsub()
        self.pubsub.subscribe(**{self.pubsub_channel_name: self.handle_payload})
        self.pubsub_thread = self.pubsub.run_in_thread(sleep_time=0.2, daemon=True, exception_handler=handle_redis_exception)
```

PR #2132, opened 2024-10-08
Add an exception handler for the pubsub thread

fcharlier:
Hi,

This is proposal to fix issues #1836 & #2070

After a connection error to Redis the pubsub thread was stopped and the worker could not receive commands anymore.

This commit adds an exception handler for the thread that adds a log message and ignores `redis.exceptions.ConnectionError`. Any other exception is re-raised.

redis-py internal mechanism allows the pubsub thread to recover its connection and reinstall the pubsub channel subscription to allow the worker to receive commands again after connection errors.

It tries to behave the same as the main worker loop retry mechanism but without the backoff wait factor because we're not operating in a loop.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -23,7 +23,7 @@ if TYPE_CHECKING:
     except ImportError:
         pass
     from redis import Redis
-    from redis.client import Pipeline, PubSub
+    from redis.client import Pipeline, PubSub, PubSubWorkerThread
 
 try:
     from signal import SIGKILL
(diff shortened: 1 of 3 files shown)


fcharlier:
N.B.: I have tested this change on rq==1.16.2 but this part of the code seems almost untouched in master.

rq/worker.py
```diff
@@ -950,12 +950,33 @@ def run_maintenance_tasks(self):
...
+        """
+        if isinstance(exc, (redis.exceptions.ConnectionError)):
+            self.log.error(
+                "Could not connect to Redis instance: %s Retrying in %d seconds...",
+                exc,
+                2,
+            )
+            print("Connection Error, waiting 2 seconds before retry")
+            time.sleep(2.0)
+        else:
```
selwin, line 969: For completeness' sake, is there an easy way to cover this line in test?

fcharlier: @selwin new commit should be fine.

selwin:
Thanks!

selwin:
@fcharlier will release 2.0 end of this month so stay tuned :)

PR #2133, opened 2024-10-10
Allow to get job count and job IDs without cleanup

anton-daneyko-ultramarin:
Prior to this commit it was not possible to get a number of jobs in a registry or list job IDs without triggering a cleanup. Calling cleanup caused issues for:
 * Monitoring tools (see https://github.com/rq/rq/pull/2104) and
 * Cases with high number of jobs in the registries (see https://github.com/rq/rq/pull/2003).
 * Cases where a failure callback is registered and but the `get_job_ids` is called not from the main thread (see https://github.com/rq/rq/pull/2104#issuecomment-2397114656).

In this commit:
 1. A new `BaseRegistry.get_job_count` method is added. It is a side effect free version of `BaseRegistry.count` and runs in O(1).
 2. A `cleanup` parameter added to `get_job_ids` that allows to avoid clean up if it is set to `False`.

Resolves https://github.com/rq/rq/pull/2104.
Resolves https://github.com/rq/rq/pull/2003.

### DIFF
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -83,12 +83,24 @@ class BaseRegistry:
 
     @property
     def count(self) -> int:
-        """Returns the number of jobs in this registry
+        """Returns the number of jobs in this registry after running cleanup
 
         Returns:
             int: _description_
         """
-        self.cleanup()
+        return self.get_job_count(cleanup=True)
+
+    def get_job_count(self, cleanup=True) -> int:
+        """Returns the number of jobs in this registry after optional cleanup.
+
+        Args:
+            cleanup (bool, optional): _description_. Defaults to True.
+
+        Returns:
+            int: _description_
+        """
+        if cleanup:
+            self.cleanup()
         return self.connection.zcard(self.key)
 
     def add(self, job: 'Job', ttl=0, pipeline: Optional['Pipeline'] = None, xx: bool = False) -> int:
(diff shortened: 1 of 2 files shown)


rq/registry.py
```diff
@@ -83,14 +83,24 @@ def __contains__(self, item: Union[str, 'Job']) -> bool:
...
-        """Returns the number of jobs in this registry
+        """Returns the number of jobs in this registry and runs cleanup
 
         Returns:
             int: _description_
         """
         self.cleanup()
         return self.connection.zcard(self.key)
 
+    def get_job_count(self) -> int:
```
selwin, line 94: Mind adding a `cleanup=True` argument here as well? As a rule of thumb, I want RQ to return the correct answer to everyone so it's beginner proof. But should you need more performance, it's available for you.

Another alternative is that we make this an internal API (`_get_job_count()`) and just point users to use `registry.count`

anton-daneyko-ultramarin: @selwin I've added the `cleanup=True` parameter; rebased on top of the current master, so you could merge cleanly; made `count` call `get_job_count`, to decrease the chance of desyncronization of the code branches.

Let me know if you need anything else.

anton-daneyko-ultramarin: @selwin Is there any plan with regards to when those changes might be released?

anton-daneyko-ultramarin: @selwin Disregard my question about the release--I have just saw your answer in another issue thread.

selwin:
Thanks!

PR #2136, opened 2024-11-01
use `Serializer` Protocol instead of concrete `DefaultSerializer`

terencehonles:
@selwin I happened to be looking at https://github.com/rq/django-rq/pull/683 and I noticed a few more uses of the concrete class that should be the protocol


selwin:
Thanks!

PR #2138, opened 2024-11-05
Adds validation to avoid ":" when user pass job_id

### LINKED ISSUE #2137: Dangling job occurs when job_id contains the character ":"
Hi,

After updating to v2.0.0, job_id cannot contain the character ":"; otherwise, the job is never picked up or started by a worker.

I'll investigate this over the next few days, but I suspect that [PR #1964](https://github.com/rq/rq/pull/1964) might restrict the job_id format. We may need to add validation when creating jobs to address this.

sanurielf:
Fixes #2137


selwin:
Thanks!

Issue #2119, opened 2024-08-22
`ended_at` not set for a job when using `is_async=False`


SpecLad:
```
import os

from rq import Queue
from redis import Redis

q = Queue(connection=Redis(), is_async=False)

job = q.enqueue(os.getcwd)

while job.get_status(refresh=False) != 'finished':
    job.refresh()

print(job.ended_at)
```

This prints `None`. If you remove `is_async=False` (and run a worker), it works correctly (i.e. prints a timestamp).

selwin:
PR welcome :)

PR #2142, opened 2024-11-13
Set ended_at for a job when using is_async=False

alexprabhat99:
Fixes: #2119


alexprabhat99:
@selwin Could you please review this PR? Your feedback would be greatly appreciated!

selwin:
Thanks!

PR #2146, opened 2024-11-15
Lift restriction on function parameter types

hongquan:
Many functions declare input parameter type as `List[T]` but only need to iterate over it. This PR changes the type of `Iterable[T]` or `Sequence[T]` to let user pass `tuple`, `deque`, generator.

`Iterable` is broader than `Sequence`, allowing user to pass generator. But a generator can only be iterated once (after that, it will exhaust), so if the function uses that argument twice, we limit the type to `Sequence`.

I also change some type hint from `Any` to generic type (using `TypeVar`) because `Any` makes type checker (MyPy, Pyright useless).

I observe that many function return data as a list, but that data then is not mutated. In that case, I think we better use `tuple` because it is has lower overhead than list and it is immutable, better match our use case.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -25,11 +25,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install black ruff
-
-    - name: Lint with black
-      run: |
-        black --check --skip-string-normalization --line-length 120 rq tests
+        pip install ruff
 
     - name: Lint with ruff
       run: |
--- a/Makefile
+++ b/Makefile
@@ -21,5 +21,4 @@ force_release: clean
 	twine upload dist/*
 
 lint:
-	@ black --check --skip-string-normalization --line-length 120 rq tests
 	@ ruff check --show-source rq tests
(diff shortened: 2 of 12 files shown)


selwin:
@hongquan can you make sure the type check test passes so I can merge this in?

hongquan:
After I merge new update from _master_ branch, the `test:typing` task is throwing new errors.

hongquan:
After replacing `Any` with more concrete types, more errors are exposed (which were previously hidden by `Any`) and the `test:typing` job failed.

hongquan:
The project black and Ruff config seems to be incompatible, every time I format with black, Ruff cries, and vice versa.


Btw, there are typing errors due to mismatch Redis return type. This is a known issue of `redis-py`.

selwin:
Mind removing black from pyproject.toml? We should use either one and at this point, I personally prefer ruff.

hongquan:
@selwin I have reformated the code with Ruff, removing `Black`, and all checking tasks have passed, except typing, which is expected because we let more typing errors expose. To fix those, we will need more changes in the code base.

selwin:
Can we get the Black/Ruff changes merged in first? I also don't mind if we merge in the typing changes slowly, over a few PRs if doing it in one fell swoop is too much work.

hongquan:
Will rework on this after https://github.com/rq/rq/pull/2152

PR #2150, opened 2024-11-16
Delete expired groups from registry

eswolinsky3241:
This was causing a bug when workers call `run_maintenance_tasks`, which involves cleaning up groups whose jobs have expired. 

If the group has already been deleted from Redis, due to all member jobs having been manually deleted with `job.delete()`, then the `Group.all()` method will raise a `NoSuchGroupError` as it iterates through the `rq:groups` set. 

This change will handle the `NoSuchGroupError` and remove the missing group from the registry.


selwin:
Thanks!

PR #2152, opened 2024-11-20
Replace Black with Ruff as formatter tool

hongquan:
Related: https://github.com/rq/rq/pull/2146#issuecomment-2487098849


selwin:
Thanks!

PR #2153, opened 2024-11-22
Lift restriction on function parameter types

hongquan:
Many functions declare input parameter type as `List[T]` but only need to iterate over it. This PR changes the type of `Iterable[T]` or `Sequence[T]` to let user pass `tuple`, `deque`, generator.

`Iterable` is broader than `Sequence`, allowing user to pass generator. But a generator can only be iterated once (after that, it will exhaust), so if the function uses that argument twice, we limit the type to `Sequence`.

I also change some type hint from `Any` to generic type (using `TypeVar`) because `Any` makes type checker (MyPy, Pyright useless).

(This is a rework of #2146).

We will need more PR to get rid `Any` type.

For discussion:
I observe that many function return data as a list, but that data then is not mutated. In that case, I think we better use `tuple` because it is has lower overhead than `list` and it is immutable, better match our use case.

### DIFF
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -11,8 +11,24 @@ import datetime as dt
 import importlib
 import logging
 import numbers
-from collections.abc import Iterable
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Type, Union
+
+# TODO: Change import path to "collections.abc" after we stop supporting Python 3.8
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Dict,
+    Generator,
+    Iterable,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    overload,
+)
 
 if TYPE_CHECKING:
     from redis import Redis
@@ -23,10 +39,16 @@ from redis.exceptions import ResponseError
 
 from .exceptions import TimeoutFormatError
 
+_T = TypeVar('_T')
+_O = TypeVar('_O', bound=object)
+_BT = TypeVar('_BT', bool, int, float, str, bytes)
+ObjOrStr = Union[_O, str]
+
+
 logger = logging.getLogger(__name__)
 
 
-def compact(lst: List[Any]) -> List[Any]:
+def compact(lst: Iterable[_T]) -> List[_T]:
     """Excludes `None` values from a list-like object.
 
     Args:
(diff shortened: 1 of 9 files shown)


selwin:
> Iterable is broader than Sequence, allowing user to pass generator. But a generator can only be iterated once (after that, it will exhaust), so if the function uses that argument twice, we limit the type to Sequence.

Agreed, I think using `Sequence` is better than `Iterable`.

> I also change some type hint from Any to generic type (using TypeVar) because Any makes type checker (MyPy, Pyright useless).

What would be the advantage of this?

rq/utils.py
```diff
@@ -147,7 +169,9 @@ def utcparse(string: str) -> dt.datetime:
             return datetime.datetime.strptime(string, '%Y-%m-%dT%H:%M:%SZ')
 
 
-def first(iterable: Iterable, default=None, key=None):
+def first(
```
selwin, line 172: I don't think this function is used, we can remove it.

hongquan:
@selwin 

> What would be the advantage of this?

Imagine that we have two types:

```py
class Dog:
   def bark(self): ....

class Cat:
    def meow(self):...
 ```

When MyPy encounter this:


```py
cat.bark()
```

and if `cat` is annotated as `Any`, MyPy will assume that the above code is always valid.
If we annotate concrete type, like: `cat: Cat`,  MyPy will know that the code above is wrong, because `Cat` doesn't have `bark` method. It will raise error, we will know what is wrong and fix.

selwin:
@hongquan I re-read my question and just realized that it was a stupid one. Thanks!

Can you remove `_BT` and `first()` in this PR? It seems like they're not used.

hongquan:
@selwin I deleted `first` and `_BT`.

hongquan:
@selwin After merging code from _master_ branch and run `pytest` on my PC:

- Testing on Py 3.8 hangs forever.
- Testing on Py 3.12 passed.

But the GH job failed at `test_pubsub_thread_survives_connection_error`.
Could you help?

selwin:
@hongquan are you able to successfully run all tests on the current master branch on your PC?

selwin:
FYI, [this PR](https://github.com/rq/rq/pull/2161) is based on the most recent commit from the master branch and all tests pass.

If you're unable to get the tests to pass, maybe try to split your PRs into a few smaller ones? We may be able to isolate the cause of the failing tests more easily this way.

hongquan:
Hi @selwin The test with master branch also fail on my PC. Maybe my local setup is wrong. I will try again.

hongquan:
Tested with Python 3.8, 3.10, 3.12, all Ok on my localhost:

image

Btw, how do you run test with different Python on your side? I asked because, to do so, I have to modify the `pyproject.toml`:

```toml
[[tool.hatch.envs.test.matrix]]
python = ["3.8", "py3.10", "3.12"]
```

hongquan:
@selwin All test cases have passed!

selwin:
Thanks!

PR #2154, opened 2024-11-22
Fix handling completed job with expired result when work horse dies

fancyweb:
Since https://github.com/rq/rq/pull/2039, `Job.get_status()` doesn't return `None` but raises `InvalidJobOperation` when `refresh=True`.

This change was not handled properly in `Worker.monitor_work_horse()`.

I stumbled upon this with a use-case where a job with `result_ttl=0` succeeds but the work horse doesn't exit with `0` (because it's wrapped). `Job.get_status()` ends up being called and raises an uncaught error.

There's no existing test for the behavior and I wasn't able to produce one.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -44,7 +44,7 @@ from .defaults import (
     DEFAULT_RESULT_TTL,
     DEFAULT_WORKER_TTL,
 )
-from .exceptions import DequeueTimeout, DeserializationError, ShutDownImminentException
+from .exceptions import DequeueTimeout, DeserializationError, InvalidJobOperation, ShutDownImminentException
 from .executions import Execution
 from .group import Group
 from .job import Job, JobStatus, Retry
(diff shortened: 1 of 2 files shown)


fancyweb:
Ref https://github.com/DataDog/dd-trace-py/pull/11512

Here's the scenario as I understood it:
1. A job with `result_ttl=0` is enqueued
2. This job is executed and completed
3. DataDog's `Worker.perform_job()` wrapper calls `get_result()`, raises the new error, and makes the horse process fails
4. rq calls `get_result()` again and raises the new error too :sweat_smile:

selwin:
Mind adding a test for this?

fancyweb:
@selwin After looking at it again, I was able to add a test demonstrating the issue.

selwin:
@fancyweb do you mind checking why the tests failed on Python 3.9?

selwin:
Fixed in https://github.com/rq/rq/pull/2168

fancyweb:
Thanks!

fancyweb:
Hello @selwin, could this be included in a new release plz?

PR #2157, opened 2024-11-30
Explicitly ask RQ to retry by returning a Retry object

selwin:


selwin:
Superseded by https://github.com/rq/rq/pull/2159

PR #2167, opened 2024-12-26
Add new queue to registry in enqueue_many

eswolinsky3241:
In https://github.com/rq/rq/commit/df5e99ba0b3274ab734cde612c270a788ee47ffc, the call to `sadd` new Queues to the `rq:queues` registry was moved from `queue._enqueue_job` to `queue.enqueue_job` due to `_enqueue_job` not being called on jobs that are deferred.

However, this broke the `queue.enqueue_many` function, which has its own handling for deferred jobs. Now, `enqueue_many` does not add the queue to `rq:queues`.

This change copies the `sadd` call to `enqueue_many` and tests that the registry contains the queue after the method is called.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -801,7 +801,7 @@ class Queue:
         pipe = pipeline if pipeline is not None else self.connection.pipeline()
 
         # Add Queue key set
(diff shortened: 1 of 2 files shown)


selwin:
Thanks!

Issue #1702, opened 2022-08-21
rq overrides default loglevel


nicbou:
I wanted to set the rq.worker log level to "WARNING" to remove noise from my logs. This was my logging configuration:

```
logging.config.dictConfig({
    'version': 1,
    'formatters': {
        'console': {
            "()": "coloredlogs.ColoredFormatter",
            'format': '%(asctime)s %(levelname)s [%(name)s:%(lineno)s] %(message)s',
        },
    },
    'handlers': {
        'console': {
            'class': 'logging.StreamHandler',
            'formatter': 'console',
            'stream': 'ext://sys.stdout',
        },
    },
    'loggers': {
        '': {
            'level': 'INFO',
            'handlers': ['console'],
        },
        'rq.worker': {
            'level': 'WARNING',
            'handlers': ['console'],
        }
    },
})
```

I kept seeing INFO messages from rq.worker. They had the correct colour and format, so I knew that the config worked properly, but the loglevel was ignored. [This person](https://stackoverflow.com/questions/60255188/python-rq-worker-not-honoring-djangos-logging-config) has the same problem.

Calling `worker.work` calls `setup_loghandlers`. If no `log_level` is passed, it defaults to`"INFO"` when it should probably be `None`.

https://github.com/rq/rq/blob/153d29cce74cf1143a5f6bbe2672e7fb55b9abfe/rq/worker.py#L566

In `setup_loghandlers`, this block overwrites the default log level and sets it to `"INFO"`.

https://github.com/rq/rq/blob/153d29cce74cf1143a5f6bbe2672e7fb55b9abfe/rq/logutils.py#L24

And thus the logging configuration is ignored, but only for the log level.

A workaround is to use `some_worker.work(logging_level=logging.WARNING)`, but rq should simply respect the logging configuration.

selwin:
Mind making a PR to fix this?

dmrzn:
Could I add a couple of notes on the issue?

I configured a logstash log handler for a root logger (using dictConfig) and found that 'rq.worker' logger stopped logging messages anywhere - no 'rq.worker' messages in logstash, no messages in std.out/std.err. 

This happens due to `if not _has_effective_handler(logger)` check in `logutils.py`. Once this check finds attached handlers to a parent logger - it doesn't add any handlers to 'rq.worker' logger. Therefore we have a logger without handlers.

Could you please advise, how do I direct all logs produced by RQ library to logstash? Thanks.

PR #2169, opened 2024-12-28
FIX bug worker override logging configuration

rlaminseok0824:
FIX Bug From #1702 
Only resolve bug that Worker override logging configuration.
Just set INFO if logging is logging.NOTSET. Otherwise,`None` is passed, makes not override default logging level.


selwin:
Thanks!

Issue #2170, opened 2024-12-28
[perf] possible to increase `sleep_time` for pubsub thread?


ankush:
To the best of my understanding 0.2s here directly translates to how fast you can join the thread after issuing a shutdown. It should not have any impact at runtime. This short timeout does result in high CPU usage when you have many workers (we often have 100s of workers in multitenant setup, so it can be effectively using 1 CPU at any time)

https://github.com/rq/rq/blob/9c5f6f3bf7990ad7900e1d5b8768b0290d866305/rq/worker.py#L1023-L1025

~~Is a PR welcome for increasing this to 2-5 seconds or making it configurable?~~ Sent a PR with a cleaner solution: https://github.com/rq/rq/pull/2172

PR #2172, opened 2024-12-30
perf: Avoid polling in PubSub thread

ankush:
My understanding of how this works could be wrong. So please take a look at it carefully.

Currently, each worker spawns a pub-sub thread to listen for shutdown or stop/kill job commands. This thread polls every 0.2 seconds.

The 0.2 seconds on its own doesn't do anything, Redis' socket connection can block on it forever if you just specify `sleep_time=None`, but now that thread will be permanently blocked until its blocking `recv` ends with a response from Redis.

There are three solutions here:
- We can increase timeout, this will increase the time it takes to shut things down.
- The thread doesn't "need" to be joined, it's a daemon thread and if it dies without joining with a worker it shouldn't be a problem.
- We can set blocking behavior AND unsubscribe pubsub before joining the thread. This way Redis will stop the blocking receive call and everything will cleanly exit.

I like the last approach as it's consistent with the current design.

This change:
- Makes pubsub calls completely blocking.
- `worker.unsubscribe` now unsubscribed before stopping pubsub thread. 
- Redis sends `unsubscribe` message back to pubsub thread, which wakes it up from long sleep and thread now exits cleanly. Here's strace output: `recvfrom(4, "*3\r\n$11\r\nunsubscribe\r\n$42\r\nrq:pu"..., 65536, 0, NULL, NULL) = 75`
- In the rare event of the thread not exiting cleanly, we exit anyway after waiting for 1 second. 

This should close https://github.com/rq/rq/issues/2170

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -1021,7 +1021,7 @@ class BaseWorker:
         self.pubsub = self.connection.pubsub()
         self.pubsub.subscribe(**{self.pubsub_channel_name: self.handle_payload})
         self.pubsub_thread = self.pubsub.run_in_thread(
-            sleep_time=0.2, daemon=True, exception_handler=self._pubsub_exception_handler
+            sleep_time=None, daemon=True, exception_handler=self._pubsub_exception_handler
         )
 
     def get_heartbeat_ttl(self, job: 'Job') -> int:
@@ -1054,9 +1054,9 @@ class BaseWorker:
         """Unsubscribe from pubsub channel"""
         if self.pubsub_thread:
             self.log.info('Unsubscribing from channel %s', self.pubsub_channel_name)
-            self.pubsub_thread.stop()
-            self.pubsub_thread.join()
             self.pubsub.unsubscribe()
+            self.pubsub_thread.stop()
+            self.pubsub_thread.join(timeout=1)
             self.pubsub.close()
 
     def dequeue_job_and_maintain_ttl(


selwin:
Thanks for the PR, give me some time to digest the proposed change.

> This thread polls every 0.2 seconds.

Just to be clear, the `polling` here - is this the Python process polling Redis every 0.2 seconds or is it simply within the Python process (our code checks whether there's a new message every 0.2 seconds)?

ankush:
> Just to be clear, the polling here - is this the Python process polling Redis every 0.2 seconds or is it simply within the Python process (our code checks whether there's a new message every 0.2 seconds)?


Ah yes, `polling` might be a poor choice of word here. It's `polling` the socket but not polling Redis. It's entirely within the Python process. 

This is internal to redis-py and to the best of my understanding, currently:

1. We call `parse_response` with `block=False` and `timeout=0.2` here: https://github.com/redis/redis-py/blob/8e2f2d3032e067123ed60d90d424a99910cf8891/redis/client.py#L894 
2. Since this times out every 0.2 second, this function is called repeatedly in a loop.
3. `self.connection` is long-lived connection and by default no health checks are performed by `redis-py`. So we don't wake up Redis to process anything, we just ask kernel if there's something new on socket.

Here's the relevant strace output:

```
poll([{fd=4, events=POLLIN}], 1, 200)   = 0 (Timeout)
ioctl(4, FIONBIO, [0])                  = 0
ioctl(4, FIONBIO, [1])                  = 0
poll([{fd=4, events=POLLIN}], 1, 200)   = 0 (Timeout)
ioctl(4, FIONBIO, [0])                  = 0
ioctl(4, FIONBIO, [1])                  = 0
poll([{fd=4, events=POLLIN}], 1, 200)   = 0 (Timeout)
ioctl(4, FIONBIO, [0])                  = 0
ioctl(4, FIONBIO, [1])                  = 0
```

It is still costly because the thread gets loaded on the CPU five times every second, especially for those running with few active workers and many idle workers at any given time.

selwin:
Thanks!

Issue #2058, opened 2024-03-15
RQ Tasks Failed Upon Importing Native MacOS modules


admirito:
On MacOS, If you attempt to import [pyarrow](https://pypi.org/project/pyarrow/) (or importing `pandas` when the `pyarrow` library is installed as well) [**edit:** or `requests` library or `urllib.request` from standard library], within the module where you implement your task as illustrated below:

```sh
pip install rq pyarrow
```

```py
# hello.py
import pyarrow
def hello_world():
    print("hello world!")
```

```py
# request_task.py
import rq
import redis
q = rq.Queue(connection=redis.Redis())
q.enqueue("hello.hello_world")
```

This will result in an error message produced in the `rq worker` output, which looks something like this:

```
18:35:15 default: hello.hello_world (c26c3735-eeb8-442f-a89e-75ecac6f1c92)
objc[34393]: +[NSString initialize] may have been in progress in another thread when fork() was called.
objc[34393]: +[NSString initialize] may have been in progress in another thread when fork() was called. We cannot safely call it or ignore it in the fork() child process. Crashing instead. Set a breakpoint on objc_initializeAfterForkError to debug.
18:35:15 Moving job to FailedJobRegistry (Work-horse terminated unexpectedly; waitpid returned 6 (signal 6); )
```

I tried the same thing on Ubuntu, and the error wasn't reproducible. Therefore, I assume it only happens within MacOS.

One workaround for this issue is to import pyarrow in your worker script prior to the worker forking its process:

```py
# myworker.py
import pyarrow
import redis
import rq
w = rq.Worker(['default'], connection=redis.Redis())
w.work()
```

Another workaround I found was to run the `rq worker` as follows:

```sh
OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES rq worker
```

**Edit:**

The issue turns out not to be specific to `pyarrow`. For instance, on MacOS if you import `urllib.request` from the standard library on top of the module that implements your RQ jobs, you will encounter the exact same problem.

After further investigation, it turns out that `urllib.request` internally imports `_scproxy`—a `darwin` platform-specific code. Therefore, the root cause of this issue is likely [the process forking issue in MacOS](https://www.wefearchange.org/2018/11/forkmacos.rst.html).

[...]

PR #2176, opened 2025-01-11
Added SpawnWorker

selwin:
Fixes https://github.com/rq/rq/issues/2058

### DIFF
--- a/docs/docs/workers.md
+++ b/docs/docs/workers.md
@@ -70,7 +70,6 @@ In addition to `--burst`, `rq worker` also accepts these arguments:
 * `--date-format`: Datetime format for the worker logs, defaults to `'%H:%M:%S'`
 * `--disable-job-desc-logging`: Turn off job description logging.
 * `--max-jobs`: Maximum number of jobs to execute.
-_New in version 1.8.0._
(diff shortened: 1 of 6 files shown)


docs/docs/workers.md
```diff
@@ -333,21 +332,112 @@ You can use the `-w` option to specify a different worker class to use:
...
+Or via CLI:
+```bash
+rq worker -w rq.worker.RandomWorker q1 q2 q3
+```
+
+### SpawnWorker
+_New in version 2.2.0._
+
+The `SpawnWorker` uses `os.spawn()` instead of `os.fork()` to run jobs. This is useful in environments
+where `fork()` is not available, like Windows or MacOS (`os.fork()` is discouraged).
```
iloveitaly, line 410: I _think_ [fork is being discouraged](https://docs.python.org/3/library/multiprocessing.html) everywhere and spawn will be the default multiprocessing method starting in 3.14:

> The default start method will change away from fork in Python 3.14. Code that requires fork should explicitly specify that via [get_context()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_context) or [set_start_method()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method).

Issue #2174, opened 2025-01-03
Sporadic exceptions caused by inconsistent handling of time zones


deathtracktor:
## Problem Statement  

When job objects are created using `Job.create`, their timestamp attributes, such as `created_at`, consistently include `tzinfo` information. However, when these same jobs are loaded from Redis, the `tzinfo` is lost because the `rq.utils.utcparse()` function returns offset-naive `datetime` instances.  
This discrepancy can lead to runtime exceptions in certain circumstances. For example:  

```
  File "/usr/local/lib/python3.11/site-packages/rq/worker.py", line 1399, in monitor_work_horse                    
    self.handle_job_failure(job, queue=queue, exc_string=exc_string)                                               
  File "/usr/local/lib/python3.11/site-packages/rq/worker.py", line 729, in handle_job_failure                     
    self.increment_total_working_time(job.ended_at - job.started_at, pipeline)                                     
                                      ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~                                                
TypeError: can't subtract offset-naive and offset-aware datetimes
```  

In this scenario:  
- The job is retrieved from Redis;  
- The `ended_at` attribute is set to `rq.utils.now()` (an offset-aware `datetime`);    
- Subtracting the parsed offset-naive `job.started_at` from `job.ended_at` causes a `TypeError`.  

_This issue is challenging to reproduce in a controlled test environment, and was encountered during "routine" debugging._  

## Proposed Solution  

Modify `rq.utils.utcparse` to always return offset-aware `datetime` instances. Additionally, ensure the relevant unit tests are updated to reflect this change.  
I have already implemented a fix for this issue and can submit a pull request (PR) if this proposal is accepted.

selwin:
Mind opening a PR for this? :)

mcmx:
I'm also seeing this error, in my case it is in a job with dependent jobs, it is created like this:

```python
job = get_current_job()
created_jobs = [...multiple jobs returned by queue.enqueue)
dependency = Dependency(created_jobs, allow_failure=True, enqueue_at_front=True
queue.enqueue(check_children_jobs, depends_on=dependency, meta={"parent_job": job.id})
```

rq: 2.1.0
python: 3.12.2
redis: (keydb actually) 6.3.4

rcstanciu:
Fixed by #2180

PR #2180, opened 2025-01-20
Force UTC time zone when parsing timestamps

deathtracktor:
## Problem description

See [original issue](https://github.com/rq/rq/issues/2174).

## Changes overview

* Updated `rq.utils.utcparse` to always return offset-aware `datetime` objects;
* Updated unit tests.

### DIFF
--- a/rq/cli/helpers.py
+++ b/rq/cli/helpers.py
@@ -22,7 +22,7 @@ from rq.defaults import (
     DEFAULT_WORKER_CLASS,
 )
 from rq.logutils import setup_loghandlers
-from rq.utils import import_attribute, parse_timeout
+from rq.utils import import_attribute, now, parse_timeout
 from rq.worker import WorkerStatus
 
 red = partial(click.style, fg='red')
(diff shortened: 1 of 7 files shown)


selwin:
Can you please make sure that all tests pass?

deathtracktor:
> Can you please make sure that all tests pass?

All tests are now passing. I'm not sure why `codecov` is flagging line 1532 - under the debugger at least one test hits this line, but in the report the entire block appears red.

selwin:
Thanks!

OlegZv:
@deathtracktor I think there is one more place where timezone-naive datetime object can creep in from: https://github.com/rq/rq/blob/d3d42833962db09d8e8a0c207599a64d4a550c64/rq/registry.py#L195
What do you think?

deathtracktor:
@OlegZv I could not make it fail in exact same way (with FakeRedis + redis-py) - `zscore` returns an integer, which should be perfectly convertible to timestamps.

The `utcfromtimestamp` method seems to be deprecated and triggers a warning instead:
```
DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC)
```

OlegZv:
It came up while I was implementing some tests and I noticed this test. I'll fix it my next PR, but was interesting how it propagates through the system.
https://github.com/rq/rq/blob/d3d42833962db09d8e8a0c207599a64d4a550c64/tests/test_registry.py#L74

lsmurray:
@selwin - can we get a release with this change

selwin:
Will make a release this weekend. I’m hoping to get https://github.com/rq/rq/pull/2192 in this release.

PR #2192, opened 2025-02-24
Hide job description on success

danilopeixoto:
Hide job description on success. This can be used to hide sensitive information.

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -1612,7 +1612,7 @@ class Worker(BaseWorker):
                         )
                     else:
                         self.log.info(
-                            'Successfully completed job in %ss on worker %s', time_taken, self.name
+                            'Successfully completed job %s in %ss on worker %s', job.id, time_taken, self.name
                         )
 
                     self.log.debug('Finished handling successful execution of job %s', job.id)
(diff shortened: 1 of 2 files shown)


rq/worker.py
```diff
@@ -1605,9 +1605,15 @@ def handle_job_success(self, job: 'Job', queue: 'Queue', started_job_registry: S
                     assert job.started_at
                     assert job.ended_at
                     time_taken = job.ended_at - job.started_at
-                    self.log.info(
-                        'Successfully completed %s job in %ss on worker %s', job.description, time_taken, self.name
```
selwin, line 1609: I think we should simply change this to `'Successfully completed job %s in %ss on worker %s', job.id, time_taken, self.name`. Showing job ID would make it easier for someone to trace a particular job's execution path.

danilopeixoto: Ok. We can apply that. Would you like to print yet the description if allowed by user?

selwin: For backwards compatibility's sake, yes.

danilopeixoto: Done!

danilopeixoto:
@selwin Can you check this? Also, I believe we need to introduce new test for it as coverage reports.

selwin:
Yes please, would be nice to have coverage on this case

danilopeixoto:
> Yes please, would be nice to have coverage on this case

@selwin Added tests.

PR #2194, opened 2025-02-26
Fix log with un-templated parameters

anna-intellegens:
Noticed a log message that was printing out raw `%s`.

Added in the (presumably) intended template arguments.


OlegZv (approved):
Looks good!

selwin:
Thanks!

PR #2198, opened 2025-03-07
Remove timezone naive coversion

OlegZv:
[Mentioned here](https://github.com/rq/rq/pull/2180#issuecomment-2696048019) there's one more timezone-naive timestamp conversion. This fixes it. 
Also updated the pre-commit versions of ruff, pyproject-fmt, and added simple pre-commit-hooks.

`pyproject.yaml` changes:
* Most formatting comes from the `pyproject-fmt`
* Added dev dependency group so `uv sync` installs everything for local development.


selwin:
Thanks!

PR #2199, opened 2025-03-07
Fix StartedJobRegistry job_id parsing

### LINKED ISSUE #2182: `StartedJobRegistry.get_job_ids()` id format
Will `StartedJobRegistry.get_job_ids()` id format comply with example `b0fc2487-425b-4a9e-9084-0d93d5a930d2:3d56d36a4cbc45b39b295bd4ccf108b1` instead `b0fc2487-425b-4a9e-9084-0d93d5a930d2`?

If yes, should this be documented?

OlegZv:
# Update the StartedJobRegistry to use correct job_id extracting (instead of job_id:execution_id, return job_id)

Changes/Updates:
* Added `parse_job_id` method. BaseRegistry would return the id as_text, but `StartedJobRegistry` would parse the job_id from `job_id:execution` format.
* StartedJobRegistry.cleanup() now properly saves the job result into the job result (using the _handle_failure)
* Added an option for the `Queue.enqueue_dependents` to prevent refreshing the job status, if the job status has been updated in the pipeline)
* Added deprecation notices to avoid adding/removing a job to the StartedJobRegistry (only Executions are allowed)
* Added a test where the job that is still executing would not be cleaned up, while in the started queue (test_no_cleanup_while_in_started_queue)
* Updated tests to exercise the StartedJobRegistry separately from the BaseRegistry.
* Added tests for a few warnings
* Update the `exc_info` in tests to avoid calling deprecated method (when available)

Issues it fixes:
* Fixes #2182 
* `get_expired_job_ids` would not return the job_ids, but rather `job_id:execution_id`. This means that the `StartedJobRegistry.cleanup()` would never actually clean anything, because the subsequent `Job.fetch` would not find the job with `job_id:execution_id`
* `IntermediateQueue.cleanup()` would never find the job in the `started_job_registry`. The `__contains__` method would only check if the job_id is in the Redis set and it would always return `False`.

### DIFF
--- a/rq/executions.py
+++ b/rq/executions.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, Dict, List, Optional
 from uuid import uuid4
@@ -118,7 +120,7 @@ class ExecutionRegistry(BaseRegistry):
         self.job_id = job_id
         self.key = self.key_template.format(job_id)
 
-    def cleanup(self, timestamp: Optional[float] = None):
+    def cleanup(self, timestamp: float | None = None, exception_handlers: list | None = None):
         """Remove expired jobs from registry.
 
         Removes jobs with an expiry time earlier than timestamp, specified as
--- a/rq/executions.py
+++ b/rq/executions.py
@@ -9,7 +9,7 @@ if TYPE_CHECKING:
 
 from .job import Job
 from .registry import BaseRegistry, StartedJobRegistry
-from .utils import as_text, current_timestamp, now
+from .utils import as_text, current_timestamp, now, parse_composite_key
 
 
 # TODO: add execution.worker
@@ -63,8 +63,8 @@ class Execution:
     @classmethod
     def from_composite_key(cls, composite_key: str, connection: Redis) -> 'Execution':
         """A combination of job_id and execution_id separated by a colon."""
-        job_id, id = composite_key.split(':')
-        return cls(id=id, job_id=job_id, connection=connection)
+        job_id, execution_id = parse_composite_key(composite_key)
+        return cls(id=execution_id, job_id=job_id, connection=connection)
 
     @classmethod
     def create(cls, job: Job, ttl: int, pipeline: 'Pipeline') -> 'Execution':
--- a/rq/job.py
+++ b/rq/job.py
@@ -77,10 +77,10 @@ class Dependency:
         Args:
             jobs (Iterable[Union[Job, str]]): An iterable of Job instances or Job IDs.
                 Anything different will raise a ValueError
-            allow_failure (bool, optional): Whether to allow for failure when running the depency,
+            allow_failure (bool, optional): Whether to allow for failure when running the dependency,
                 meaning, the dependencies should continue running even after one of them failed.
                 Defaults to False.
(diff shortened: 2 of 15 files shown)


rq/executions.py
```diff
@@ -118,7 +120,7 @@ def __init__(self, job_id: str, connection: Redis):
         self.job_id = job_id
         self.key = self.key_template.format(job_id)
 
-    def cleanup(self, timestamp: Optional[float] = None):
+    def cleanup(self, timestamp: float | None = None, exception_handlers: list | None = None):
```
selwin, line 123: `float | None` type hint only works on Python >= 3.10. Mind using `Optional[float]`?

OlegZv: I believe it works in previous python versions as well if you include `from __future__ import annotations` and the top of the file. This makes it easer to take advantage of a bit more modern type hinting. I think at the core is [the PEP 563.](https://peps.python.org/pep-0563/).

Similar to how we can take advantage of the `@depracted` decorator by importing it from `typing_extensions` (`from typing_extensions import deprecated`). Even though this decorator was added to `warnings` [in python 3.13.](https://peps.python.org/pep-0702/)

Also, I tried using `rq` locally with VSCode and python 3.8. The type hints seemed correct. Here's an example of trying to use an wrong input type for the `cleanup` function.
image

Let me know your thoughts! Would be nice to keep the newer type annotations moving forward.

selwin: I did a little bit of digging and it seems that `from __future__ import annotations` is strictly to backport `Postponed Evaluation of Annotations`.

It was typing-extensions that allowed `float | None` to work in Python < 3.10. I'm not keen on adding a dependency just to support syntactic sugar that's used internally. Mind removing this?

rq/job.py
```diff
@@ -825,11 +825,11 @@ def exc_info(self) -> Optional[str]:
         """
         Get the latest result and returns `exc_info` only if the latest result is a failure.
         """
-        warnings.warn('job.exc_info is deprecated, use job.latest_result() instead.', DeprecationWarning)
 
         from .results import Result
 
         if self.supports_redis_streams:
+            warnings.warn('job.exc_info is deprecated, use job.latest_result() instead.', DeprecationWarning)
```
selwin, line 832: Any reason why this warning was moved?

OlegZv: This is because users on older Redis versions cannot use `latest_result()` yet. So it would only warn users that do use Redis with streams.

OlegZv:
@selwin Created https://github.com/rq/rq/pull/2208 with just the changes related to the `StartedJobRegistry`.
It doesn't have any new type hints, no `@depracation` decorator, and no `exc_info` changes.

rq/executions.py
OlegZv: Thank you very much for your feedback and opinion. I respect your position on these items, so I'll avoid using newer typing features. I also think it's better called "developer experience" than "syntactic sugar."

* We use typing to help developers understand inputs/outputs. Things like `X | Y`, `list`, `tuple`, etc. are simpler than importing classes just for typing from the `typing_extension`. Also, 3.8 is EOL already, 3.9 to be EOL October this year.
* The same is true for the `@deprecation` decorator. `typing-extensions` is brings new features to older Python versions. I prefer my IDE to tell me right away, not to use a function, rather than wait until I run the code to discover a _log entry_ telling me something is deprecated.

P.S. I still agree that introducing all these changes in a single PR is an oversight on my side.

PR #2205, opened 2025-03-12
Fix example

isidroas:
Failing since 67624eba0b09 (#2043)
with the following error:

```python
  File "/tmp/rq/examples/run_example.py", line 6, in <module>
    from rq import Connection, Queue
ImportError: cannot import name 'Connection' from 'rq' (/tmp/rq/rq/__init__.py). Did you mean: 'connections'?
```

### DIFF
--- a/examples/run_example.py
+++ b/examples/run_example.py
@@ -2,17 +2,18 @@ import os
 import time
 
 from fib import slow_fib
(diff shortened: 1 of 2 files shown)


selwin:
Thanks!

PR #2208, opened 2025-03-15
Update the StartedJobRegistry to use correct job_id extracting

OlegZv:
* Added `parse_job_id` method. BaseRegistry would return the id as_text, but StartedJobRegistry would parse the job_id from `job_id:execution` format.
* StartedJobRegistry.cleanup() now properly saves the job result into job result (using the _handle_failure)
* Added an option for the `Queue.enqueue_dependents` to prevent refreshing the job status, if the job status has been updated in the pipeline)
* Added deprecation notices to avoid adding/removing a job to the StartedJobRegistry (only Executions are allowed)
* Added a test where the job that is still executing would not be cleaned up, while in the started queue (test_no_cleanup_while_in_started_queue)
* Updated tests to exercise the StartedJobRegistry separately from the BaseRegistry.
* Added tests for a few warnings

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -1679,6 +1679,17 @@ class Job:
 
         return all(status.decode() in allowed_statuses for status in dependencies_statuses if status)
 
+    def remove_executions(self, pipeline: Optional['Pipeline'] = None) -> None:
+        """Method to remove all job executions.
+
+        Args:
+            pipeline (Optional[Pipeline]): a pipeline or a connection to use. Defaults to None.
+        """
+        connection = pipeline if pipeline is not None else self.connection
+        execution_ids = [execution.composite_key for execution in self.get_executions()]
+        if execution_ids:
+            connection.zrem(self.started_job_registry.key, *execution_ids)
+
 
 _job_stack = LocalStack()
 
--- a/rq/job.py
+++ b/rq/job.py
@@ -1621,6 +1621,7 @@ class Job:
         parent_job: Optional['Job'] = None,
         pipeline: Optional['Pipeline'] = None,
         exclude_job_id: Optional[str] = None,
+        refresh_job_status: bool = True,
     ) -> bool:
         """Returns a boolean indicating if all of this job's dependencies are `FINISHED`
 
@@ -1634,7 +1635,8 @@ class Job:
         Args:
             parent_job (Optional[Job], optional): The parent Job. Defaults to None.
             pipeline (Optional[Pipeline], optional): The Redis' pipeline. Defaults to None.
-            exclude_job_id (Optional[str], optional): Whether to exclude the job id.. Defaults to None.
+            exclude_job_id (Optional[str], optional): Whether to exclude the job id. Defaults to None.
+            refresh_job_status (bool): whether to refresh job status when checking for dependencies. Defaults to True.
 
         Returns:
             are_met (bool): Whether the dependencies were met.
@@ -1656,9 +1658,9 @@ class Job:
             # If parent job is not finished, we should only continue
             # if this job allows parent job to fail
             dependencies_ids.discard(parent_job.id)
-            if parent_job.get_status() == JobStatus.CANCELED:
+            if parent_job.get_status(refresh=refresh_job_status) == JobStatus.CANCELED:
                 return False
-            elif parent_job._status == JobStatus.FAILED and not self.allow_dependency_failures:
+            elif parent_job.get_status(refresh=False) == JobStatus.FAILED and not self.allow_dependency_failures:
                 return False
 
             # If the only dependency is parent job, dependency has been met
--- a/rq/job.py
+++ b/rq/job.py
@@ -1250,12 +1250,10 @@ class Job:
             from .registry import StartedJobRegistry
 
             # TODO: need to cleanup job executions too
-
             registry = StartedJobRegistry(
                 self.origin, connection=self.connection, job_class=self.__class__, serializer=self.serializer
             )
             registry.remove_executions(self, pipeline=pipeline)
-            registry.remove(self, pipeline=pipeline)
 
         elif self.is_scheduled:
             from .registry import ScheduledJobRegistry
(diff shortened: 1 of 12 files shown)


rq/registry.py
```diff
@@ -286,59 +296,138 @@ def cleanup(self, timestamp: Optional[float] = None, exception_handlers: Optiona
...
+        score: Union[int, str] = ttl if ttl < 0 else current_timestamp() + ttl
         if score == -1:
             score = '+inf'
 
         return pipeline.zadd(self.key, {execution.composite_key: score}, xx=xx)  # type: ignore
 
-    def remove_execution(self, execution: 'Execution', job: 'Job', pipeline: 'Pipeline', delete_job: bool = False):
+    def remove_execution(
+        self,
+        execution: 'Union[Execution, str]',
```
selwin, line 334: Not sure why `remove_execution()` should also accept `str`. I think `remove_execution` is only called with `execution` as argument.

rq/registry.py
```diff
@@ -286,59 +296,138 @@ def cleanup(self, timestamp: Optional[float] = None, exception_handlers: Optiona
...
+        result = connection.zrem(self.key, to_delete)
         # if delete_job:
         #     job.delete()
         return result
 
-    # TODO: needs to add a method to cleanup executions
+    def get_job_and_execution_ids(
+        self, start: int = 0, end: int = -1, desc: bool = False, cleanup: bool = True
+    ) -> List[Tuple[str, str]]:
+        """Returns list of all job ids.
```
selwin, line 357: Can you clarify the doc string? This returns a list of tuple of job and execution ID.

In the future, I may add a dataclass to represent this structure too, to improve readability.

rq/registry.py
```diff
@@ -286,59 +296,138 @@ def cleanup(self, timestamp: Optional[float] = None, exception_handlers: Optiona
...
+    @staticmethod
+    def parse_job_id(entry: str) -> str:
+        # other classes might have a different entry format
+        # base registry stores just the job id
+        if not isinstance(entry, str):
+            entry = as_text(entry)  # type: ignore
+        job_id, _execution_id = parse_composite_key(entry)
+        return job_id
+
+    def remove_executions(self, job: Job, pipeline: Optional['Pipeline'] = None):
```
selwin, line 420: Do you mind moving this method to `job.remove_executions()`? It's cleaner this way.

rq/registry.py
OlegZv: Hmm... Now I'm trying to remember... I think originally I had some other code that just passed the `composite_key` instead of the execution object. This is because the function doesn't depend on the whole `Execution` object, but on the `composite_key`. **The Single Responsibility Principle**.
(Later I realized that this would break backwards compatibility, so I made it a `Union`. I think I'll return in back to the way it was)

rq/registry.py
OlegZv: Clarified, good catch. I like the dataclass idea!

rq/registry.py
OlegZv: I can totally do that. Would that be a breaking change though?
Maybe for now I can leave the registry method with a deprecation warning and inside just call `job.remove_executions()`?

OlegZv: I made the change, and also noticed `def cleanup_execution()` for the base worker uses the deprecated `.remove` for the started_job_registry. I updated it as well.

rq/job.py
```diff
@@ -1677,6 +1679,17 @@ def dependencies_are_met(
 
         return all(status.decode() in allowed_statuses for status in dependencies_statuses if status)
 
+    def remove_executions(self, pipeline: Optional['Pipeline'] = None) -> None:
```
selwin, line 1682: It seems like this method is not needed. Mind deleting this?

rq/registry.py
```diff
@@ -286,59 +296,136 @@ def cleanup(self, timestamp: Optional[float] = None, exception_handlers: Optiona
...
+        warnings.warn(
+            'StartedJobRegistry can only contain execution records. Use add_execution method.', DeprecationWarning
+        )
+        # in the future will raise
+        # raise NotImplementedError('')
+        job_id = job.id if isinstance(job, self.job_class) else job
+        # use job_id: so the stored record is always a tuple, when split by ':'
+        return super().add(f'{job_id}:', ttl, pipeline, xx)
+
+    def remove(self, job: Union[Job, str], pipeline: Optional['Pipeline'] = None, delete_job: bool = False):
```
selwin, line 396: I think it's safe to change `add()` and `remove()` methods in `StartedJobRegistry` to raise `NotImplementedError`. The fact that they're still present is a bug, they should not be used in RQ >= 2.

rq/registry.py
```diff
@@ -286,59 +296,136 @@ def cleanup(self, timestamp: Optional[float] = None, exception_handlers: Optiona
...
+    @staticmethod
+    def parse_job_id(entry: str) -> str:
+        # other classes might have a different entry format
+        # base registry stores just the job id
+        if not isinstance(entry, str):
+            entry = as_text(entry)  # type: ignore
+        job_id, _execution_id = parse_composite_key(entry)
+        return job_id
+
+    def remove_executions(self, job: Job, pipeline: Optional['Pipeline'] = None) -> None:
```
selwin, line 418: Can we delete this? Having a function that's calling another function is adding an unnecessary level of indirection.

tests/test_executions.py
```diff
@@ -9,13 +11,15 @@
 
 
 class TestRegistry(RQTestCase):
+    """Test the execution registry."""
+
     def setUp(self):
         super().setUp()
+        self.q = Queue(connection=self.connection)
```
selwin, line 18: Nitpick: mind changing `self.q` to `self.queue`? I want to keep things consistent whenever possible.

rq/job.py
selwin, line 1682: I'm really sorry about this, it seems like I was mistaken in reading the diffs from the previous PR. `started_job_registry.remove_executions()` was an existing method, right? In that case, there's no need to remove this method to job class, there's no point in moving things around without a good reason.

rq/registry.py
OlegZv: Roger that! will add NotImplemented.

rq/registry.py
OlegZv: Sure thing. I wanted to maintain backward compatibility **in case anyone is using this method**. Let's make sure we call this out in the release notes.

rq/job.py
OlegZv: I wish I saw this comment before removing this method :)
All good! I'll leave remove it from the `Job` class and keep it in the `StartedJobRegistry`.

rq/registry.py
```diff
@@ -103,32 +103,33 @@ def get_job_count(self, cleanup=True) -> int:
             self.cleanup()
         return self.connection.zcard(self.key)
 
-    def add(self, job: 'Job', ttl=0, pipeline: Optional['Pipeline'] = None, xx: bool = False) -> int:
+    def add(self, job: Union[Job, str], ttl: int = 0, pipeline: Optional['Pipeline'] = None, xx: bool = False) -> int:
```
selwin, line 106: Just want to make sure that adding `str` type is necessary.

Otherwise I think this PR looks good.

OlegZv: Updated!

selwin:
Thanks!

PR #2209, opened 2025-03-17
feat: deprecate 3.7, add 3.12

ccrvlh:


selwin:
Thanks for the PR. Any reason why support for Python 3.7 was removed? I'd rather not remove support for older versions of Python unless:
1. It complicates maintenance
2. There's a compelling new feature that we want to use in newer Python versions (like better support for type hints, but that's introduced only in Python 3.10).

taleinat:
Worth noting that the GitHub Actions "test" job in `workflow.yml` tests with Python versions 3.9-3.13, without going through tox.

That, and the files currently changed by this PR, should all be brought in line, testing with versions 3.7-3.13.

taleinat:
@ccrvlh do you intend to follow up on this PR?

ccrvlh:
Only objective was to align the lib with officially [supported Python versions](https://devguide.python.org/versions/) (it was even my mistake to leave 3.8 in there, was checking the matrix again and 3.8 is also EOL), no restrictions on supported tooling (eg. `Protocol`), and less overhead on maintenance and multiple environments. I'll close this for now and leave the decision to you guys on what to support.

PR #2210, opened 2025-03-17
feat: cron

ccrvlh:
This was originally implemented on `rq-scheduler`.
a couple of failing tests on my end, `test_cron_rescheduled_correctly` and `test_cron_schedules_correctly`, wip investigating them.

### DIFF
--- a/examples/run_worker.py
+++ b/examples/run_worker.py
@@ -1,4 +1,5 @@
 from redis import Redis
+
 from rq import Queue, Worker
 
 if __name__ == '__main__':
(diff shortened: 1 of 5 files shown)


selwin:
Hi, this is super exciting and definitely something that I want to get to eventually.

However, I'd like to take a more conservative approach and introduce things bit by bit to make sure the underlying implementation is solid.

Mind starting with the `repeat` feature first?
CleanShot 2025-03-19 at 08 55 18


I have this API in mind: `queue.enqueue(foo, repeat=Repeat(times=3, interval=10))`. It should also work for `queue.schedule()`

selwin:
@ccrvlh I merged in `Repeat` functionality into RQ in this PR: https://github.com/rq/rq/pull/2224

I'm also working on cron functionality, though I'm not sure when I'll be able to finish it.

PR #2212, opened 2025-03-17
fix example V2

isidroas:
- **Fix forgotten parenthesis**
- **Unnecesary run_worker.py**
- **use the same convetion as in documentation**
- **more precise name for `jobs`**
- **Precise with job.get_status()**
- **don't refresh to maintain consistency**
- **convert jobs from dict to list**
- **Split main() in 2 smaller functions**

Please see git messages for rationale.

The important commit is *Fix forgotten parenthesis*, the rest are a refactor.


selwin:
Thanks!

PR #2214, opened 2025-03-24
fix docs

isidroas:
- **Do not specify property,**
- **use the introduced method count()**
- **remove dead link**
- **enqueue_call() is never referenced in docs,**
- **real example of Retry()**


selwin:
Thanks!

Issue #2216, opened 2025-03-25
Docs: incorrect paragraph


isidroas:
In [docs/result](https://python-rq.org/docs/results/#:~:text=Additionally%2C%20you%20can%20use%20this%20for%20keeping%20around%20finished%20jobs%20without%20return%20values%2C%20which%20would%20be%20deleted%20immediately%20by%20default.) the following paragraph:

> Additionally, you can use this for keeping around finished jobs without return
> values (the function returned None), which would be deleted immediately by default.
>
> q.enqueue(func_without_rv, result_ttl=500) \# job kept explicitly

seems to be incorrect. According to my test, jobs returning None are not deleted immediately by default.

## Reproduction

I tested this behavior with the following patch:

``` diff
diff --git a/tests/test_worker.py b/tests/test_worker.py
@@ -1016,6 +1016,13 @@ class TestWorker(RQTestCase):
         w.work(burst=True)
         assert q.count == 0

+    def test_result_is_not_immediately_deleted_if_function_returns_None(self):
+        q = Queue(connection=self.connection)
+        job = q.enqueue(do_nothing)
+        w = Worker([q])
+        w.work(burst=True)
+        assert connection.exists(job.key)
+
     def test_worker_hash_(self):
         """Workers are hashed by their .name attribute"""
         q = Queue('foo', connection=self.connection)


```

The test passes, confirming that jobs returning None are retained.

## Additional Context

This documentation was introduced in commit [28bbeca66aa2dd24f0ff6270c086bad55efe0fde](https://github.com/rq/rq/commit/28bbeca66aa2dd24f0ff6270c086bad55efe0fde#diff-ee97874270c8de80984726b5504c9d2e05542bd7a728638f6651c1b457199f68R38-R41).

The test has been passing since v0.13 (older versions fail due to Python version incompatibility).

selwin:
Fixed in https://github.com/rq/rq/pull/2217

PR #2217, opened 2025-03-25
docs: remove paragraph

isidroas:
- **no return means function/task returns None**
- **Remove incorrect paragraph**

Related to https://github.com/rq/rq/issues/2216


selwin:
Thanks!

PR #2221, opened 2025-03-31
Fix duplicate multi() on the same pipeline

### LINKED ISSUE #2211: `redis.Pipeline` not working when enqueuing multiple jobs
Hi, 

I am trying to enqueue multiple jobs to several queues in a single `redis.Pipeline` using `enqueue_call` method. Please see the attached code snippet for details: 

```python

def _send_job(
    q_name: str,
    func: Callable,
    kwargs: dict = None,
    pipeline: redis.client.Pipeline = None,
):
    q = Queue(q_name, connection=rdb)
    job = q.enqueue_call(
        func=func,
        kwargs=kwargs,
        pipeline=pipeline,
    )

    return job

with rdb.pipeline() as pipe:
    for (q_name, kwarg) in SOME_LIST: 
        _send_job(q_name, SOME_FUNC, kwarg, pipe)
    pipe.execute(raise_on_error=True)
```

As you may noticed, I am not using `enqueue_many` method as I want to operate mutiple Queues in a pipe. The error raised is:

```
RedisError("Cannot issue nested calls to MULTI")
``` 

After checking the source code of RQ, the following code could be the cause: 

```python
# rq/queue.py:L594
    def setup_dependencies(self, job: 'Job', pipeline: Optional['Pipeline'] = None) -> 'Job':
        if len(job._dependency_ids) > 0:
            orig_status = job.get_status(refresh=False)
            pipe = pipeline if pipeline is not None else self.connection.pipeline()
            # ...omitted...
            pipe.multi()
            # ...omitted...
        elif pipeline is not None:
            pipeline.multi()  # Ensure pipeline in multi mode before returning to caller
        return job
```

I am wondering if it's a bug or my usage on RQ is not correct? Thanks in advance!

Nativu5:
This PR enhances the handling of Redis pipelines within the `setup_dependencies` method in `rq/queue.py`. 

Specifically, it addresses the scenario where jobs without dependencies are enqueued using a shared Redis pipeline across multiple queues. Previously, the pipeline's multi() method was called unconditionally, which prevented enqueuing multiple jobs across different queues using the same pipeline.

This PR will resolve #2211.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -660,7 +660,9 @@ class Queue:
                         # if pipeline comes from caller, re-raise to them
                         raise
         elif pipeline is not None:
-            pipeline.multi()  # Ensure pipeline in multi mode before returning to caller
+            # Ensure pipeline in multi mode before returning to caller (if not set before)
+            if not pipeline.explicit_transaction:
+                pipeline.multi()
         return job
 
     def enqueue_call(
--- a/tests/test_queue.py
+++ b/tests/test_queue.py
@@ -647,6 +647,29 @@ class TestQueue(RQTestCase):
             self.assertEqual(len(q), 3)
             self.assertEqual(q.job_ids, ['fake_job_id_3', 'fake_job_id_1', 'fake_job_id_2'])
 
+    def test_enqueue_different_queues_with_passed_pipeline(self):
+        """Jobs should be enqueued into different queues in a provided pipeline"""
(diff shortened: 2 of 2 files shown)


selwin:
@Nativu5 thanks for the PR. Mind adding a test case for this?

selwin:
Thanks!

PR #2232, opened 2025-04-15
Using a queue class from a classvar Worker in WorkerPool

amonsh1:
Hi. Sorry for my English. If I use a custom worker class and it has a custom queue class, i get the exception https://github.com/rq/rq/blob/master/rq/worker.py#L483. Is there a reason why this is impossible?


selwin:
Hi @amonsh1 , thanks for the PR. Do you mind adding a `queue_class` argument so it's consistent with `Worker` API? https://github.com/rq/rq/blob/94949a18272b4706f3ed3e9953a3613d49838243/rq/worker.py#L149

PR #2241, opened 2025-05-06
Remove job from dependency's `dependent_key` on cancel

Marishka17:
When a job is canceled, it should be removed from its dependencies `dependents_key`s. Otherwise, the code responsible for enqueuing parent job dependencies may fail.

This issue can occur in the following case:
There are two jobs: `X1` and `X2`, where `X2` depends on `X1`. If `X2` is recreated (canceled, then deleted, and a new job with the same ID is created), `X1` may fail when trying to enqueue its dependencies. This happens because `X2` no longer has any dependencies but still exists in `X1` `dependents_key`
```
  ...
  File "/opt/venv/lib/python3.10/site-packages/rq/job.py", line 1577, in dependencies_are_met
    connection.watch(*[self.key_for(dependency_id) for dependency_id in self._dependency_ids])
  File "/opt/venv/lib/python3.10/site-packages/redis/client.py", line 2138, in watch
    return self.execute_command("WATCH", *names)
  ...
redis.exceptions.ResponseError: wrong number of arguments for 'watch' command
```

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -1196,6 +1196,15 @@ class Job:
                     if pipeline is None:
                         pipe.watch(self.dependents_key)
                     q.enqueue_dependents(self, pipeline=pipeline, exclude_job_id=self.id)
+
+                # Go through all dependencies and remove the current job from each dependency's dependents_key
+                for dependency in self.fetch_dependencies(pipeline=pipe):
+                    if not dependency:
+                        continue
+
+                    dependents_key = dependency.dependents_key
+                    pipe.srem(dependents_key, self.id)
+
                 self._remove_from_registries(pipeline=pipe, remove_from_queue=True)
 
                 registry = CanceledJobRegistry(


Marishka17:
I'm not sure why the pytest tests are failing here, but I've compared local runs and the results are the same:
on master:
image
`RUN_SLOW_TESTS_TOO=1 hatch run test:cov --durations=5`
image

on my branch:
image
image
image
`RUN_SLOW_TESTS_TOO=1 hatch run test:cov --durations=5`
image

Marishka17:
@selwin, Hi, could you please take a look at the PR?

selwin:
Are you sure you’re running the tests on the latest commit in the master branch? The current master branch is passing all the tests

Marishka17:
> Are you sure you’re running the tests on the latest commit in the master branch? The current master branch is passing all the tests

Absolutely

It seems that [some](https://github.com/rq/rq/actions/runs/14881933585/job/41797146498) tests may be failing with the latest redis-py release due to recent [changes](https://github.com/redis/redis-py/pull/3614).

rq/worker.py
```diff
@@ -1728,6 +1728,7 @@ def fork_work_horse(self, job: 'Job', queue: 'Queue'):
         """Spawns a work horse to perform the actual work using os.spawn()."""
         os.environ['RQ_WORKER_ID'] = self.name
         os.environ['RQ_JOB_ID'] = job.id
+        connection_kwargs = {k: v for k, v in self.connection.connection_pool.connection_kwargs.items() if k != 'retry'}
```
selwin, line 1731: Do you mind removing this workaround and rebase this PR to follow master? I decided to disallow redis-py 6.0 in this PR: https://github.com/rq/rq/pull/2244

RQ should work with redis-py >= 6.0.1 since what caused this error has been fixed in redis-py's master branch.

selwin:
@Marishka17 mind resolving the conflicts in this branch? I plan to cut a new release soon.

selwin:
@Marishka17 after thinking about this further, I think we may have to add an option whether to remove job key from it's dependents on `job.cancel()`. When job B depends on job A, there could be situations when we don't want job B to run because it depends on something being done in job A.

selwin:
@Marishka17 https://github.com/rq/rq/pull/2255

PR #2249, opened 2025-05-21
Improve the typing of serializers

SpecLad:
There are two main changes here.

1. Even though several serializer variables are annotated as `Type[Serializer]`, as far as I can see, a serializer is not actually required to be a type. Change such annotations to just `Serializer` and update the `Serializer` protocol to a conventional protocol with methods.

2. Since RQ now requires Python 3.8, it can make use of `runtime_checkable`. Do that to simplify the serializer interface check.

In addition, add some more type annotations to serializer-type variables.

### DIFF
--- a/rq/serializers.py
+++ b/rq/serializers.py
@@ -1,21 +1,22 @@
 import json
 import pickle
 from functools import partial
-from typing import TYPE_CHECKING, Callable, Optional, Type, Union
+from typing import Any, Callable, ClassVar, Optional, Protocol, Union, runtime_checkable
 
 from .utils import import_attribute
 
-if TYPE_CHECKING:
-    from typing_extensions import Protocol
 
-    class Serializer(Protocol):
-        dumps: Callable[..., bytes]
-        loads: Callable[..., object]
+@runtime_checkable
+class Serializer(Protocol):
+    def dumps(self, obj: Any, /) -> bytes:
+        ...  # pragma: no cover
 
+    def loads(self, data: bytes, /) -> Any:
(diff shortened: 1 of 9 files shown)


selwin:
> I don't know, but the same tests are failing on the master branch, so I don't think it's anything I did.

Yeah, I'm trying to figure it out myself, I think something has changed with Github's Ubuntu images that cause tests to fail.

SpecLad:
Hmm... seems like this has somehow slowed down the testsuite, but only on Python 3.10. 🤔 I'll have to look into this.

rq/serializers.py
```diff
@@ -1,21 +1,22 @@
...
-if TYPE_CHECKING:
-    from typing_extensions import Protocol
 
-    class Serializer(Protocol):
-        dumps: Callable[..., bytes]
-        loads: Callable[..., object]
+@runtime_checkable
+class Serializer(Protocol):
+    def dumps(self, obj: Any, /) -> bytes:
+        ...
```
selwin, line 12: Mind adding ` # pragma: no cover` at the end of the `...` lines so Codecov doesn't mark is as untested?

selwin:
> Okay, I fixed two testsuite problems that could cause this issue, and it seems to have worked. 🤞 Let me know if you'd prefer me to submit these in a separate PR.

We can keep it in this PR :). Thanks!

rq/serializers.py
SpecLad: Sure; done.

selwin:
Thanks!

PR #2250, opened 2025-05-24
Update python versions set in the project

Jankovn:
- Removed depreciated python versions
- When latest depreciated was specified, the soon to be depreciated replaces it
- Add missing recent python versions in CI/CD and tests
- Fix mypy error when importing signal value by shadowing the previous import in try except condition

### DIFF
--- a/rq/worker.py
+++ b/rq/worker.py
@@ -1728,6 +1728,12 @@ class SpawnWorker(Worker):
         """Spawns a work horse to perform the actual work using os.spawn()."""
         os.environ['RQ_WORKER_ID'] = self.name
         os.environ['RQ_JOB_ID'] = job.id
+
+        redis_kwargs = self.connection.connection_pool.connection_kwargs
+        if redis_kwargs.get('retry'):
+            # Remove retry from connection kwargs to avoid issues with os.spawnv
+            del redis_kwargs['retry']
+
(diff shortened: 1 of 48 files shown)


rq/worker.py
```diff
@@ -1232,7 +1227,7 @@ def push_exc_handler(self, handler_func):
         """Pushes an exception handler onto the exc handler stack."""
         self._exc_handlers.append(handler_func)
 
-    def kill_horse(self, sig: signal.Signals = SIGKILL):
+    def kill_horse(self, sig: signal.Signals = signal.SIGTERM if not hasattr(signal, 'SIGKILL') else signal.SIGKILL):
```
selwin, line 1230: Can we please make this as a module level variable?

Jankovn: I've just set it. Should be good

rq/worker.py
```diff
@@ -102,6 +97,7 @@ def signal_name(signum):
     except ValueError:
         return 'SIG_UNKNOWN'
 
+SIGNAL_SHUTDOWN = signal.SIGTERM if not hasattr(signal, 'SIGKILL') else signal.SIGKILL
```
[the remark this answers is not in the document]
Jankovn: Renamed it

Issue #2269, opened 2025-06-20
IndexError during group cleanup


dixoncrews-gdl:
Running into an IndexError during a `Group.cleanup()` [call](https://github.com/rq/rq/blob/cb86d0a03d4f402c00bf0d5d94097462a1ae166b/rq/group.py#L48):

```
IndexError: list index out of range
Traceback (most recent call last):  File "/usr/local/lib/python3.12/site-packages/rq/worker.py", line 606, in work
    self.run_maintenance_tasks()
File "/usr/local/lib/python3.12/site-packages/rq/worker.py", line 987, in run_maintenance_tasks
    Group.clean_registries(connection=self.connection)
File "/usr/local/lib/python3.12/site-packages/rq/group.py", line 115, in clean_registries
  group.cleanup(pipeline=p)
File "/usr/local/lib/python3.12/site-packages/rq/group.py", line 48, in cleanup
  expired_job_ids.append(job_ids[i])
```

dixoncrews-gdl:
[Fix released in 2.4.1](https://github.com/rq/rq/releases/tag/v2.4.1)

PR #2270, opened 2025-06-20
Avoid IndexError during group cleanup

dixoncrews-gdl:
https://github.com/rq/rq/issues/2269

* Use a fresh pipeline for each call to `cleanup()`
   * Should avoid the case where `job_ids` and `results` have different lengths
* Return early if no jobs are found


selwin:
Thanks!

PR #2309, opened 2025-09-14
WorkerPool: single pool-level scheduler
  option + CLI flag

sanurielf:
Add with_scheduler to WorkerPool.start() (default: True) to manage a single pool-level RQScheduler.
  Spawn workers with with_scheduler=False when enabled. CLI: add --with-scheduler/--without-scheduler to rq worker-pool. Tests added:
  test_worker_pool_starts_single_scheduler_non_burst; test_worker_pool_scheduler_burst_no_process; test_worker_pool_without_scheduler.

### DIFF
--- a/rq/worker_pool.py
+++ b/rq/worker_pool.py
@@ -238,16 +238,18 @@ class WorkerPool:
 
     def stop_scheduler(self) -> None:
         """Ensure the scheduler process is stopped and joined."""
-        if not self.scheduler:
-            return
-        try:
-            if self.scheduler._process and self.scheduler._process.pid:
-                os.kill(self.scheduler._process.pid, signal.SIGTERM)
-        except OSError:
-            pass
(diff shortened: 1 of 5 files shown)


rq/worker_pool.py
```diff
@@ -177,26 +182,74 @@ def start_worker(
...
+            proc = scheduler._process
+            if proc is not None:
+                pid = getattr(proc, 'pid', None)
+                try:
+                    if pid:
+                        os.kill(pid, signal.SIGTERM)
+                except OSError:
+                    pass
+                finally:
+                    proc.join()
```
selwin: Thanks for the PR. In general, I think too much code is added for this simple feature. At the very least, I think we should think about unifying the `start_scheduler()` and `stop_scheduler()` code blocks with the same functionality in `Worker`.

PR #2319, opened 2025-10-10
Add typechecking for exception handlers

### LINKED ISSUE #2229: on_failure callback's traceback argument will sometimes be a StackSummary instance
Contrary to the documentation and code comments, the final argument `traceback` of an `on_failure` callback will not always be a traceback object. Rather, sometimes it will be an instance of `traceback.StackSummary`, which is the result of the `extract_stack()` method of traceback objects.

This is significant when writing failure callbacks since such objects need to be handled a bit differently.

If this behavior is intentional, then the documentation and comments should likely be corrected.

(The way I eventually figured this out was inspecting the values of that argument at runtime, and then reading through rq's code to figure out the source of the issue (the calls to `.extract_stack()`.))

hovsater:
This attempt to address the problem described in #2229. The type checker now fails at the problem call sites where `traceback.extract_trace()` is being used. I'm not entirely sure what the appropriate fix would be since we don't really have a real traceback. We could generate a synthetic one by doing something like:

```python
try:
  raise AbandonedJobErr()
except:
  exc_info = sys.exc_info()
```

And then make the approriate changes to

https://github.com/rq/rq/blob/55ec3ac34c19fb996e477048cba8b3d991e31e3d/rq/registry.py#L278-L280

and

https://github.com/rq/rq/blob/55ec3ac34c19fb996e477048cba8b3d991e31e3d/rq/registry.py#L284-L286

It's a hack for sure, but I'm not sure what the alternative would be either.

Fix #2229

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -1479,7 +1479,7 @@ class Job:
         with death_penalty_class(self.success_callback_timeout, JobTimeoutException, job_id=self.id):
             self.success_callback(self, self.connection, result)
 
-    def execute_failure_callback(self, death_penalty_class: type[BaseDeathPenalty], *exc_info):
+    def execute_failure_callback(self, death_penalty_class: type[BaseDeathPenalty], *exc_info: Unpack[ExcInfo]):
         """Executes failure_callback with possible timeout"""
         if not self.failure_callback:
             return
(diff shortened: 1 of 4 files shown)


hovsater:
Thoughts @selwin?

PR #2341, opened 2025-12-01
Modify ShutDownImminentException inheritance

JimNero009:
Change ShutDownImminentException to inherit from BaseException instead of Exception. Consider the difference:

```
try:
  foo = do_long_task()
except Exception:
  foo = "fallback"
```

If ShutDownImminentException is raised during `do_long_task()`, then code will continue to execute which is against the spirit of ShutDownImminentException. It is semantically similar to something like a KeyboardInterupt, and so should be modelled like one.

### DIFF
--- a/rq/exceptions.py
+++ b/rq/exceptions.py
@@ -26,7 +26,9 @@ class DequeueTimeout(Exception):
     pass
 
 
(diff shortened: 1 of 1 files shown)


selwin:
Thanks!

PR #2351, opened 2026-01-04
Implement job uniqueness check

selwin:


alex-deltia:
will the `unique` param be also supported by `rq.cron.register`?

PR #2374, opened 2026-03-07
Job IDs may only contain allowed characters

selwin:


alicederyn:
Why did this breaking change get landed? Was something not working correctly if you used "invalid" job IDs?

selwin:
Hi @alicederyn, I merged this in because I consider this a security fix. Previously `job.id` allowed quote/control characters that can break out of string literals and inject Python statements. This is problematic, if used with a `SpawnWorker`.

May I know which characters outside of the currently allowed ones that you use? I'll see if I can safely add those back in to try and minimize the backward incompatibility.

alicederyn:
I was using periods, `.`.

alicederyn:
Isn't `SpawnWorker` escaping job IDs with `repr`?

mhoeger:
Was also impacted by this breaking change and also using `.` 🙏

selwin:
Sorry about that, period is not something that can be safely allowed 🙏

alicederyn:
Apologies if this comes across as hostile, but I'm extremely alarmed that (a) you're writing code that's pasting unescaped strings into eval and (b) your fix is to adopt a strategy from 1990s PHP webservers instead of fixing the vulnerable code. Calling repl on a string is a trivial way to properly escape it for emplacing in source code if that's needed. Or, if this is a shell eval, there's https://docs.python.org/3/library/shlex.html#shlex.quote

selwin:
> Apologies if this comes across as hostile, but I'm extremely alarmed that (a) you're writing code that's pasting unescaped strings into eval and (b) your fix is to adopt a strategy from 1990s PHP webservers instead of fixing the vulnerable code. 

No worries :). I try to code securely, but obviously I make mistakes. If you spot any more vulnerabilities, feel free to let me know and I'll do my best to fix them.

selwin:
@alicederyn this PR should tighten things up further: https://github.com/rq/rq/pull/2398

This PR prevents users from injecting malicious job IDs to Redis via RQ. The [other PR](https://github.com/rq/rq/pull/2398) prevents `SpawnWorker` from executing malicious job ID stored in a compromised Redis server.

OlegZv:
Hello @selwin! This is a second time where the job ID changes break our application and require a lot of cleanup/rewrite to support newer RQ versions (first one was colons `:` in https://github.com/rq/rq/pull/1964). I believe both could've been done in a backwards compatible manner to allow old behavior and/or smoother transition between versions.

For this specifically, since it's a security concern - it's important to get things right. We, same as others mentioned, are using `.` as a delimiter (after `:` was restricted). Would you be open to a PR that allows a `.` but maybe shows a security warning? I'm not sure how a `.` can make the SpawnWorker vulnerable, but this way it at least provides a path forward to applications to keep up with the great new features and updates of the `rq`.

Thanks again for your work on the package!

UPD: alternatively could make this configurable, like `allowed_job_id_regex` as a class property, settable via `@classmethod` on the `Job` class.

PR #2395, opened 2026-04-29
fix race condition preventing restarting worker

terencehonles:
This change fixes a race condition, where a worker's key hits its expiration and is removed but a call to ``hset`` recreates the hash resulting in a record that is not considered alive nor dead, by checking for an expired key by the existence of the hash property "birth" no longer being defined.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stale or incomplete worker entries are now fully removed, preventing phantom workers.
  * Heartbeat handling is more robust: timestamps update reliably, job keys created during heartbeats are cleaned up, and worker metadata is refreshed if it expired mid-heartbeat.

* **Refactor**
  * Worker state construction and registration/heartbeat flows centralized for clearer logic and atomic updates when refreshing worker info.

### DIFF
--- a/rq/worker/base.py
+++ b/rq/worker/base.py
@@ -292,7 +292,12 @@ class BaseWorker:
         if not worker_key.startswith(prefix):
             raise ValueError(f'Not a valid RQ worker key: {worker_key}')
 
-        if not connection.exists(worker_key):
+        if (
+            not (key_exists := connection.exists(worker_key))
+            or not connection.hexists(worker_key, 'birth')
+        ):
+            if key_exists:
+                connection.delete(worker_key)
             connection.srem(cls.redis_workers_keys, worker_key)
             return None
 
@@ -875,7 +880,11 @@ class BaseWorker:
     def register_birth(self):
         """Registers its own birth."""
         self.log.debug('Worker %s: registering birth', self.name)
-        if self.connection.exists(self.key) and not self.connection.hexists(self.key, 'death'):
+        if (
+            self.connection.exists(self.key)
+            and self.connection.hexists(self.key, 'birth')
+            and not self.connection.hexists(self.key, 'death')
+        ):
             msg = 'There exists an active worker named {0!r} already'
             raise ValueError(msg.format(self.name))
         key = self.key
--- a/rq/worker/base.py
+++ b/rq/worker/base.py
@@ -216,7 +216,8 @@ class BaseWorker:
         self.failed_job_count: int = 0
         self.total_working_time: float = 0
         self.current_job_working_time: float = 0
-        self.birth_date = None
+        self.birth_date: datetime | None = None
+        self.last_heartbeat: datetime | None = None
         self.scheduler: RQScheduler | None = None
         self.pubsub: PubSub | None = None
         self.pubsub_thread = None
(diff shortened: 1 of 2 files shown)


terencehonles:
The record I was seeing was a result of this call https://github.com/rq/rq/blob/b418d8d053447e0b8256fa1e0050be35f174334e/rq/worker/base.py#L762-L771

e.g.:
`{'current_job_working_time': <some value>}`

I have updated the `find_by_key` method to also maintain (delete) the worker's hash in addition to the worker set if the worker is no longer considered alive, but there is no other code that looks at the worker's record, right? (It should also defensively look for an "alive" marker)

terencehonles:
> ## [Codecov](https://app.codecov.io/gh/rq/rq/pull/2395?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=rq) Report
> ✅ All modified and coverable lines are covered by tests. ✅ Project coverage is 93.59%. Comparing base ([`2de9491`](https://app.codecov.io/gh/rq/rq/commit/2de949126fb0213c3654051d49fafc84e1e54193?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=rq)) to head ([`8f0a1f1`](https://app.codecov.io/gh/rq/rq/commit/8f0a1f1b8a37dc15b6b2c643a563179833ed1849?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=rq)). ⚠️ Report is 254 commits behind head on master.
> 
> Additional details and impacted files
> ```diff
> @@            Coverage Diff             @@
> ##           master    #2395      +/-   ##
> ==========================================
> - Coverage   93.61%   93.59%   -0.03%     
> ==========================================
>   Files          28       38      +10     
>   Lines        3760     4871    +1111     
> ==========================================
> + Hits         3520     4559    +1039     
> - Misses        240      312      +72     
> ```
> 
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/rq/rq/pull/2395?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=rq). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=rq).

I'm not sure why codecov is using such an old commit as the base (it's not), but I just updated my fork's `master` to match this repo's master.

selwin:
But the fact that `set_current_job_working_time` is called, this means the worker is actually alive, right? I think in this case we should actually recreate the entire worker hash.

terencehonles:
> But the fact that `set_current_job_working_time` is called, this means the worker is actually alive, right? I think in this case we should actually recreate the entire worker hash.

Yes, I guess that's an additional facet that probably should be accounted for. At the time that `set_current_job_working_time` was called the worker was alive, but it _did_ die. I was noticing that one worker was not restarting on a machine and looked at the logs and it was convinced it was still alive (I knew it wasn't: the machine was restarted), and I noticed that there was the broken hash which never would expire since calling `hset` creates a hash if it doesn't exist, and since there was no expiration set it would stick around forever unlike the one created/updated during creation/heartbeart/death.

terencehonles:
I started looking at how to recreate the hash and I realized this bug is caused by the "horse" monitor running slower than the worker's TTL. The monitor will see that the process is still running and set the job's current working time and handle the heartbeats:

https://github.com/rq/rq/blob/e7e14a7b73158189f6b4f817bf799fa6eb4f6267/rq/worker/worker_classes.py#L89-L101

I would have thought that one of the calls to `heartbeat` would refresh the expiration, so I'm not sure what's going on there... Maybe that's when things started to crash on my side and why the worker was in a broken state. It's also possible, since that doesn't use a transaction/pipeline, the worker was cut off from Redis at that point and it only set the job execution time, but couldn't get to updating the heartbeat (I was also running into a networking error due to an OOM/memory pressure breaking systemd-networkd's DHCP renewal).

Should the worker's TTL not be the job's timeout if it's set? Additionally, it also seems like the worker TTL in `maintain_heartbeats` should be longer than `self.job_monitoring_interval` and possibly `self.job_monitoring_interval + self.worker_ttl` (and maybe still some buffer still). As written `register_birth` sets a default 7min TTL, but then `maintain_heartbeats` will keep it at least 1.5 min, but that's actually dropping it over 5 minutes when the worker starts (the `LT` flag is not set for the expiration).

I understand my system was degraded at the time, and possibly completely broken (if it was indeed cut off from the network in between the calls and that would have to be pretty unlucky), but I don't think the system should remain in a stuck state and require manual **Redis** intervention to remove the offending key.

selwin:
> I would have thought that one of the calls to heartbeat would refresh the expiration, so I'm not sure what's going on there... 

Redis/network latency issues could cause keys to expire before heartbeats arrive, although we already try to pad TTLs by 60 seconds. Looking at the [heartbeat code](https://github.com/rq/rq/blob/e7e14a7b73158189f6b4f817bf799fa6eb4f6267/rq/worker/base.py#L1162), I'd guess that the key expires before the TTL gets refreshed.

CleanShot 2026-05-04 at 20 38 42

I see two options:
1. Simply swap the order of the two commands (`expire` and `hset`) and run the heartbeat command in a pipeline [here](https://github.com/rq/rq/blob/e7e14a7b73158189f6b4f817bf799fa6eb4f6267/rq/worker/worker_classes.py#L96). In this case, even if the key is recreated, it will still eventually die if the worker is truly dead because the `expire` command will set a TTL.
2. `hset` command returns the number of new fields created. If it returns `1`, this means the worker key has already expired.

In both cases, we can try to repopulate the hash with whatever worker information that we have.

rq/worker/base.py
```diff
@@ -872,35 +878,37 @@ def _start_scheduler(
...
+
     def register_birth(self):
         """Registers its own birth."""
         self.log.debug('Worker %s: registering birth', self.name)
-        if self.connection.exists(self.key) and not self.connection.hexists(self.key, 'death'):
+        if (
+            self.connection.exists(key := self.key)
+            and self.connection.hexists(key, 'birth')
+            and not self.connection.hexists(key, 'death')
+        ):
```
selwin: I don't think we need to change anything in `register_birth()`, mind reverting this change?

terencehonles:
> > I would have thought that one of the calls to heartbeat would refresh the expiration, so I'm not sure what's going on there...
> 
> Redis/network latency issues could cause keys to expire before heartbeats arrive, although we already try to pad TTLs by 60 seconds. Looking at the [heartbeat code](https://github.com/rq/rq/blob/e7e14a7b73158189f6b4f817bf799fa6eb4f6267/rq/worker/base.py#L1162), I'd guess that the key expires before the TTL gets refreshed.
> 
> CleanShot 2026-05-04 at 20 38 42
> I see two options:
> 
> 1. Simply swap the order of the two commands (`expire` and `hset`) and run the heartbeat command in a pipeline [here](https://github.com/rq/rq/blob/e7e14a7b73158189f6b4f817bf799fa6eb4f6267/rq/worker/worker_classes.py#L96). In this case, even if the key is recreated, it will still eventually die if the worker is truly dead because the `expire` command will set a TTL.
> 2. `hset` command returns the number of new fields created. If it returns `1`, this means the worker key has already expired.
> 
> In both cases, we can try to repopulate the hash with whatever worker information that we have.

Yeah, I had done these two things. I added the "birth" checking to be more defensive in case this creeps in later, but I backed it out at your request.

rq/worker/base.py
terencehonles: done

selwin:
Thanks!

PR #2397, opened 2026-05-03
Add namedtuple for `Queue.parse_args` result

libmilos-so:
This change allows `Queue.parse_args`'s result to be safely used and ensure future compatibility.

For example, in the project I'm using RQ in, I want to perform some checks over the args/kwargs before a job is enqueued. Given RQ has some "reserved" kwargs (e.g. `timeout`), I was using `Queue.parse_args` for this so far. Recently, the `unique` var was added to the result, breaking my code, and this change should make this one usable without such issues and future-proof it

## Summary by CodeRabbit

* **Refactor**
  * Argument parsing for enqueue operations now returns a structured, named result instead of an unnamed positional tuple. Improves clarity when inspecting or passing enqueue parameters while preserving existing runtime behavior.

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -12,6 +12,7 @@ from functools import total_ordering
 from typing import (
     TYPE_CHECKING,
     Any,
+    NamedTuple,
     Optional,
     cast,
 )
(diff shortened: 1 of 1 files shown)


selwin:
Hey, I think this is a great idea. Can you rename it to `QueueArgs` and use [NamedTuple class]( https://typing.python.org/en/latest/spec/namedtuples.html#defining-named-tuples)?

libmilos-so:
Thanks! I'll do it tomorrow when I have the chance. I copied the boilerplate from the above's `EnqueueData`, want me to swap that one for `NamedTuple` as well while I'm at it?

selwin:
Thanks!

PR #2411, opened 2026-05-26
Update scheduler locking to capture ownership and detect expiring locks

### LINKED ISSUE #2338: RQ scheduler can sometimes duplicate jobs
This is essentially a carry on of https://github.com/rq/rq/issues/1354, in which the issue was mitigated but not truly solved.

Specifically, when you have multiple instances working on multiple queues, the problem described in the previous issue can still happen if you are simply unlucky with the timing.

Instead of simply bumping the amount of time a lock is held for, would a better solution here be for some actual 'ownership' of the lock to be conducted? Right now, the scheduler simply writes a pid into the cache value, but it could instead write e.g. a random, unique value that is associated at startup/construction time. Then when it comes to heartbeating the locks, it can sanity check that it itself actually does hold the ownership for the lock value and degrade gracefully if this assumption breaks down?

I've not thought it through in all that much detail, but I do think the current implementation is not solid enough for high-scale production workflows and I'd like to work through something with you to tighten it up.

terencehonles:
This updates the RQ scheduler to store a start-up token in Redis, and check the token upon start-up and heartbeat refreshes to ensure the lock is still owned by the current scheduler.

This also changes the worker's burst mode to not start the scheduler when running the maintenance tasks.

fixes: https://github.com/rq/rq/issues/2338

## Summary by CodeRabbit

* **New Features**
  * Optional "burst" scheduler mode, improved scheduler PID reporting and more accurate lock ownership resolution.
* **Bug Fixes**
  * Robust lock handling, validation and recovery to avoid lost or misattributed locks; fixes for races during acquisition, heartbeat and shutdown; scheduler auto-start behavior refined.
* **Refactor**
  * Centralized script registration/caching and targeted scheduler-lock operations; unique enqueue/schedule paths consolidated and annotated.
* **Tests**
  * Added tests for lock acquisition races and heartbeat ownership changes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rq/rq/pull/2411?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/rq/queue.py
+++ b/rq/queue.py
@@ -268,8 +268,16 @@ class Queue:
     def scheduler_pid(self) -> int | None:
         from rq.scheduler import RQScheduler
 
-        pid = self.connection.get(RQScheduler.get_locking_key(self.name))
-        return int(pid.decode()) if pid is not None else None
+        if (value := self.connection.get(RQScheduler.get_locking_key(self.name))) is None:
+            return None
+
+        parts = value.decode('ascii').split(':')
+        if len(parts) == 1:  # an old scheduler which hasn't been restarted
+            return int(parts[0])
+
+        ppid, pid = parts[:2]
+        # PID is only set after forking, so fall back to the parent's PID if the child has not forked yet
+        return int(ppid if not pid else pid)
 
     def acquire_maintenance_lock(self) -> bool:
         """Returns a boolean indicating whether a lock to clean this queue
--- a/rq/scheduler.py
+++ b/rq/scheduler.py
@@ -10,6 +10,7 @@ from datetime import datetime
 from enum import Enum
 from multiprocessing import Process, get_context
 from multiprocessing.process import BaseProcess
+from secrets import token_hex
 
 from redis import ConnectionPool, Redis
 
@@ -20,7 +21,7 @@ from .logutils import setup_loghandlers
 from .queue import Queue
 from .registry import ScheduledJobRegistry
 from .serializers import resolve_serializer
-from .utils import current_timestamp, parse_names
+from .utils import current_timestamp, parse_names, split_list
 
 ForkProcess: type[BaseProcess]
 try:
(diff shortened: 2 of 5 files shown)


rq/queue.py
```diff
@@ -268,8 +268,16 @@ def registry_cleaning_key(self):
...
+        if (value := self.connection.get(RQScheduler.get_locking_key(self.name))) is None:
+            return None
+
+        parts = value.decode('ascii').split(':')
+        if len(parts) == 1:  # an old scheduler which hasn't been restarted
+            return int(parts[0])
+
+        ppid, pid = parts[:2]
+        # PID is only set after forking, so fall back to the parent's PID if the child has not forked yet
+        return int(ppid if not pid else pid)
```
terencehonles, line 280: I noticed that `os.getpid()` is going to be different between the first call to `acquire_locks` which is done by the parent, and subsequent calls which will be done by the child. I have captured both PIDs in the value stored in Redis, and I assume this behavior here is fine because it captures what was previously happening during start-up.

I don't believe anyone was using the PID here to specifically signal the scheduler's process, but if they do it too early they will signal the worker and not the scheduler.

selwin:
If the objective is to check whether the lock is still owned by the current scheduler, wouldn't atomically checking the PID inside the locking key before renewing the lock in a single atomic lua script be sufficient?

terencehonles:
> If the objective is to check whether the lock is still owned by the current scheduler, wouldn't atomically checking the PID inside the locking key before renewing the lock in a single atomic lua script be sufficient?

The PID will collide for us since we run in a container and they'll be the only thing we're starting up. In the general case it's just asking for a collision error, which may be unlikely, but will surface as a hard to reproduce bug.

selwin:
> The PID will collide for us since we run in a container and they'll be the only thing we're starting up. 

In that case, let's use a random scheduler.name instead? It's more consistent with how workers are designed. We should also have a `scheduler.register_birth()` to store more scheduler related data such as date created, last heartbeat etc.

I think `scheduler.register_birth()` would have to be a separate PR though. Changing locking mechanism is scary enough for me, I don't want to cram too many things into a single PR.

terencehonles:
Sure that works, I just picked the secrets token because it was an easy way to get random data. I'm not sure if I arrived at that before looking at the implementation more closely, but while a uuid would lose some randomness it's definitely sufficient!

terencehonles:
> Sure that works, I just picked the secrets token because it was an easy way to get random data. I'm not sure if I arrived at that before looking at the implementation more closely, but while a uuid would lose some randomness it's definitely sufficient!

I was starting to make this change, and I realized we're probably both talking about the `CronScheduler` which is different. The `Scheduler` doesn't have a name. I was also probably thinking of `WorkerPool` which auto-names its workers, because the `CronScheduler` only has a partially random name and only if not provided (unrelated and not important).

In that case, is there anything wrong with the scheduler maintaining its own token?

selwin:
> I was starting to make this change, and I realized we're probably both talking about the CronScheduler which is different. The Scheduler doesn't have a name.

I was indeed referring to [RQScheduler](https://github.com/rq/rq/pull/2411/changes#diff-cbbecc8541e9c6995c1b7a76a73ecb343f6cf69f11a5ad05944646ea768521fbR43), which does not currently have a name and a separate hash to store it's metadata. This is why I mentioned this should first be implemented in a different PR since the churn will be quite high.

Having `name` and a separate key will bring it's design close to RQ's `Worker` and `CronScheduler`. I'll work on this sometime this week.

selwin:
FYI, I'm working on `scheduler.name` here https://github.com/rq/rq/pull/2414

selwin:
@terencehonles https://github.com/rq/rq/pull/2414 is now merged.

PR #2414, opened 2026-06-01
RQScheduler name and metadata

selwin:
## Summary by CodeRabbit

* **New Features**
  * Schedulers now persist stable identities and metadata; workers register lifecycle when running scheduled jobs in burst mode.

* **Improvements**
  * Queue locks reference scheduler identity; scheduler PID resolution and heartbeat handling are more robust.
  * Heartbeats refresh scheduler and lock TTLs; shutdown cleans up scheduler metadata.

* **Tests**
  * Expanded coverage for scheduler lifecycle, heartbeat, PID/lock behavior and error paths.

* **Chores**
  * CI matrix updated to skip an incompatible Redis/runtime combination.

### DIFF
--- a/.github/workflows/valkey.yml
+++ b/.github/workflows/valkey.yml
@@ -54,8 +54,8 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install hatch
-          # Install the specific redis-py version for this matrix combination
-          pip install redis==${{ matrix.redis-py-version }}
+          hatch run test:python -m ensurepip --upgrade
+          hatch run test:python -m pip install redis==${{ matrix.redis-py-version }}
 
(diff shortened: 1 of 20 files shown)


selwin:
Note that this is still not handled (perhaps separate PR).
```
Bug/race: lock refresh and release do not verify ownership
rq/scheduler.py:267 expires lock keys by name only, and rq/scheduler.py:279 deletes lock keys by name only. Since locks now store self.name, the scheduler can cheaply verify ownership. If scheduler A stalls past TTL, scheduler B can acquire the lock; when A wakes, it can extend or delete B’s lock.
```

terencehonles (approved):
LGTM

PR #2438, opened 2026-07-10
feat: add RQ_KEY_PREFIX env var for configurable Redis key namespace

dmonroy:
## Summary

Adds a `RQ_KEY_PREFIX` environment variable (default: `rq`) that controls the prefix used for all Redis keys managed by RQ.

All hardcoded `rq:` key prefixes across workers, queues, jobs, registries, scheduler, results, pub/sub, suspension, groups, cron, executions, and scripts are now driven by `RQ_KEY_PREFIX` in `defaults.py`, read from the environment at import time.

```python
# defaults.py
RQ_KEY_PREFIX = os.environ.get('RQ_KEY_PREFIX', 'rq')
```

Setting `RQ_KEY_PREFIX` before importing rq namespaces all Redis keys under a custom prefix. Default behaviour is unchanged.

## Motivation

Managed Redis providers (e.g. AWS ElastiCache with ACL users) often restrict which keyspaces a user can access. With all keys hardcoded to `rq:`, there is no way to align RQ's keys to an ACL-permitted keyspace pattern without forking the library.

## Files changed

- `rq/defaults.py` — adds `RQ_KEY_PREFIX` constant (reads `RQ_KEY_PREFIX` env var, default `rq`)
- `rq/worker_registration.py`, `rq/worker/base.py`, `rq/queue.py`, `rq/job.py`, `rq/registry.py`, `rq/results.py`, `rq/scheduler.py`, `rq/executions.py`, `rq/group.py`, `rq/command.py`, `rq/suspension.py`, `rq/scripts.py`, `rq/cron.py`, `rq/cron_scheduler_registry.py` — replace hardcoded `'rq:'` with `RQ_KEY_PREFIX + ':'`
- `tests/test_key_prefix.py` — new test module covering default prefix and custom prefix via `mock.patch`

## Notes

- The env var must be set **before** importing rq, as module-level constants are evaluated at import time
- Default value is `rq`, so existing deployments are unaffected
- **This branch is based on the `v2.10` release tag**, not on `master`, so it does not include any unreleased changes. This was intentional — we are using `wedgworth/rq@feature/redis-key-prefix` directly in production and did not want to pull in changes that haven't shipped yet. If you'd prefer this rebased onto `master` before merge, happy to do that.

## Summary by CodeRabbit

- **New Features**
  - Added configurable Redis key prefixes through the `RQ_KEY_PREFIX` environment variable.
  - Cron jobs now support custom names and accessible, paginated execution history.
  - Added support for stopping a specific job execution.
  - Added tracking and processing for jobs ready to be enqueued.

- **Bug Fixes**
  - Improved scheduler lock handling and recovery when locks expire or ownership changes.
  - Improved cleanup and failure handling for abandoned or failed jobs.
  - Scheduler and cron registries now use configurable expiration settings.

### DIFF
--- a/rq/worker/base.py
+++ b/rq/worker/base.py
@@ -40,6 +40,7 @@ from ..defaults import (
     DEFAULT_MAINTENANCE_TASK_INTERVAL,
     DEFAULT_RESULT_TTL,
     DEFAULT_WORKER_TTL,
+    RQ_KEY_PREFIX,
 )
 from ..exceptions import DequeueTimeout, DeserializationError, StopRequested
 from ..executions import Execution, cleanup_execution, prepare_execution
@@ -114,7 +115,7 @@ class WorkerStatus(str, Enum):
 
 
 class BaseWorker:
-    redis_worker_namespace_prefix = 'rq:worker:'
+    redis_worker_namespace_prefix = RQ_KEY_PREFIX + ':worker:'
     redis_workers_keys = worker_registration.REDIS_WORKER_KEYS
     death_penalty_class = get_default_death_penalty_class()
     queue_class = Queue
@@ -988,7 +989,7 @@ class BaseWorker:
 
         This can be used to make `ps -ef` output more readable.
         """
-        setprocname(f'rq:worker:{self.name}: {message}')
+        setprocname(f'{RQ_KEY_PREFIX}:worker:{self.name}: {message}')
 
     def set_shutdown_requested_date(self):
         """Sets the date on which the worker received a (warm) shutdown request"""
--- a/rq/command.py
+++ b/rq/command.py
@@ -8,10 +8,11 @@ if TYPE_CHECKING:
 
     from .worker import BaseWorker
 
+from rq.defaults import RQ_KEY_PREFIX
 from rq.exceptions import InvalidJobOperation
 from rq.job import Job
 
-PUBSUB_CHANNEL_TEMPLATE = 'rq:pubsub:%s'
+PUBSUB_CHANNEL_TEMPLATE = RQ_KEY_PREFIX + ':pubsub:%s'
 
 
 def send_command(connection: 'Redis', worker_name: str, command: str, **kwargs):
--- a/rq/cron.py
+++ b/rq/cron.py
@@ -18,7 +18,7 @@ from redis import Redis
 from redis.client import Pipeline
 
 from . import cron_scheduler_registry
-from .defaults import DEFAULT_LOGGING_DATE_FORMAT, DEFAULT_LOGGING_FORMAT, DEFAULT_RESULT_TTL
+from .defaults import DEFAULT_LOGGING_DATE_FORMAT, DEFAULT_LOGGING_FORMAT, DEFAULT_RESULT_TTL, RQ_KEY_PREFIX
 from .exceptions import SchedulerNotFound, StopRequested
 from .job import Job
 from .logutils import setup_loghandlers
@@ -471,7 +471,7 @@ class CronScheduler:
     @property
     def key(self) -> str:
         """Redis key for this CronScheduler instance"""
-        return f'rq:cron_scheduler:{self.name}'
+        return f'{RQ_KEY_PREFIX}:cron_scheduler:{self.name}'
 
     def to_dict(self) -> dict:
         """Convert CronScheduler instance to a dictionary for Redis storage"""
@@ -521,7 +521,7 @@ class CronScheduler:
     @classmethod
     def fetch(cls, name: str, connection: Redis) -> CronScheduler:
         """Fetch a CronScheduler instance from Redis by name."""
-        key = f'rq:cron_scheduler:{name}'
+        key = f'{RQ_KEY_PREFIX}:cron_scheduler:{name}'
         raw_data = connection.hgetall(key)
 
         if not raw_data:
--- a/rq/cron_scheduler_registry.py
+++ b/rq/cron_scheduler_registry.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 from redis import Redis
 from redis.client import Pipeline
 
+from .defaults import RQ_KEY_PREFIX
 from .exceptions import DuplicateSchedulerError, SchedulerNotFound
 
 if TYPE_CHECKING:
@@ -14,7 +15,7 @@ if TYPE_CHECKING:
 
 def get_registry_key() -> str:
     """Get the Redis key for the CronScheduler registry"""
-    return 'rq:cron_schedulers'
+    return RQ_KEY_PREFIX + ':cron_schedulers'
 
 
 def register(cron_scheduler: CronScheduler, pipeline: Pipeline | None = None) -> None:
--- a/rq/defaults.py
+++ b/rq/defaults.py
@@ -1,3 +1,9 @@
+import os
+
+RQ_KEY_PREFIX = os.environ.get('RQ_KEY_PREFIX', 'rq')
+"""The prefix used for all Redis keys managed by RQ.
+Override via the RQ_KEY_PREFIX environment variable before importing rq."""
+
 DEFAULT_JOB_CLASS = 'rq.job.Job'
 """ The path for the default Job class to use.
 Defaults to the main `Job` class within the `rq.job` module
--- a/rq/executions.py
+++ b/rq/executions.py
@@ -12,6 +12,7 @@ if TYPE_CHECKING:
 
     from .worker.base import BaseWorker
 
+from .defaults import RQ_KEY_PREFIX
 from .job import Job
 from .registry import BaseRegistry, StartedJobRegistry
 from .utils import as_text, current_timestamp, now, parse_composite_key
@@ -37,7 +38,7 @@ class Execution:
 
     @property
     def key(self) -> str:
-        return f'rq:execution:{self.composite_key}'
+        return f'{RQ_KEY_PREFIX}:execution:{self.composite_key}'
 
     @property
     def job(self) -> Job:
@@ -116,7 +117,7 @@ class ExecutionRegistry(BaseRegistry):
     Each job has its own execution registry.
     """
 
-    key_template = 'rq:executions:{0}'
+    key_template = RQ_KEY_PREFIX + ':executions:{0}'
 
     def __init__(self, job_id: str, connection: Redis):
         self.connection = connection
--- a/rq/group.py
+++ b/rq/group.py
@@ -7,6 +7,7 @@ from redis import Redis
 from redis.client import Pipeline
 
 from . import Queue
+from .defaults import RQ_KEY_PREFIX
 from .exceptions import NoSuchGroupError
 from .job import Job
 from .queue import EnqueueData
@@ -16,8 +17,8 @@ from .utils import as_text
 class Group:
     """A Group is a container for tracking multiple jobs with a single identifier."""
 
-    REDIS_GROUP_NAME_PREFIX = 'rq:group:'
-    REDIS_GROUP_KEY = 'rq:groups'
+    REDIS_GROUP_NAME_PREFIX = RQ_KEY_PREFIX + ':group:'
+    REDIS_GROUP_KEY = RQ_KEY_PREFIX + ':groups'
 
     def __init__(self, connection: Redis, name: str | None = None):
         self.name = name if name else str(uuid4().hex)
--- a/rq/job.py
+++ b/rq/job.py
@@ -15,7 +15,7 @@ from uuid import uuid4
 
 from redis import WatchError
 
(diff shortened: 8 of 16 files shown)


selwin:
@dmonroy I think this looks good. Did you look at https://github.com/rq/rq/pull/2418#issuecomment-4700757717 ? There are also conflicts that you need to resolve.

@curdbecker mind chiming in if you have the time? This is related to your PR https://github.com/rq/rq/pull/2418

dmonroy:
> @dmonroy I think this looks good. Did you look at [#2418 (comment)](https://github.com/rq/rq/pull/2418#issuecomment-4700757717) ? 

Thanks, I hadn't and wasn't aware that a similar approach was already being worked on. Must have checked there first. Will make sure this aligns with the discussion there. 

> There are also conflicts that you need to resolve.

Noticed. I'll update my current deployments of this branch so they continue to apply this on top of v2.10. Then, this PR will rebase the main branch, resolving any conflicts.

rq/worker/base.py
```diff
@@ -1022,7 +989,7 @@ def procline(self, message):
 
         This can be used to make `ps -ef` output more readable.
         """
-        setprocname(f'rq:worker:{self.name}: {message}')
+        setprocname(f'{RQ_KEY_PREFIX}:worker:{self.name}: {message}')
```
selwin, line 992: This one doesn't need to be changed since this is only a process name.

selwin:
Mind changing the approach to the specs in https://github.com/rq/rq/pull/2418#issuecomment-4700757717 ? A PR implementing the first step would be good.

selwin:
@dmonroy so I actually did a few rounds of review on how to approach Redis Cluster support, between what I proposed in https://github.com/rq/rq/pull/2418#issuecomment-4700757717 and your PR.

I actually prefer your approach since it's a lot less invasive, but I still want to be able to support Redis Cluster without the user having to explicitly configure `RQ_KEY_PREFIX`. 

Do you mind rebasing this PR to the current master so I can merge it in, with a few additions later?

For full transparency, this is the spec (the plan may get updated as we go).

# Configurable key prefix (cluster key-name foundation)

## Context

Two problems, one mechanism:

1. **Restricted keyspaces**: managed Redis providers (e.g. ElastiCache with ACL users) restrict
   which key patterns a user may access. A configurable prefix lets RQ align with an
   ACL-permitted keyspace without forking.
2. **Redis Cluster (future)**: every key touched by one atomic operation must hash to the same
   slot. A hash-tagged prefix (`{rq}` instead of `rq`) colocates the whole RQ keyspace on one
   slot. This phase only makes key names *cluster-compatible*; it is not cluster support —
   workers and schedulers currently fail on cluster before key slots matter
   (`get_connection_kwargs`/`parse_connection` and `BaseWorker._set_connection` assume a
   standalone `connection_pool`, and `ClusterPipeline` is non-transactional by default). Those
   are addressed in the follow-up phases below, and cluster support is not advertised until they
   land.

Design: `RQ_KEY_PREFIX` environment variable, default `rq`, read once at import time in
`defaults.py`. All key constants are built from it at import, so keys stay static strings —
key construction never needs a connection, and `Job.key_for(job_id)` and friends keep their
signatures. No auto-detection, no runtime prefix mutation, no resolution ordering: the prefix is
fixed for the life of the process. With the default, every key is byte-for-byte identical to
today — no migration, no ecosystem breakage.

This is the rework target for PR #2438: rebase onto master and close the gaps below.

## 1. The constant

```python
# defaults.py
RQ_KEY_PREFIX = os.environ.get('RQ_KEY_PREFIX', 'rq')
"""The prefix for all Redis keys managed by RQ. Because key constants are
built at import time, set the `RQ_KEY_PREFIX` environment variable before
importing rq."""
```

Every hardcoded `rq:` key literal becomes `RQ_KEY_PREFIX + ':...'` (class attributes, module
constants, key templates) or an f-string over it (properties, functions). Since the value never
changes after import, baked and call-time construction are equivalent — follow whichever gives
the smaller diff at each site, matching PR #2438's shape.

## 2. Site sweep

PR #2438's site list (queue, job, registry, worker, group, scheduler, results, executions, cron,
cron_scheduler_registry, suspension, worker_registration, command, scripts), rebased onto master,
plus:

- **`rate_limit.py`** (missing from the PR — it's based on v2.10, which predates #2386):
  - `config_key`, `allowed_key`, `rate_limited_key` properties and the `rl_keys_key` constant.
  - `ACQUIRE_AND_ENQUEUE_SCRIPT` (and `RELEASE_AND_ENQUEUE_SCRIPT`, which embeds it) hardcodes
    `'rq:job:' ..` and `'rq:queue:' ..` — build the script constants as module-level f-strings
    interpolating `RQ_KEY_PREFIX`.
- **Any other post-v2.10 literals** surfaced by the rebase; the grep test below catches them.
- `Queue.empty()`'s Lua needs no direct edit — it already interpolates
  `job_class.redis_job_namespace_prefix`, which the sweep covers.

Excluded:

- **`procline`** in `worker/base.py` — `rq:worker:{name}` there is a process title for `ps`
  output, not a Redis key. Drop this hunk from the PR.
- **`intermediate_queue.py`** — its keys derive from the queue key passed in.

## 3. Lua hygiene: cleanup script key declaration

`CLEANUP_REGISTRY_SCRIPT` in `rate_limit.py` passes `rl_keys_key` through `ARGV[1]` and then uses
it as a Redis key (`SREM ARGV[1] ARGV[2]`), violating the EVAL contract that scripts receive all
key names via `KEYS`. Move it to `KEYS[4]`:

```lua
redis.call('SREM', KEYS[4], ARGV[1])
```

with `keys=[self.allowed_key, self.rate_limited_key, self.config_key, self.rl_keys_key]` and
`args=[self.key]` in `cleanup()`. Behavior-neutral on standalone; separate commit.

## 4. Verification

- **Identity/regression**: with the env var unset, all existing suites pass unchanged — they
  assert hardcoded `rq:` literals and are the byte-compatibility proof.
- **Env end-to-end**: one subprocess test running a small script with `RQ_KEY_PREFIX=myapp` that
  asserts built keys (`Queue('x').key == 'myapp:queue:x'`, `Job.key_for('id')`, a registry key,
  a `RateLimitRegistry` script text containing `myapp:job:`) — covers the read-at-import
  contract without monkeypatching module globals. No shared-state teardown needed because
  nothing mutates at runtime.
- **Literal-grep test**: recursively scan `rq/**/*.py` for `'rq:`/`"rq:` string literals outside
  `defaults.py` and the `procline` exception; fail on new ones. This is what catches future
  bare-literal sites that would silently ignore the prefix (three such gaps were found while
  reviewing this plan; the test exists because builder-level tests can't see sites that bypass
  the builders).
- Docs (on request): `RQ_KEY_PREFIX` usage and the set-before-import rule, with an explicit note
  that this does not yet make RQ cluster-compatible at runtime.

## Follow-up phases (not in this PR)

**Phase 2 — live cluster test environment.** Add a Redis Cluster service to CI (or a compose
file) and inventory actual failures: worker and scheduler startup, enqueue/dequeue, dependency
`WATCH` flows, queue deletion, `Queue.empty()`, `Queue.compact()` `RENAME`, multi-queue
`BLPOP`/`BLMOVE`, rate-limit scripts. Part of the inventory is a classification of RQ's ~47
`.pipeline()` call sites (none pass `transaction` explicitly today): atomicity-required
(enqueue's job-hash+`LPUSH`+registry bundle, job-completion registry moves,
`StartedJobRegistry.cleanup`, and the `WATCH`/CAS loops — dependency setup in
`Queue.setup_dependencies`, dependents enqueueing in `enqueue_dependents`,
`Job.fetch_dependencies`, the `ScheduledJobRegistry` move) versus batching-only bulk reads
(`Job.fetch_many`, rate-limit status reads), which are fine non-transactional.

**Phase 3 — cluster runtime fixes**, driven by the phase 2 inventory. Known already:

- `get_connection_kwargs`/`parse_connection` (`connections.py`) and
  `BaseWorker._set_connection` read `connection.connection_pool`, which `RedisCluster` does not
  expose the same way.
- Standalone `Redis.pipeline()` defaults to `transaction=True` (`MULTI`/`EXEC`);
  `RedisCluster.pipeline()` defaults to non-transactional client-side batching — the same RQ
  code silently loses atomicity and `WATCH` semantics on cluster. The `WATCH`/CAS loops are the
  critical paths: unhonored `WATCH` in dependency setup means a dependency finishing mid-enqueue
  leaves the dependent deferred forever (a lost job, not a degradation). Fix: pass
  `transaction=True` explicitly at every atomicity-required site from the phase 2 classification
  (a no-op on standalone). Modern redis-py supports cluster transactions including `WATCH` when
  all keys share one slot — which the `{rq}` prefix guarantees — but this support does not exist
  at RQ's `redis>=5.0.1` floor, so cluster support requires bumping the floor or gating cluster
  usage on the redis-py version. Bonus: transactional cluster pipelines raise
  `CrossSlotTransactionError` on any key outside the shared slot, turning a missed prefix site
  into a loud failure — runtime enforcement of the phase 1 invariant.
- Tests for the above, at two layers: construction-level assertions that atomic paths build
  transactional pipelines on a cluster connection (so the silent downgrade can't regress), and
  live-cluster race tests (finish a dependency concurrently during `setup_dependencies` and
  assert the dependent runs exactly once; drop a connection mid-enqueue and assert no partial
  job state).
- Lua scripts that compute keys instead of declaring them in `KEYS` (`Queue.empty()` deletes
  `prefix..job_id`; the rate-limit acquire/release scripts discover job and queue keys at
  runtime from popped IDs). Colocation makes undeclared same-slot access work on current
  servers, but it is not a sound contract — these need real-cluster verification or redesign.

**Phase 4 — cluster ergonomics**, once phase 3 makes cluster actually work:

- Construction-time guard: cluster connection + prefix without a valid hash tag → raise with an
  actionable message. Validation follows Redis's exact rule — first `{`, then first `}` after
  it, at least one byte between — so `{`, `{}`, `foo{bar` and `{}{rq}` are all rejected
  (Redis only considers the first brace pair).
- Docs: on cluster, set `RQ_KEY_PREFIX='{rq}'`.
- Auto-detection of cluster connections is deliberately deferred and reconsidered here at the
  earliest: it requires runtime-mutable prefix state plus resolution hooks on every raw-
  connection entry point (`Queue.all`, `Group.all`, `CronScheduler.fetch`, suspension functions,
  `worker_registration.get_keys`, `RateLimitRegistry.all`, …), and the explicit-env-plus-guard
  design may make it unnecessary.

curdbecker:
@selwin Sorry for not having time yet to further look at it, but unfortunately I have a bit of a backlog that I need to still deal with 😅

Regarding the approach: I was a bit confused, since I didn’t want to be too invasive in my PR, but also this PR is basically at just the first step that we discussed without having actual Redis Cluster support.

The thing was just that my PR needed to be moved into production, so it needed a full test suite including SSL support as that caused more trouble than I magined before, then also including Valkey as AWS is currently pushing into the Valkey direction - didn’t really have luck with the timing here 😂. That’s why it got a bit “cramped”.

But happy to rebase it, or rather further split it up, once the prefix approach is decided.

selwin:
> Regarding the approach: I was a bit confused, since I didn’t want to be too invasive in my PR, but also this PR is basically at just the first step that we discussed without having actual Redis Cluster support.

@curdbecker are you ok with opening another PR implementing the prefix or would @dmonroy be able to rebase this PR against master?

curdbecker:
I like @dmonroy’s PR and if he is also already using it as I understood, then happy to adapt mine to his changes 😊

Edit: Just to be clear, after it is merged of course. Then his contribution is also completely clear PR-wise 😊

selwin:
> I like @dmonroy’s PR and if he is also already using it as I understood, then happy to adapt mine to his changes 😊

Thanks @curdbecker ! @dmonroy mind updating this PR to track the master branch? 😀

dmonroy:
> > I like @dmonroy’s PR and if he is also already using it as I understood, then happy to adapt mine to his changes 😊
> 
> Thanks @curdbecker ! @dmonroy mind updating this PR to track the master branch? 😀

Doing it this week 👍

dmonroy:
> I like @dmonroy’s PR and if he is also already using it as I understood...

I confirm my branch is in prod since before the PR was opened. That's why it is based on v2.10 and not the main branch.

curdbecker:
Hahaha - yes, mine too... Without cluster support, I would have not been able to use the AWS setup as intended. And I briefly tried celery, but then quickly decided that for posterity and myself the actually easier approach is to just tackle cluster support here again. Likely very similar to the situation you encountered 😂

selwin:
> Doing it this week 👍

@dmonroy just bumping this in case you forgot :D

paltman:
@selwin @dmonroy I just took the liberty of rebasing and resolving the conflicts.  Let me know if i screwed anything up :).  Appreciate it!

PR #2444, opened 2026-07-21
Support DICT_CONFIG config file setting in rq worker-pool

### LINKED ISSUE #2051: How to configure worker-pool via configuration file ?
This is question on whether the `worker-pool` supports `-c config.py` cli argument ?

If not, how do I load `DICT_CONFIG` to the worker-pool. I am trying to setup a custom logging for the worker-pool.

razchiriac:
Fixes #2051

`rq worker-pool` accepts `-c/--config` and already picks up the Redis connection settings and `QUEUES` from the settings module, but it silently ignores `DICT_CONFIG` — so there is no way to configure logging for the pool via a config file, which is exactly what #2051 asks for.

This applies `logging.config.dictConfig(settings['DICT_CONFIG'])` in `worker_pool` the same way the `worker` (and `cron`) commands do. Since `WorkerPool` forks its workers, the logging configuration is inherited by the pooled workers as well, and `setup_loghandlers` already respects pre-existing handlers, so a root-logger `DICT_CONFIG` isn't clobbered.

Also documents the `-c/--config` option (and which settings `worker-pool` honors) in the Worker Pool section of `docs/docs/workers.md`, and adds a CLI test. The test asserts on the installed root handler's formatter rather than captured output because the test suite globally disables sub-ERROR logging (`RQTestCase.setUpClass`).

Context: we run an RQ worker pool in production and configure it (worker count, logging) from settings rather than CLI flags, so config-file parity between `worker` and `worker-pool` scratches a real itch.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring worker-pool logging through the `-c` / `--config` option.
  * Worker pools now honor dictionary-based logging settings from configuration files.
  * Command-line options take precedence over configuration file values.

* **Documentation**
  * Documented the worker-pool configuration option and supported settings.

### DIFF
--- a/docs/docs/workers.md
+++ b/docs/docs/workers.md
@@ -578,3 +578,4 @@ Options:
 * `-P` or `--path <path>`: multiple import paths are supported (e.g `rq worker --path foo --path bar`).
 * `-j` or `--job-class <path.to.Job>`: defaults to `rq.job.Job`.
 * `--exception-handler`: path to exception handler class. Multiple exception handlers are supported (e.g `rq worker-pool --exception-handler 'path.to.my.ErrorHandler' --exception-handler 'another.ErrorHandler'`).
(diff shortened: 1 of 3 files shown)


selwin:
Thanks!

PR #2454, opened 2026-08-10
Prevent job code from swallowing timeout exceptions

### LINKED ISSUE #2296: JobTimeoutException should be derived from BaseException instead of Exception
JobTimeoutException is derived from Exception which means that it can be mistakenly swallowed by the job function if it is doing it's own handling of unhandled exceptions.

```
def my_job(...) -> None:
    try:
        some_long_running_task()
    except Exception as e:
        handle_task_error(e)
```

Exception class is usually used for non-fatal exceptions caused directly by the application code. If you're injecting some exceptions via SIGALRM handling to indicate the need for ASAP finalization due to fatal error - BaseException is much better choice.

Actually that's pretty common problem e.g. https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError used to be derived from Exception as well until 3.8. They changed that because some coroutines performing their own exception handling were uncancellable: https://bugs.python.org/issue32528

I would normally make a PR but I noticed that in some places rq is possibly catching `JobTimeoutException` using `except Exception` as well. I'm not very familiar with the exception flow in the code but I saw such handling in few places:
- https://github.com/rq/rq/blob/91d9974b265c88da3480c19b5009063ff184b197/rq/job.py#L1506
- https://github.com/rq/rq/blob/91d9974b265c88da3480c19b5009063ff184b197/rq/job.py#L1493

tianrking:
Fixes #2296.

## Problem

`JobTimeoutException` inherits from `Exception`, so a job that uses a broad application-level `except Exception` handler can accidentally catch RQ's timeout signal and continue running. That defeats the worker's execution deadline and can leave the job reported as successful.

## Change

Make RQ's timeout base class inherit from `BaseException`. RQ's control-flow handling remains explicit: job execution already catches the timeout at the worker boundary, horse monitoring catches `HorseMonitorTimeoutException`, and death-penalty cleanup catches `BaseTimeoutException` directly.

Add a regression that models an application handler and verifies that `JobTimeoutException` escapes it.

## Testing

- Regression failed on unmodified `master` because the application handler swallowed the timeout; passes after the change
- `pytest tests/test_timeouts.py` against a local Redis-compatible TCP test service (3 passed)
- Live `TimerDeathPenalty` propagation exercise on Windows
- `ruff check rq/timeouts.py tests/test_timeouts.py`
- `ruff format --check rq/timeouts.py tests/test_timeouts.py`
- `mypy rq/timeouts.py`
- `python -m compileall -q rq/timeouts.py tests/test_timeouts.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Job timeout exceptions now reliably propagate instead of being inadvertently caught by broad application error handlers.
  * Timeout behavior is clarified and more consistent for applications handling exceptions.

### DIFF
--- a/rq/timeouts.py
+++ b/rq/timeouts.py
@@ -3,8 +3,12 @@ import signal
 import threading
 
 
-class BaseTimeoutException(Exception):
-    """Base exception for timeouts."""
+class BaseTimeoutException(BaseException):
+    """Base exception for RQ's control-flow timeouts.
+
+    Timeout signals must not be intercepted by application-level
+    ``except Exception`` handlers.
+    """
 
     pass
 
--- a/tests/test_timeouts.py
+++ b/tests/test_timeouts.py
@@ -1,14 +1,28 @@
 import time
+import unittest
 from unittest.mock import patch
 
 from rq import Queue, SimpleWorker
 from rq.registry import FailedJobRegistry, FinishedJobRegistry
 from rq.timeouts import (
+    JobTimeoutException,
     TimerDeathPenalty,
     UnixSignalDeathPenalty,
     get_default_death_penalty_class,
 )
 from tests import RQTestCase
+
+
+class TestTimeoutExceptionHierarchy(unittest.TestCase):
+    def test_job_timeout_is_not_caught_by_application_exception_handlers(self):
+        def application_job():
+            try:
+                raise JobTimeoutException('timeout')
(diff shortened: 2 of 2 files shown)


selwin:
I like this change, by I'm just wondering whether there are any backward incompatible scenarios?

PR #2460, opened 2026-08-16
Fix import_attribute swallowing internal ImportError

### LINKED ISSUE #2089: `import_attribute(...)` in `utils.py` swallows import errors in custom workers classes when running via the CLI.
If you assign a `worker_class` via the CLI using dot notation for modules, e.g., `myproject.myworker.MyWorker`, then in the `utils.py` module there is a function `import_attribute(...)` that is called. One part of the logic is to automatically try to find the module by chopping off parts of the right hand side of the class path until it finds a valid module. This is done by checking for `ImportError` exceptions. The problem with this is that if your implementation of the custom worker class has an import error you get a really hard bug to understand just by running the code via the CLI. 

You get this top level error:

...
  File "/usr/local/lib/python3.11/site-packages/rq/cli/helpers.py", line 356, in __init__
    self.worker_class = import_attribute(worker_class)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/rq/utils.py", line 120, in import_attribute
    raise ValueError('Invalid attribute name: %s' % attribute_name)
ValueError: Invalid attribute name: MyWorker
```

Fine, but then you get this thing as a deeper exception:

```
  File "/usr/local/lib/python3.11/site-packages/rq/utils.py", line 118, in import_attribute
    attribute_owner = getattr(module, attribute_owner_name)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'myproject' has no attribute 'myworker'
```

This is not at all what the error is. The actual error is that `myworker` has a faulty import, but this code swallows the error:

```
        try:
            module_name = '.'.join(module_name_bits)
            module = importlib.import_module(module_name)
            break
        except ImportError as ex:
            attribute_bits.insert(0, module_name_bits.pop())
```

https://github.com/rq/rq/blob/fc4884a0f22964893e8b2b8c08eb388759498936/rq/utils.py#L101

It continues, then fails at an unrelated place further down.

### PR DESCRIPTION
**Pull request by aryansk**
## Summary

Stop `import_attribute()` from converting dependency/import failures raised inside a valid job module into the misleading `Invalid attribute name` error.

## Changes

- only shorten the candidate dotted module path when `ModuleNotFoundError.name` is the candidate itself or one of its missing parent prefixes
- allow ordinary `ImportError` and nested dependency `ModuleNotFoundError` exceptions to propagate unchanged
- retain the existing fallback behavior for genuinely nonexistent dotted paths
- add regression coverage for an internal import failure

## Validation

Focused test:
`pytest -q tests/test_utils.py -k import_attribute`

Fixes #2089

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling when loading components from dotted module paths.
  - Internal dependency errors are now surfaced accurately instead of being reported as invalid attribute paths.
- **Tests**
  - Added regression coverage to ensure import errors from valid modules are preserved.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -171,6 +171,14 @@ class TestUtils(RQTestCase):
         self.assertRaises(ValueError, import_attribute, 'non.existent.module')
         self.assertRaises(ValueError, import_attribute, 'rq.worker.WrongWorker')
 
+    def test_import_attribute_preserves_internal_import_error(self):
+        with patch(
+            'rq.utils.importlib.import_module',
+            side_effect=ImportError('dependency import failed'),
+        ):
+            with self.assertRaisesRegex(ImportError, 'dependency import failed'):
+                import_attribute('rq.worker.SimpleWorker')
+
     def test_ceildiv_even(self):
         """When a number is evenly divisible by another ceildiv returns the quotient"""
         dividend = 12
--- a/rq/utils.py
+++ b/rq/utils.py
@@ -190,7 +190,11 @@ def import_attribute(name: str) -> Callable[..., Any]:
             module_name = '.'.join(module_name_bits)
             module = importlib.import_module(module_name)
             break
-        except ImportError:
+        except ModuleNotFoundError as exc:
+            if exc.name is None or not (
+                module_name == exc.name or module_name.startswith(f'{exc.name}.')
+            ):
+                raise
             attribute_bits.insert(0, module_name_bits.pop())
 
     if module is None:

### CODE REVIEW

File: tests/test_utils.py [hunk 1]
```diff
@@ -171,6 +171,14 @@ def test_import_attribute(self):
         self.assertRaises(ValueError, import_attribute, 'non.existent.module')
         self.assertRaises(ValueError, import_attribute, 'rq.worker.WrongWorker')
 
+    def test_import_attribute_preserves_internal_import_error(self):
+        with patch(
+            'rq.utils.importlib.import_module',
+            side_effect=ImportError('dependency import failed'),
```
selwin, line 177: Instead of mocking this import, can you actually create a test worker with a broken import?

PR #2461, opened 2026-08-16
Add built-in cloudpickle serializer alias

### LINKED ISSUE #2429: Add a built-in cloudpickle serializer alias
RQ supports custom serializers with `-S <serializer-class>`, and recent versions
also support built-in aliases like `json` and `pickle`.

It would be useful to add a built-in `cloudpickle` serializer alias:

```bash
rq worker default -S cloudpickle
```

and possibly:

```python
Queue("default", serializer="cloudpickle")
```

Motivation: Hydra's RQ launcher needs cloudpickle-compatible result
serialization. Without a matching worker-side serializer, RQ can store an
unserializable-result sentinel, and the launcher later fails with a confusing
deserialize error:

- https://github.com/facebookresearch/hydra/issues/2338

Hydra can work around this by shipping its own serializer class and documenting:

```bash
rq worker default -S hydra_plugins.hydra_rq_launcher.serializer.CloudpickleSerializer
```

but `cloudpickle` seems common enough that a first-class alias may be
friendlier.

Implementation could keep `cloudpickle` optional and fail with a clear message
if the alias is requested without the dependency installed.

Related serializer work:

- https://github.com/rq/rq/issues/973
- https://github.com/rq/rq/pull/1219
- https://github.com/rq/rq/pull/2401

aryansk:
Fixes #2429

## Summary
- add a built-in cloudpickle serializer alias
- import cloudpickle lazily so it remains optional
- raise a clear import error when the optional package is unavailable
- add serializer alias regression coverage

## Testing
Validated with git apply --check against current master; full test suite not run locally. Draft PR for CI validation.

### DIFF
--- a/rq/serializers.py
+++ b/rq/serializers.py
@@ -24,6 +24,30 @@ class DefaultSerializer:
 PickleSerializer = DefaultSerializer
 
 
+class CloudpickleSerializer:
+    """Serializer backed by the optional cloudpickle dependency."""
+
+    @staticmethod
+    def dumps(obj: Any, /) -> bytes:
+        try:
+            import cloudpickle
+        except ImportError as exc:
+            raise ImportError(
+                "The 'cloudpickle' serializer requires the 'cloudpickle' package."
+            ) from exc
+        return cloudpickle.dumps(obj)
+
+    @staticmethod
+    def loads(data: bytes, /) -> Any:
+        try:
+            import cloudpickle
+        except ImportError as exc:
+            raise ImportError(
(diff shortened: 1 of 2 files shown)


selwin:
Hey, sorry I'm closing this PR. `cloudpickle` is not natively supported by RQ so we don't be adding this shorthand in RQ's core library.

aryansk:
Understood @selwin — thanks for explaining the rationale; closing here is entirely reasonable since cloudpickle isn't a native RQ serialization path.

PR #2463, opened 2026-08-18
fix: defer result_ttl for jobs with unfinished dependents

### LINKED ISSUE #2452: result_ttl deletes a finished job while a dependent that hasn't run yet still needs it
Hi, and thanks for RQ, we rely on it a lot.

We hit something with job dependencies that feels like a bug, and wanted to check with you before proposing anything.

When a job finishes, its `result_ttl` starts counting right away, even if it has dependents that haven't run yet. If one of those dependents only runs later, the dependency has already been evicted by the time the dependent runs, and the dependent fails looking it up.

Minimal repro (needs a local Redis): https://gist.github.com/jimzer/af1146df7220c9b043d98900c4928ca0

It enqueues A, then B `depends_on` A, runs only A, waits for A's short `result_ttl` to pass, then runs B:

```
after running A only:  A = JobStatus.FINISHED | B = JobStatus.QUEUED
A still in Redis?      False (B still depends on it and has not run)
after running B:       B = JobStatus.FAILED | rq.exceptions.NoSuchJobError: No such job: rq:job:...
```

Setting `RESULT_TTL=3600` keeps A around and B finishes cleanly, so `result_ttl` is the only variable.

The way we see it, the `result_ttl` of a job that is still a dependency of unfinished jobs shouldn't count from when that job finishes, but from when its dependents are done. Basically, a finished result that something still depends on shouldn't be garbage collected yet. Everything else keeps the normal "clean up finished results after N seconds" behavior.

There are a couple of details to pin down (like what to keep if a dependent fails), happy to work through those. Does this match how you see it, or is there a reason it works this way today? If you're open to the direction, we'd gladly put together a PR (opt-in if you'd rather not change the current default).

Thanks!

aryansk:
## Summary

Fixes #2452

When a job finishes but has dependents that haven't run yet, its result_ttl countdown now starts only after all dependents complete, not immediately upon the parent's completion.

## Problem

Previously, a parent job's result could be garbage-collected while dependents still needed it, causing NoSuchJobError on lookup. This happened because finished_job_registry.add() and job.cleanup() ran unconditionally in _handle_success, regardless of dependent state.

## Fix

- Adds a defer_finished_registry flag to _handle_success that skips adding to finished_job_registry when set
- In handle_job_success, checks job.dependent_ids before deciding whether to defer
- After pipeline execution, checks each dependency of the current job: if it is FINISHED and has no remaining dependents, adds it to finished_job_registry and starts its cleanup
- Same logic applied in handle_job_failure for consistency

## Testing

Added test_result_ttl_deferred_when_dependents_pending to tests/test_dependencies.py.

🤖 Generated with Codebuff
Co-Authored-By: Codebuff <noreply@codebuff.com>

### DIFF
--- a/rq/job.py
+++ b/rq/job.py
@@ -1665,6 +1665,7 @@ class Job:
         execution_id: str | None = None,
         execution_started_at: datetime | None = None,
         execution_ended_at: datetime | None = None,
+        defer_finished_registry: bool = False,
     ):
         """Saves and cleanup job after successful execution"""
         self.log.debug('Job %s: handling success...', self.id)
@@ -1686,7 +1687,7 @@ class Job:
             execution_ended_at=execution_ended_at,
         )
 
-        if result_ttl != 0:
+        if result_ttl != 0 and not defer_finished_registry:
             finished_job_registry = self.finished_job_registry
             finished_job_registry.add(self, result_ttl, pipeline)
 
--- a/rq/worker/base.py
+++ b/rq/worker/base.py
@@ -44,6 +44,7 @@ from ..exceptions import DequeueTimeout, DeserializationError, StopRequested
 from ..executions import WORKER_EXECUTIONS_KEY_TEMPLATE, Execution, cleanup_execution, prepare_execution
 from ..group import Group
 from ..job import Job, JobStatus, Retry
+from ..utils import as_text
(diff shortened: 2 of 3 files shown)


alex-deltia:
does this fix handle nested dependencies? 🤔

PR #2466, opened 2026-08-26
Don't mark finished jobs as abandoned leftover in StartedJobRegistry

### LINKED ISSUE #2424: Successful job logs `Job OK`, then later `StartedJobRegistry.cleanup()` moves it to FailedJobRegistry with AbandonedJobError
This looks similar to #2290 / #2289 and older #1507, but we are seeing it on RQ `2.6.0`, which I believe should already include the fix from #2294.

A job completes successfully and logs `Job OK`, then a few minutes later another worker's registry cleanup moves the same job to `FailedJobRegistry` due to `AbandonedJobError`.

### Environment

- RQ version: `2.6.0`
- django-rq version: `3.1`
- redis-py version: `7.0.1`
- Redis server version: `v=6.2.6 sha=00000000:0 malloc=jemalloc-5.1.0 bits=64 build=72acd210e669dec4`
- Python version: `3.12.3`
- OS: `Ubuntu 24.04.3 LTS (GNU/Linux 6.17.0-1017-aws x86_64)`
- Worker class: `django-rq worker`
- Worker command:
    ```bash
    /path/to/venv/bin/python /path/to/app/manage.py rqworker --name worker-%i high default low $RQ_FLAGS
   ```
- Per-worker flags:
    worker-1: --with-scheduler
    worker-6: no scheduler flag
- Scheduler: `worker-1` runs with `--with-scheduler`; `worker-6`, which performed the cleanup that moved the successful job to `FailedJobRegistry`, does not.
- Queues: `high default low`
- Number of workers: `6`
- Queue affected: `low`
- Job timeout: 180
- Result TTL: default 500

### What happened

The job was a short S3 upload task:

id: `1234567890`
queue: `low`
callable: `app.module.upload_to_s3`
worker: `worker-1`
kwargs:
  filepath: `/path/to/local/file.wav`
  bucket: `example-bucket`
  key: `anonymised/path/file.wav`

The job function is essentially:

```python
def upload_to_s3(filepath, bucket, key):
    s3_client = boto3.client("s3")
    s3_client.upload_file(filepath, bucket, key)
    logger.info("Uploaded file to s3: %s, %s", bucket, key)
```

The worker logs show the job succeeded:

```
2026-06-19T19:10:45+00:00 ... rq.worker@1095:
low: app.module.upload_to_s3(bucket='example-bucket', filepath='/path/to/local/file.wav', key='anonymised/path/file.wav')
(1234567890)

2026-06-19T19:10:46+00:00 ... urllib3.connectionpool:
https://example-bucket.s3.<region>.amazonaws.com:443 "PUT /anonymised/path/file.wav HTTP/1.1" 200 0

2026-06-19T19:10:46+00:00 ... app.module:
Uploaded file to s3: example-bucket, anonymised/path/file.wav

2026-06-19T19:10:46+00:00 ... rq.worker@1619:
Successfully completed app.module.upload_to_s3(...) job in 0:00:00.351361s on worker worker-1

2026-06-19T19:10:46+00:00 ... rq.worker@1701:
low: Job OK (1234567890)

2026-06-19T19:10:46+00:00 ... rq.worker@1710:
Result is kept for 500 seconds
```

But about 4.5 minutes later, registry cleanup moved the same successful job to failed:

```
2026-06-19T19:15:10+00:00 ... rq.worker@449:
Worker worker-2: cleaning registries for queue: low

2026-06-19T19:15:10+00:00 ... rq.registry@303:
StartedJobRegistry cleanup: 1234567890 Moved to FailedJobRegistry, due to AbandonedJobError, at 2026-06-19
19:15:10.338304+00:00
```

Afterwards:

```python
from django_rq import get_connection
from rq.job import Job

job = Job.fetch("1234567890", connection=get_connection("low"))
print(job.get_status(refresh=True))
print(job.exc_info)
result = job.latest_result()
print(result.type if result else None)
print(result.exc_string if result else None)
print(job.worker_name, job.started_at, job.ended_at)
```

Output:

```
JobStatus.FAILED
None
None
None
worker-1 2026-06-19 19:10:45.823210+00:00 2026-06-19 19:10:46.174571+00:00
```

The uploaded object exists in S3, and dependent/downstream work also ran, so the job genuinely completed.

### Expected behavior

After Job OK, the job execution should be removed from `StartedJobRegistry`, and later cleanup should not mark it failed as abandoned.

### Actual behavior

The job logs successful completion, but later `StartedJobRegistry.cleanup()` treats it as abandoned and moves it to `FailedJobRegistry`.

### Impact

This creates many false failed jobs. Since the result expires after 500 seconds, later inspection shows no traceback and no latest result, making it look like a silent failure.

gyanu2507:
Fixes #2424.

If `handle_job_success` hits a WatchError, `cleanup_execution` has already popped the in-memory execution. On retry the zset entry never gets removed, the job is already FINISHED, and a later `StartedJobRegistry.cleanup()` treats it as abandoned.

Two changes:
- put the execution back on WatchError so the retry still zrem's it
- skip FINISHED / FAILED / STOPPED / CANCELED jobs in cleanup instead of moving them to FailedJobRegistry

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented expired jobs that are not actively running from being incorrectly treated as abandoned during cleanup.
  * Avoided unintended retries, failure records, callbacks, dependent job enqueuing, and rate-limit changes for queued, scheduled, deferred, or completed jobs.
  * Improved cleanup reliability after successful jobs when Redis operations require retries.
* **Tests**
  * Added coverage confirming finished jobs are removed cleanly without status changes or failure registration.

### DIFF
--- a/rq/executions.py
+++ b/rq/executions.py
@@ -245,8 +245,6 @@ def cleanup_execution(worker: BaseWorker, job: Job, pipeline: Pipeline, executio
         execution: The execution to clean up
     """
     logger.debug('Cleaning up execution of job %s', job.id)
-    if execution is None:
-        execution = worker.execution
     if execution:
         execution.delete(job=job, pipeline=pipeline)
         worker.executions.pop(execution.id, None)
--- a/rq/executions.py
+++ b/rq/executions.py
@@ -245,6 +245,8 @@ def cleanup_execution(worker: BaseWorker, job: Job, pipeline: Pipeline, executio
         execution: The execution to clean up
     """
     logger.debug('Cleaning up execution of job %s', job.id)
+    if execution is None:
+        execution = worker.execution
     if execution:
         execution.delete(job=job, pipeline=pipeline)
         worker.executions.pop(execution.id, None)
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -289,6 +289,25 @@ class StartedJobRegistry(BaseRegistry):
                 except NoSuchJobError:
                     continue
 
+                # A leftover StartedJobRegistry entry after a successful (or already
+                # failed/stopped) job must not be treated as AbandonedJobError.
+                # WatchError retries in handle_job_success can leave the zset entry
+                # behind even though the job already finished; just drop it.
+                status = job.get_status()
+                if status in {
+                    JobStatus.FINISHED,
+                    JobStatus.FAILED,
+                    JobStatus.STOPPED,
+                    JobStatus.CANCELED,
+                }:
+                    logger.debug(
+                        '%s cleanup: skipping %s with terminal status %s',
+                        self.__class__.__name__,
+                        job.id,
+                        status,
+                    )
+                    continue
+
                 # No real failure traceback exists for an abandoned job (the work-horse died
                 # in another process), so pass None in the exc_info traceback slot.
                 # A raising failure callback must not abort the batch or stop the job from
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -289,23 +289,9 @@ class StartedJobRegistry(BaseRegistry):
                 except NoSuchJobError:
                     continue
 
-                # A leftover StartedJobRegistry entry after a successful (or already
-                # failed/stopped) job must not be treated as AbandonedJobError.
-                # WatchError retries in handle_job_success can leave the zset entry
-                # behind even though the job already finished; just drop it.
-                status = job.get_status(refresh=False)
-                if status in {
-                    JobStatus.FINISHED,
-                    JobStatus.FAILED,
-                    JobStatus.STOPPED,
-                    JobStatus.CANCELED,
-                }:
-                    logger.debug(
-                        '%s cleanup: skipping %s with terminal status %s',
-                        self.__class__.__name__,
-                        job.id,
-                        status,
-                    )
+                # Leftover registry entries after success/failure are not abandoned
+                # jobs. Only STARTED still means the worker died mid-request.
+                if job.get_status(refresh=False) != JobStatus.STARTED:
                     continue
 
                 # No real failure traceback exists for an abandoned job (the work-horse died
--- a/rq/registry.py
+++ b/rq/registry.py
@@ -293,7 +293,7 @@ class StartedJobRegistry(BaseRegistry):
                 # failed/stopped) job must not be treated as AbandonedJobError.
                 # WatchError retries in handle_job_success can leave the zset entry
                 # behind even though the job already finished; just drop it.
-                status = job.get_status()
+                status = job.get_status(refresh=False)
                 if status in {
                     JobStatus.FINISHED,
                     JobStatus.FAILED,
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -1278,6 +1278,34 @@ class TestWorker(RQTestCase):
             #   So before that fix the call count was 4 instead of 3
             self.assertEqual(mocked.call_count, 3)
 
+    def test_handle_job_success_watcherror_removes_execution(self):
+        """WatchError retry still removes the job from StartedJobRegistry."""
+        queue = Queue(connection=self.connection)
+        worker = Worker([queue], connection=self.connection)
+        job = queue.enqueue(say_hello)
+        job.started_at = now()
+        job.ended_at = now()
+        execution = worker.prepare_execution(job)
+        registry = StartedJobRegistry(connection=self.connection)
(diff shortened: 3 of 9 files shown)


rq/executions.py
```diff
@@ -245,6 +245,8 @@ def cleanup_execution(worker: BaseWorker, job: Job, pipeline: Pipeline, executio
         execution: The execution to clean up
     """
     logger.debug('Cleaning up execution of job %s', job.id)
+    if execution is None:
+        execution = worker.execution
```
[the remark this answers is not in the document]
selwin: Mind removing these two lines? In the issue that this PR addresses, `cleanup_execution()` is always called with `execution` argument.

rq/registry.py
```diff
@@ -289,6 +289,25 @@ def cleanup(self, timestamp: float | None = None, exception_handlers: list | Non
                 except NoSuchJobError:
                     continue
 
+                # A leftover StartedJobRegistry entry after a successful (or already
+                # failed/stopped) job must not be treated as AbandonedJobError.
+                # WatchError retries in handle_job_success can leave the zset entry
+                # behind even though the job already finished; just drop it.
+                status = job.get_status()
```
selwin: Can you use `get_status(refresh=False)` here so we don't make an extra Redis call during cleanup?

tests/test_worker.py
```diff
@@ -1278,6 +1278,34 @@ def new_move_dependents_to_ready(self, job, *args, **kwargs):
             #   So before that fix the call count was 4 instead of 3
             self.assertEqual(mocked.call_count, 3)
 
+    def test_handle_job_success_watcherror_removes_execution(self):
```
selwin: This test is not relevant to the `StartedJobRegistry.cleanup()` fix, can you delete this?

rq/registry.py
```diff
@@ -289,6 +289,25 @@ def cleanup(self, timestamp: float | None = None, exception_handlers: list | Non
                 except NoSuchJobError:
                     continue
 
+                # A leftover StartedJobRegistry entry after a successful (or already
+                # failed/stopped) job must not be treated as AbandonedJobError.
+                # WatchError retries in handle_job_success can leave the zset entry
+                # behind even though the job already finished; just drop it.
+                status = job.get_status(refresh=False)
+                if status in {
```
selwin: One more thing, can you change this to `if job.get_status(refresh=False) != JobStatus.STARTED:`? Since this is started job registry, this is the only status we need to care about, much simpler. Thanks!

selwin:
Mind fixing the test errors @gyanu2507 ?

gyanu2507:
Fixed in a372198.

The failures were all the same thing: seven tests fake an abandoned job by `zadd`-ing an expired entry into `StartedJobRegistry` while the job is still `QUEUED`. A real worker goes through `prepare_for_execution()` first, which writes `STARTED`, so those fixtures were describing a state that can't actually occur — and the new `!= JobStatus.STARTED` check skips them. I added the missing `set_status(JobStatus.STARTED)` to each setup rather than loosening the check.

One of them was worth flagging: `test_abandoned_retry_does_not_fire_failed_webhook` was still green, because `assert_not_called()` is also satisfied when the job never gets processed. It now exercises the retry path it was written for.

Full suite is green locally (704 passed, 22 skipped).

gyanu2507:
Follow-up: I've dropped the `handle_job_success` change too, in 1a2c1a1. `rq/registry.py` is now the only production change in this PR.

Deleting that `test_worker.py` test sent me to look at what it was actually covering, and the answer is nothing reachable. The change put the popped entry back into `worker.executions` on a `WatchError`, but nothing reads that dict on the retry — as you pointed out, `cleanup_execution()` is always passed `execution` explicitly, so the retry re-queues the `ZREM` from the local object regardless. I reverted just the worker hunk on top of this branch and drove a `WatchError` through `handle_job_success`: `StartedJobRegistry` ends up empty either way, and the job lands `FINISHED` either way. So the two lines were unreachable, which is also why codecov was reporting 50% patch coverage.

The `tests/test_executions.py` edit came out as well — it belonged to the `worker.execution` fallback you had me remove, and the original passes unchanged without it.

Diff is now `rq/registry.py` (+5) plus the test fixtures. Full suite: 717 passed, 22 skipped, no new failures against the branch point.

gyanu2507:
@selwin test fixtures now set `STARTED` before they fake an abandoned job, and the `handle_job_success` hunk is gone — `rq/registry.py` is the only production change. Ready for another look.

selwin:
Thanks!

PR #2479, opened 2026-09-10
Close event loops after coroutine job execution

### LINKED ISSUE #1960: Too many open files with async functions
I'm processing async functions with rq simple worker.

After running for a while, I am getting: `[Errno 24] Too many open files` every time a new event loop is being created

I think too many selectors are being created since the event loops aren't explicitly closed

Kuang-xianxin:
Coroutine jobs executed through `Job.perform()` create an event loop but never close it, on either success or failure. In a long-lived SimpleWorker this can accumulate loop resources, as reported in #1960.

Use `asyncio.run()` to manage the loop, cancel and await remaining tasks, and finalize asynchronous generators. Regression tests cover return values, exception/cancellation propagation and resource cleanup. They now live beside the existing async-job test in `tests/test_job.py::TestJob`, using `RQTestCase`, `self.connection` and unittest assertions.

Fixes #1960.

Validation:
- The four regressions fail against the original implementation; those four plus the existing async-job test pass on Python 3.10.20 and 3.13.14 with a dedicated local Redis 8.10.1 Windows build.
- The complete `tests/test_job.py` run on Windows has 60 passing tests and two existing worker-process failures (`test_blocking_result_fetch` and `test_job_access_within_job_function`). Both failures reproduce against the unchanged base; the worker uses POSIX `os.fork()` unavailable on native Windows. The complete Linux integration matrix remains an upstream CI check.
- Full-source/test Ruff, changed-file formatting and applicable pre-commit hooks pass. Linux-target mypy passes for all 45 source files. The native Windows mypy hook is skipped; its previously verified platform diagnostics are unchanged by this test-only follow-up.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved execution of asynchronous jobs by ensuring their event loops are properly managed and closed after completion.
  - Ensured asynchronous resources, including pending tasks and generators, are cleaned up reliably.
  - Preserved job context correctly when asynchronous jobs succeed, fail, or are cancelled.


selwin:
Hi, thanks for the PR. I'd like to merge this PR in, but before that, can you move the tests to an existing test file, following the same testing convention used in this repo?

Kuang-xianxin:
Moved the four regressions into the existing `TestJob` class in `tests/test_job.py` in eda90e9, alongside the async-job test. They now use `RQTestCase`, `self.connection`, unittest assertions and separate named cases; the extra test file is removed.

All four still fail against the original implementation. The four plus the adjacent async-job test pass on Python 3.10/3.13 with local Redis; the full file has 60 passes and two native-Windows worker failures also reproduced on the unchanged base. Ruff, applicable hooks and Linux-target mypy pass. The production fix is unchanged.

Kuang-xianxin:
Hi @selwin, following up on the requested test restructuring in `eda90e9`: the four regressions are now in the existing `TestJob` class and use the repository's `RQTestCase` conventions. The current head has 30 successful checks and no merge conflicts as of September 20. Could you take another look when you have time, or let me know if anything else is needed before merging? Thanks!

selwin:
Thanks!

PR #2482, opened 2026-09-11
Support methods and callable objects as job callbacks

### LINKED ISSUE #1500: Support (class) methods, callables for job callbacks
Currently, only functions are supported for the callbacks:

https://github.com/rq/rq/blob/5590aab4581f9f2f4d7c34a3d09c8eb708aee3c5/rq/job.py#L124-L132

It would be useful to be able to pass things like class methods for callbacks, similar to how class methods are supported as the work function itself:

https://github.com/rq/rq/blob/5590aab4581f9f2f4d7c34a3d09c8eb708aee3c5/rq/job.py#L124-L132

nickita-khylkouski:
## Summary
- accept bound methods and callable instances as callbacks
- persist callback instances with the job serializer so state survives save and fetch
- preserve timeout handling and reject async callable instances

Closes #1500

## Tests
- callback and decorator tests (29 passed, 12 subtests passed)
- Ruff, formatting, and mypy
- full suite excluding the unrelated local Redis blocking timeout test (714 passed, 22 skipped, 1 deselected)

## Summary by CodeRabbit

- **New Features**
  - Callback handlers now support callable instances, bound methods, and class methods.
  - Callback instance state is preserved when jobs are stored and restored.
  - Persisted callback instances are loaded when needed, including after job restoration.

- **Bug Fixes**
  - Callback validation now consistently rejects non-callable values and classes.
  - Asynchronous callback handlers, including partially wrapped callbacks, are rejected with a clear error instead of being executed unsupported.
  - Callback deserialization failures now raise a clear error.

### DIFF
--- a/rq/callbacks.py
+++ b/rq/callbacks.py
@@ -15,8 +15,8 @@ if TYPE_CHECKING:
 
 class Callback:
     def __init__(self, func: str | Callable[..., Any], timeout: Any | None = None):
-        if not isinstance(func, str) and not inspect.isfunction(func) and not inspect.isbuiltin(func):
-            raise ValueError('Callback `func` must be a string or function')
+        if not isinstance(func, str) and (inspect.isclass(func) or not callable(func)):
+            raise ValueError('Callback `func` must be a string or callable')
 
         self.func = func
         self.timeout = parse_timeout(timeout) if timeout else CALLBACK_TIMEOUT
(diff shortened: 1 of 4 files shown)


nickita-khylkouski:
Addressed the partial-wrapped async callable finding in 7d94911. The shared detector now unwraps nested functools.partial objects before checking the function or callable instance, and regression coverage exercises success, failure, and stopped callbacks. Local validation: 716 passed, 22 skipped; Ruff and mypy also pass.

PR #2484, opened 2026-09-13
fix(queue): pass the caller's pipeline through when enqueueing a rate-limited job

### LINKED ISSUE #2483: enqueue_job drops the caller's pipeline for rate-limited jobs, so a rolled-back job still runs
## What happens

`Queue.enqueue_job()` takes a `pipeline` and passes it to `_enqueue_job()`, but the rate-limited branch one line above drops it:

```python
if job.has_rate_limit:
    return self._enqueue_rate_limited_job(job, at_front=at_front)
return self._enqueue_job(job, pipeline=pipeline, at_front=at_front, unique=unique)
```

`_enqueue_rate_limited_job` then sees `pipeline is None`, opens its own pipeline, executes it, and calls `acquire_and_enqueue`. So the job is written and pushed onto the queue immediately, inside what the caller believes is an open transaction.

If the caller discards that transaction, the job still runs.

## Reproduction

```python
from redis import Redis
from rq import Queue
from rq.rate_limit import RateLimit

conn = Redis(db=15)
conn.flushdb()
q = Queue('rl-repro', connection=conn)

pipe = conn.pipeline()
job = q.enqueue_call(say_hello, rate_limit=RateLimit(key='k', concurrency=5), pipeline=pipe)
pipe.reset()   # caller discards the transaction

print(conn.exists(job.key))  # 1
print(job.get_status(refresh=True))  # JobStatus.QUEUED
print(q.count)  # 1
```

Same call without `rate_limit=` writes nothing and leaves `q.count` at 0, which is the behaviour I expected in both cases.

On 2.12.0 and on master at 0e9ffe2.

## Why I think it is an oversight rather than a design choice

`_enqueue_rate_limited_job` already supports a caller-owned pipeline, and its docstring spells out the contract: buffer the ops, let the caller EXEC, then call `acquire_and_enqueue` so promotion observes committed state. Both other call sites do exactly that, `registry.py:621` and `scheduler.py:271`. `enqueue_job` is the only place that has a pipeline in hand and does not pass it on.

## The fix, and the part worth your opinion

Passing it through is one line and makes the discarded-transaction case match the unlimited one.

The trade-off is that a caller doing `enqueue_call(..., rate_limit=..., pipeline=pipe)` followed by `pipe.execute()` then gets a job sitting in `rate_limited` until something calls `acquire_and_enqueue`, rather than one already on the queue. That is the documented contract for the pipeline path and `RateLimitRegistry.cleanup()` picks such jobs up during maintenance, but it is a visible change for anyone using that combination today, so I would rather you called it than assume.

Happy to send the patch, I have it working with tests.

chiruu12:
Fixes #2483.

`Queue.enqueue_job()` was dropping the caller's `pipeline` on the rate-limited branch, so a rate-limited job was written and pushed onto the queue immediately instead of being buffered onto the caller's transaction. Discarding that transaction left the job queued and it ran anyway.

`_enqueue_rate_limited_job` already handles a caller-owned pipeline, and `registry.py:621` and `scheduler.py:271` both use it that way. This just passes the pipeline along at the third call site.

Five tests in `TestRateLimitEnqueueWithPipeline`. Three fail without the change:

- `test_discarded_pipeline_writes_nothing`
- `test_executed_pipeline_commits_then_promotes`
- `test_discarded_pipeline_leaves_an_existing_registry_untouched`

The other two are regression guards and pass either way: the unlimited path already discarded correctly, and the pipeline-less path still promotes within the call.

Full suite goes from 706 to 711 passed, 22 skipped, no other change.

Worth a second opinion on one thing, which I raised on the issue as well. After this, `enqueue_call(..., rate_limit=..., pipeline=pipe)` followed by `pipe.execute()` leaves the job in `rate_limited` until something calls `acquire_and_enqueue`, where today it would already be queued. That is the documented contract for the pipeline path and `RateLimitRegistry.cleanup()` picks those up, but it is a visible change for anyone using that combination, so say the word if you want promotion wired in somewhere instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rate-limited jobs now correctly use the caller’s pipeline context during enqueueing.
  * Discarded pipelines no longer leave behind jobs, queue entries, or registrations.
  * Rate-limit slots are preserved when a transaction is dropped.
  * Jobs committed through a pipeline remain rate-limited until execution completes.


tests/test_rate_limit.py
```diff
@@ -827,3 +827,76 @@ def test_abandoned_retry_slot_management(self):
...
+
+
+class TestRateLimitEnqueueWithPipeline(RQTestCase):
+    """Test that enqueueing a rate-limited job honors a caller-owned pipeline."""
+
+    def setUp(self):
+        super().setUp()
+        self.queue = Queue('default', connection=self.connection)
+
+    def test_discarded_pipeline_writes_nothing(self):
```
selwin, line 839: Mind putting this test in an existing testcase class rather than creating a new one?

tests/test_rate_limit.py
```diff
@@ -827,3 +827,76 @@ def test_abandoned_retry_slot_management(self):
...
+        pipe.reset()
+
+        self.assertFalse(self.connection.exists(job.key))
+        self.assertEqual(self.queue.count, 0)
+        registry = RateLimitRegistry(key='test', connection=self.connection)
+        self.assertEqual(registry.get_allowed_job_count(), 0)
+        self.assertEqual(registry.get_rate_limited_job_count(), 0)
+        self.assertNotIn(b'test', self.connection.smembers(RateLimitRegistry.rl_keys_key))
+
+    def test_discarded_pipeline_matches_the_unlimited_case(self):
```
selwin, line 854: Would be nice if we can condense everything to a single test case too, I think the tests are too verbose :)

PR #2487, opened 2026-09-17
Close the event loop after running a coroutine job

VARSITYLUNCHBUG:
# Coroutine jobs leak an event loop

`Job._execute()` creates a new asyncio event loop for `async def` jobs and never closes it. Each job leaks a few fds. After enough jobs the worker hits ulimit and unrelated I/O starts failing.

We hit this in production after ~17 days as asyncio's `deallocating an open event loop` warning.

The loop is created here and just dropped:

```python
if asyncio.iscoroutine(result):
    loop = asyncio.new_event_loop()
    coro_result = loop.run_until_complete(result)
    return coro_result
```

Repro: enqueue and `perform()` an async job in a loop in one process, then watch `/proc/self/fd`. On current master (`0e9ffe2`), 25 jobs went 6 → 81 fds. Closing the loop keeps it at 6.

```python
import os
from redis import Redis
from rq import Queue
from rq.job import Job
from tests.fixtures import say_hello_async

conn = Redis()
queue = Queue(connection=conn)
before = len(os.listdir(f"/proc/{os.getpid()}/fd"))

for i in range(25):
    job = queue.enqueue(say_hello_async, name=f"job{i}")
    Job.fetch(job.id, connection=conn).perform()

print(before, len(os.listdir(f"/proc/{os.getpid()}/fd")))
```

Fix is just closing it in a `finally`:

```python
if asyncio.iscoroutine(result):
    loop = asyncio.new_event_loop()
    try:
        return loop.run_until_complete(result)
    finally:
        loop.close()
```

Sanity checked against both `redis:7-alpine` and `valkey/valkey:8-alpine` (disposable
containers, not shared infra).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup when running asynchronous jobs by closing their event loops after completion.
  * Reduced the risk of resource leaks when processing multiple asynchronous jobs.
  * Ensured asynchronous jobs continue to return results correctly, including when jobs raise errors.
* **Tests**
  * Added coverage confirming event loops are closed for successful and failed asynchronous jobs.


VARSITYLUNCHBUG:
Did not see #2479 

It is a more comprehensive implementation.

